PR #4, opened 2023-04-10
G0.01 configuration r rules

reynoldtan:
This PR is setting up R transformation rules configuration page and a dashboard describing what each configuration and phenotypes sub-module status (enabled or disabled).

Screen Shot 2023-04-10 at 1 49 36 PM

Screenshot showing dashboard in host/admin/tripal/extension/tripal-cultivate/phenotypes and R rules configuration page in host/admin/tripal/extension/tripal-cultivate/phenotypes/r-rules

Included with this PR are passing tests for module install and update R rules using the form.
Screen Shot 2023-04-10 at 1 53 26 PM
Screen Shot 2023-04-10 at 1 54 13 PM


trpcultivate_phenotypes/tests/src/Functional/InstallTest.php
```diff
@@ -32,20 +32,14 @@ class InstallTest extends ChadoTestBrowserBase {
    */
   protected static $module_machinename = 'trpcultivate_phenotypes';
 
-  /**
-   * A small excert from your help page.
-   * Do not cross newlines.
-   */
-  protected static $help_text_excerpt = 'pports collecting all data for a specific trait (e.g. Plant Height) into a single';
-
```
laceysanderson, line 40: This is used in the testHelp() check below.

trpcultivate_phenotypes/tests/src/Functional/InstallTest.php
```diff
@@ -60,40 +54,5 @@ public function testLoad() {
     $status_code = $session->getStatusCode();
     $this->assertEquals(200, $status_code, "The module install page should be able to load $context.");
     $this->assertSession()->pageTextContains( self::$module_name );
-
-  }
-
-  /**
-   * Tests the module overview help.
-   */
-  public function testHelp() {
```
laceysanderson: ⚠️ 
Please don't remove my test ;-p I know it's not amazing but we do want to keep help pages in our modules (which we should also ensure are helpful which they currently are not) and this test ensures that we keep coverage up for them.

trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
```diff
@@ -0,0 +1,21 @@
+# Tripal Cultivate Phenotypes Configuration/Settings.
+
+# trpcultivate.phenotypes.r_config:
```
laceysanderson: ⚠️
Move these comments to right above the config they are referring to. This ensures that once our settings become expansive, it's still easy to find this documentation when updating defaults.

trpcultivate_phenotypes/config/schema/trpcultivate_phenotypes.schema.yml
```diff
@@ -0,0 +1,32 @@
...
+        phenotypes:
+          type: mapping
+          label: 'Phenotypes Global Configuration'
+          mapping:
+            r_config:
+              type: mapping
+              label: 'R Translation Rules'
+              mapping:
+                words:
+                  type: string
```
laceysanderson: 🤔
Perhaps these settings should be arrays instead of strings. The current approach makes it easy for the settings form but then requires additional processing every time we want to apply these rules. If you make this a simple array in the settings then its a simple implode/explode in the settings form and when it's being used no extra processing is required.

trpcultivate_phenotypes/config/schema/trpcultivate_phenotypes.schema.yml
```diff
@@ -0,0 +1,32 @@
...
+                  label: 'R Rules for words to remove'
+                chars:
+                  type: string
+                  label: 'R Rules for characters to remove'
+                replace:
+                  type: string
```
laceysanderson: 🤔 
We may want to think more about the format of this one in particular. A simple array might not be the best solution but rather have an array of elements 'find' and 'replace' 🤔 but not sure as that may be overkill. We'll want to specifically think about how this will be used and use that to inform how to format it in the settings.

trpcultivate_phenotypes/src/Controller/TripalCultivatePhenotypesSettingsController.php
```diff
@@ -0,0 +1,93 @@
...
+      best support your data.';  
+
+    $element['ontologies'] = [
+      '#type' => 'details',
+      '#title' => $this->t('ONTOLOGY TERMS'),
+      '#description' => $this->t('@describe<br>@link', ['@describe' => $describe, '@link' => $link]),
+      '#open' => TRUE,
+    ];
+
+    $page_elements = \Drupal::service('renderer')
```
laceysanderson: ⚠️ 
I believe you can just pass the $elements array as your return without first rendering it 🤔 By passing the render array, other modules could alter or modify this if needed before it is finally rendered by Drupal. This supports an extension module easily adding their own config to this page.

trpcultivate_phenotypes/src/Form/TripalCultivatePhenotypesRSettingsForm.php
```diff
@@ -0,0 +1,90 @@
...
+    ];
+
+    
+    return parent::buildForm($form, $form_state);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function submitForm(array &$form, FormStateInterface $form_state) {
```
laceysanderson, line 119: ⚠️ 
It would be good to validate the user input and provide feedback rather than just trusting them and setting the settings. That way, when you later use these settings you can assume they are correct because they were tested here 😜

trpcultivate_phenotypes/tests/src/Functional/ConfigRRulesTest.php
```diff
@@ -0,0 +1,79 @@
...
+   * Admin user with admin privileges.
+   *
+   * @var \Drupal\user\UserInterface
+   */
+  protected $admin_user;
+
+  /**
+   * Test R transformation rules configuration page.
+   */
+  public function testForm() {
```
laceysanderson, line 38: This is a nice functional review :-)

⚠️ 
You'll also want to add a call to specific methods in the controllers as more of a unit test. The functional test confirms from a user perspective and the unit test confirms from a developer API perspective. Also, you'll notice your code coverage is more focused on the unit test versus the functional one... so this PR has very low test coverage despite the good functional test.

laceysanderson (changes requested):
My review comments should show up in-line. Those with a 🤔 emoji are ones you may want to consider and ones with a ⚠️ should be addressed.

trpcultivate_phenotypes/tests/src/Functional/InstallTest.php
reynoldtan: Restored and tests pass

trpcultivate_phenotypes/config/schema/trpcultivate_phenotypes.schema.yml
reynoldtan: Revised schema from string to array for all 3 configurations

trpcultivate_phenotypes/config/schema/trpcultivate_phenotypes.schema.yml
reynoldtan: Revised schema and converted data type from string to sequence

trpcultivate_phenotypes/src/Controller/TripalCultivatePhenotypesSettingsController.php
reynoldtan: Passed element render array without first rendering 👍

trpcultivate_phenotypes/src/Form/TripalCultivatePhenotypesRSettingsForm.php
reynoldtan: Implemented validateForm() and unit test

trpcultivate_phenotypes/tests/src/Functional/ConfigRRulesTest.php
reynoldtan: Prepared functional test and unit test to RRules configuration

trpcultivate_phenotypes/tests/src/Unit/ConfigRRulesFormTest.php
```diff
@@ -0,0 +1,145 @@
...
+    
+    // Validation: WORDS Rule
+    // words - any words at least 2 characters long and no empty value.
+    // Failed, Has validation error:
+    $field = 'words';
+    foreach(['R', 'r', '.', '~', '1', '123', ' '] as $rule) {
+      $form_state->setValue($field, $rule);
+      $rrules->validateForm($form, $form_state);
+
+      $this->assertTrue($form_state->hasAnyErrors(), $rule);
```
laceysanderson: I think you need the `$form_state->clearErrors();` inside this loop 🤔 

I know you're setting the same field but I believe the errors may just be building up. You can check this by printing the errors in the form state to the screen after the loop completes.

laceysanderson (commented):
The unit tests look good although I'd do a quick check as mentioned in the comment + update all tests that expect errors to clear errors within the loop if it proves true.

You do still need a very simple unit test for 
- buildForm: literally just call the buildForm method and assert it returns an array.
- submitForm: mock the form state with working values for words, char and replace then call the submitForm method and check that the settings in the test environment were set.

trpcultivate_phenotypes/tests/src/Unit/ConfigRRulesFormTest.php
reynoldtan: Part of the recent commit included passing unit test for buildForm and formValidate. The issue with submit was it could save configuration because it could not retrieve the configuration name. I will revisit this issue.

laceysanderson (approved):
Tests are passing now and code looks good. I'm going to approve this and we can come back to the submit unit tests another time. All other code looks good.

PR #11, opened 2023-06-02
Fixes merge issue in ontology config branch

reynoldtan:
This PR contains codes specific to Ontology Configuration page only.

Ontology and Terms configuration page.

Screen Shot 2023-06-01 at 1 48 45 PM

This PR is for Ontology and Terms configuration page and contains passing functional tests and partial and passing unit tests.

Note:
1. A # sign in instructions/description seen in the form where a links used to be in AP for Drupal 7. Links are to create term and vocabulary which are not available in my Tripal development site.


laceysanderson (changes requested):
This is a massive PR 🥺 
The services should have been their own separate PR... I am holding off on reviewing this PR until we have come to an agreement on the services.

In our meeting this morning we have discussed changing the terms service as follows: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/12

The database and vocabulary services will go away since they are a single, simple query and not specific to our module (they are very generic and if we did need such a service it would go in core).

I have includes comments on the database service as an example of dependency injection even though that service will be going away.

We have not yet discussed the ontology service so we will want to address that before this PR is ready for review.

trpcultivate_phenotypes/trpcultivate_phenotypes.services.yml
```diff
@@ -0,0 +1,14 @@
...
+  trpcultivate_phenotypes.terms:
+    class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTermsService'
+    arguments: ['@config.factory']
+  trpcultivate_phenotypes.ontology:
+    class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesOntologyService'
+    arguments: ['@config.factory']
+  trpcultivate_phenotypes.vocabulary:
+    class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesVocabularyService'
+  trpcultivate_phenotypes.database:
+    class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesDatabaseService'
```
laceysanderson, line 13: To use dependency Injection in a service, we tell Drupal to grab the chado database connection from the container in the services.yml as follows. This allows the database connection to be a parameter of the __construct().

```suggestion
    class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesDatabaseService'
    arguments: ['@tripal_chado.database']
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesDatabaseService.php
```diff
@@ -0,0 +1,47 @@
...
+  
+  /**
+   * Chado Db.
+   */
+  protected $chado;
+
+  /**
+   * Constructor.
+   */
+  public function __construct() {
```
laceysanderson, line 23: Based on the change to the services.yml, the database connection is now available here:
```suggestion
  public function __construct(ChadoConnection $connection) {
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesDatabaseService.php
```diff
@@ -0,0 +1,47 @@
...
+    // Chado database.
+    $this->chado = \Drupal::service('tripal_chado.database');
```
laceysanderson, line 25:
```suggestion
    $this->chado = $connection;
```

laceysanderson:
This PR is superceeded by #32

Issue #12, opened 2023-06-08
G4.12 Terms Service for creating default terms on install


laceysanderson:
This service is the upgraded version of [Analyzed Phenotypes CV API](https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/7.x-3.x/api/analyzedphenotypes.cv.api.inc) with a focus on the new D10 functionality.

Service Name: TripalCultivatePhenotypesTermsService
Found here: trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTermsService.php

## Associated Config:

We will use a YAML file to define the terms rather than the old nested array structure. This has been done in Tripal Core here:

- [Tripal defines the schema](https://github.com/tripal/tripal/blob/4.x/tripal/config/schema/tripal.schema.yml#L128-L187)
- [Service to create the content types defined in the YML](https://github.com/tripal/tripal/blob/4.x/tripal/src/Services/TripalEntityTypeCollection.php)
- [Tripal Chado implementing that schema](https://github.com/tripal/tripal/blob/4.x/tripal_chado/config/install/tripal.tripalentitytype_collection.default_chado.yml)

## Methods:

**Note:** 

- current defineTerms() is replaced by the associated config above.
- current mapDefaultTermToConfig() will no longer be needed as we will have a term_key (i.e. experiment_year, experiment_location).
- current getTermConfigValue() becomes getTermId() below.
- current getTerm() will go away. We think we don't need it yet due to the autocomplete which was contributed to core. If we do need something like this, we need to design such a service in Tripal Core and then just use it in this module.

```php
/**
 * Insert cvterms (+ cvs as needed) into Chado based on YAML file 
 * and set term configuration variable defaults.
 * 
 * @return boolean
 *   True if all terms were inserted successfully and false otherwise.
 */
public function loadTerms();

/**
 * Retrieves the ID of the term configured for a specific role.
 * 
 * It is expected that the administrator can configure these terms. As such
 * this method will pull the value from configuration rather then look it up
 * in the database.
 * 
 * @param string $term_key
 *   The unique identifier for the term of interest. This should be one of:
 *   data_collector, genus, experiment_location, collection_method, experiment_container,
 *   unit_to_method_relationship_type, method_to_trait_relationship_type, 
 *   experiment_replicate, experiment_year.
 * @return integer
 *   The chado cvterm_id for the term associated with that key.
 */
public static function getTermId(string $term_key);
```

PR #13, opened 2023-06-12
G4.12 Term Service

reynoldtan:
Issue #12 
**Dependant on PRs #17 and #19.**

This PR sets up Term Service described in this Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/12

Included is a unit test that tests methods defineTerms() and getTermId(). I am unable to test loadTerms() as I get error regarding function that does not exist (chado_get_cvterm()).


laceysanderson (commented):
Adding terms during install has added this to every functional test you run. As such all functional tests will need a test chado... I added to the install test as an example.

Dougs recent PR #1537 removed the 'chado' default for the schema name from the chado_get_* commands. This broke your usage which implies you are not able to access config values for tripal_chado in the install of this module. As such, we should move this functionality into a Tripal job as I suggested which is added in the install.

NOTE: You'll want to remove the hardcoding of the 'chado' default once you do this. I only added it to prove the failing was related to this specific change.

reynoldtan:
Replaced the method to install terms during install from calling the method to load terms to creating a Tripal Job.

Here is the proof of concept :)
This is displayed after an install:
Screenshot 2023-06-27 at 15-22-59 reynold Tripal 4 on Drupal 9 DEVELOPMENT

This is to check that term got installed.
Screenshot 2023-06-27 at 15-30-47 Execute PHP code Tripal 4 on Drupal 9 DEVELOPMENT

reynoldtan:
Big sigh, I simplified the loadTerms method and reduced the use of Tripal 3 api down to just chado_insert_cvterm()

laceysanderson (approved):
I made the following changes:
- Removed the assumption that the default chado was named `chado` by setting the default schema name to null so chado_insert_cvterm() will look up the default if we don't know it.
- Ensured your kernel test extended ChadoKernelTestBase and associated changes
- Your tests were using the production chado database... Instead I created a test chado instance, set the container and ensured that it was being used.

After these changes, the code looks good ✅ 

**NOTE: As mentioned in G4.14 you never add a message when using asserts. Best practice is to always pass in a message describing the test case to make it more readable when a test fails.**

For example,
```php
$is_saved = $this->service->loadGenusOntology();
$this->assertTrue($is_saved);
```
Will just say that an assert failed because False is not True and then give you a line number to dig up.

Instead you should do,
```php
$is_saved = $this->service->loadGenusOntology();
$this->assertTrue($is_saved, 
  "Unable to save a default value for all genus as loadGenusOntology() did not return TRUE.");
```

**Despite this, I'm approving this PR so we can keep moving but please always pass in a message to asserts in the future.**

Issue #14, opened 2023-06-13
G4.14 (G0.03.2) Service for managing CV and Cvterm configuration values


reynoldtan:
This service is the upgraded version of [Analyzed Phenotypes Ontology API](https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/7.x-3.x/api/analyzedphenotypes.ontology.api.inc) with a focus on the new D10 functionality.

Service Name: TripalCultivatePhenotypesGenusOntologyService
Found: trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesOntologyService.php

## Associated Config:
Schema:

Genus:
  -**trait** // Trait Vocabulary
  -**method** // Method Vocabulary
  -**unit** // Unit Vocabulary
  -**database** // Database id
  -**crop_ontology** // Crop Ontology Vocabulary

allownew // True or False, Allow new traits to be added during upload 
  
Access to configuration values will be based on the genus.

## Methods

```
/**
 * Fetch all genus from chado.organism in the host site and construct a genus ontology
 * configuration values described above. Each genus will contain a configuration value 
 * for trait+unit+method, database and crop ontology.
 *
 * @return array
 *    Associative array where each element is keyed by genus and configuration values array
 *    as the value.
 */
public function defineGenusOntology()
```

```
/**
 * Register a configuration entry and set each genus ontology configuration values to
 * a default value of 0 (not set).
 *
 * @return boolean
 *   True all genus ontology configuration created and set a default value, False on error.
 */
public function loadGenusOntology()
```
  
```
/**
 * Get genus ontology configuration values.
 *
 * @param string $genus
 *    Genus
 *
 * @return array
 *    Associated configuration values trait, unit, method, database and crop ontology.
 */

public function getGenusOntologyConfigValue($genus)
```

```
/**
 * Prepare genus text to be used as variable name by removing any formatting and converting spaces to underscore.
 *
 * @param string $genus
 *   Genus
 *
 * @return string
 *    Lowercase version and all spaces replaced by underscore.
 */

public function formatGenus($genus)
```

PR #15, opened 2023-06-15
G4.14 Genus Ontology Service

reynoldtan:
This PR sets up Genus Ontology Service described in this Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/14

With the addition of a method to save configuration values coming from a form submit.

```
/**
 * Get genus ontology configuration values.
 *
 * @param string $genus
 *   Genus
 *
 * @return array
 *   Associated genus configuration values trait, unit, method, database and crop ontology.
 */
public function getGenusOntologyConfigValues($genus)
```

Included is a kernel test that tests all methods of the service.

### DIFF
--- a/trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
+++ b/trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
@@ -42,10 +42,6 @@ trpcultivate:
         - '/ = div'
         - '? = unsure'
         - '- = to'
-    
-    ontology:
-      cvdbon:
-      allownew: true
 
     # trpcultivate.phenotypes.watermark:
     # Configurations relating to the watermarking of charts.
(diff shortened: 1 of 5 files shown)


laceysanderson (approved):
Service Code:
 - It was a bit confusing how you used the genus as the key but after looking into it, I did confirm that Drupal does support non-integer keys in config sequences 🤷‍♀️ So I added a description to the config to make that clear since it is a pretty big assumption.
- ✅ **Code looks good: no changes needed**

Kernel Test:

- Originally the kernel tests were failing in Drupal 10.0 and 10.1 due to a more specific signature for the setUp() method.
- Additionally, you used the Drupal Kernel base when you should have used the ChadoKernelTestBase.
- I updated your tests with these two changes so they now pass on all versions of Drupal.

**One thing I did notice in your tests but did not fix was that you never add a message when using asserts. Best practice is to always pass in a message describing the test case to make it more readable when a test fails.**

For example, 
```php
$is_saved = $this->service->loadGenusOntology();
$this->assertTrue($is_saved);
```
Will just say that an assert failed because False is not True and then give you a line number to dig up.

Instead you should do,
```php
$is_saved = $this->service->loadGenusOntology();
$this->assertTrue($is_saved, 
  "Unable to save a default value for all genus as loadGenusOntology() did not return TRUE.");
```

**Despite this, I'm approving this PR so we can keep moving but please always pass in a message to asserts in the future.**

Issue #10, opened 2023-06-01
G2.10 Data Importer Importer + Validator API


reynoldtan:
Create importer + validator api for phenotypes collect and phenotypes share.

laceysanderson:
This sounds like it should be multiple branch/PRs. We want to keep these as small as possible. For example, only the importer class outline for one of the modules with the form defined and routes to support should be a single PR.

Can you come up with a full design here in the issue including how you intend to break this up into reviewable chunks?

Some questions you may want to answer/think about:

- Are you still thinking phonocollect + phenoshare will share a form or did we manage to convince you in the office they should be two separate forms? 
- What sort of things do you want to validate? 
- Are you planning a multi-stage form like we had for analyzed phenotypes? 
- Are you supporting the same formats as before? 
- How do you see the raw phenotypes changing to support the new trait-method-unit approach?

laceysanderson:
Closing in favour of smaller issues/PRs.

PR #16, opened 2023-06-21
G2.10 Importer API

reynoldtan:
Screen Shot 2023-06-21 at 10 48 14 AM

Issue #10

This PR contains an instance of Tripal Importer, stylesheet and Javascript required to construct the form shown above. The functional test included test the existence of elements in the page and navigating to each stages (3 stages in this setup).

This will be the base layout and logic for switching pages for data importer.


laceysanderson:
Oh no... I hadn't reviewed this one yet 😬 

I will still review and test it here since at least I can still see the changes, etc but I guess we'll have to do another PR to fix any changes neede (hopefully there are none needed 🙏 )

trpcultivate_phenoshare/trpcultivate_phenoshare.module
```diff
@@ -27,3 +27,24 @@ function trpcultivate_phenoshare_help($route_name, RouteMatchInterface $route_ma
...
+ */
+function trpcultivate_phenoshare_form_alter(&$form, &$form_state, $form_id) {
+  // This hook will omit from Tripal Importer Plugin.
+  // - Field file upload.
+  // - Submit button.
+  // - Advanced Option (Database options).
+  if(isset($form['importer_plugin_id']['#value']) 
+    && $form['importer_plugin_id']['#value'] == 'trpcultivate-phenotypes-share') {
+      // Disable selection of previously uploaded file. 
+      $form['file']['file_upload_existing'] = [];
```
laceysanderson, line 42: This should not be needed as specifying FALSE for file_upload, file_load, and file_remote should be enough. If that is not enough then you need to file a bug with Tripal core, not alter the form.

trpcultivate_phenoshare/trpcultivate_phenoshare.module
```diff
@@ -27,3 +27,24 @@ function trpcultivate_phenoshare_help($route_name, RouteMatchInterface $route_ma
...
+      // Form will implement a form button that can be re-named
+      // relevant to each stage.
+      $form['button'] = [];
```
laceysanderson, line 45: Again, this should not be needed as setting use_button to FALSE should be enough. If that is not enough then you need to file a bug with Tripal core, not alter the form.

trpcultivate_phenoshare/trpcultivate_phenoshare.module
```diff
@@ -27,3 +27,24 @@ function trpcultivate_phenoshare_help($route_name, RouteMatchInterface $route_ma
...
+      // Exclude the database option (advanced options) since phenotypes
+      // schema is always installed in chado database (default schema).
+      $form['advanced']['#access'] = FALSE;
```
laceysanderson, line 48: It is never right to exclude the option for the user to choose the Chado schema. Please do not ever disable this. Rather you will want to work it into your Stage 1. You can theme it to blend in more if you want but it should not be removed.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesExperimentService.php
```diff
@@ -0,0 +1,47 @@
...
+ * @file
+ * Tripal Cultivate Phenotypes Experiments/Projects service definition.
+ */
+
+namespace Drupal\trpcultivate_phenotypes\Service;
+
+/**
+ * Class TripalCultivatePhenotypesExperimentService.
+ */
+class TripalCultivatePhenotypesExperimentService {
```
laceysanderson, line 13: Why do we need this service? Shouldn't experiments be entered using the project autocomplete you made for Tripal core? Please either remove this or actually define what services will be used for in the file docblock.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesExperimentService.php
```diff
@@ -0,0 +1,47 @@
...
+   * Chado Db.
+   */
+  protected $chado;
+
+  /**
+   * Constructor.
+   */
+  public function __construct() {
+    // Chado database.
+    $this->chado = \Drupal::service('tripal_chado.database');
```
laceysanderson, line 25: This should use dependency injection to get the database if you determine this service is still needed.

trpcultivate_phenotypes/trpcultivate_phenotypes.libraries.yml
```diff
@@ -0,0 +1,24 @@
+# Tripal Cultivate Phenotypes Libraries.
+
+# Script to auto-select field content for quick alteration of value.
+script-autoselect-field:
```
laceysanderson, line 4: Maybe we should prefix these with `tcp` or `trpcultivatepheno` instead of `script` to make their names more unique?

laceysanderson (commented):
When I click the "Need help?" link, nothing seems to happen... is this expected at this point? I'm guessing this is not supposed to happen based on the following errors and the existence of the script-pull-window.js...

In the web developer console on Firefox I see the following errors on loading of the importer without interacting with anything:
```
Uncaught TypeError: buttonToggle.once is not a function
    attach http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:4103
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:206
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:202
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:728
    listener http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:716
    domReady http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:722
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:727
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:730

Uncaught TypeError: buttonToggle.once is not a function
    attach http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:4103
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:206
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:202
    bigPipeProcessDocument http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7172
    bigPipeProcessDocument http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7171
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7209
    EventListener.handleEvent* http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7205
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7211

Uncaught TypeError: buttonToggle.once is not a function
    attach http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:4103
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:206
    attachBehaviors http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:202
    insert http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:3024
    each http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2
    each http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2
    insert http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:3022
    commandExecutionQueue http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2696
    promise callback*Drupal.Ajax.prototype.commandExecutionQueue/< http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2689
    commandExecutionQueue http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2686
    success http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:2743
    bigPipeProcessPlaceholderReplacement http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7140
    bigPipeProcessDocument http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7179
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7209
    EventListener.handleEvent* http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7205
    <anonymous> http://localhost/sites/default/files/js/js_MMpdWWmnr4S2p7hsHGBGh8UM-oI0DJL7c7N_zu2MrDA.js:7211

```

Additionally, I updated the importer title and description in a commit on the original branch (`g0.04-Importer-api`). Please make changes and fixes to the original branch as well and when done open a new PR for it.

trpcultivate_phenoshare/trpcultivate_phenoshare.module
reynoldtan: setting it to FALSE would suffice 👍

trpcultivate_phenoshare/trpcultivate_phenoshare.module
reynoldtan: setting it to FALSE works 👍

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesExperimentService.php
reynoldtan: No longer required and removed, will be replaced by Tripal Autocomplete Project

trpcultivate_phenotypes/trpcultivate_phenotypes.libraries.yml
reynoldtan: Renamed accordingly

reynoldtan:
This should work now, I forgot to include dependencies for this script library

trpcultivate_phenoshare/trpcultivate_phenoshare.module
reynoldtan: My goal is to keep the importer form as simple as possible, no advanced configurations or other unnecessary field elements. If this module is closely tied to chado I think, there shouldn't be the need for this field... but then maybe I am missing a detail if we could discuss more about this. The current PR includes this advanced option

laceysanderson: This field doesn't allow them to select something other than chado. Rather it lets them select which install of chado they want to use. 

For example, a site may decide to have a chado install for private data (named `private_biodata`) and a separate chado install for public data (named `public_biodata`). In this example, a curator would choose from this drop down either `private_biodata` or `public_biodata` based on whether the researchers ability to release the data. Both `private_biodata` and `public_biodata` would be Chado schema with all the same tables.

Another example might be having one chado install for rough or in-progress data and another for published data only. This way they could use another module or postgresql itself to ensure the published data is read-only once inserted.

While this module doesn't need to support any of this advanced functionality, it needs to allow the Importer API to add things like this to the form so that others can implement it. You just need to make sure you do not unset anything from the form created by the importer API. Instead you can keep the simplicity of this form by themeing the advanced fieldset to be collapsed by default and less conspicuous.

Issue #20, opened 2023-06-23
G4.01 Service to associate genus with project


reynoldtan:
This service is the upgraded version of [Analyzed Phenotypes Genus API](https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/7.x-3.x/api/analyzedphenotypes.genus.api.inc) with a focus on the new D10 functionality.

Service will handle setting of genus to project and resolving genus of a project. Mapping of genus is done using projectprop table where type_id is the configuration value of the term genus.

Service Name: TripalCultivatePhenotypesGenusProjectService
Found: trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php

## Associated Config:
trpcultivate.phenotypes.ontology.terms.**genus**

```
/**
 * Assign a genus to an experiment/project.
 * 
 * @param int $project
 *   Project (project id number) the parameter $genus will be assigned to.
 * @param string $genus
 *   Genus name/title.
 * @param boolean $replace
 *   True to replace existing genus of a project with a different genus.
 *   Default to False.
 *
 * @return boolean
 *   True, genus was set successfully or false on error/fail.
 */
function setGenusToProject($project, $genus, $replace = FALSE)
```

```
/**
 * Get genus of an experiment/project.
 * 
 * @param int $project
 *   Project (project_id number) to search.
 *
 * @return array
 *   Keys is genus/organism id number and value is the genus name/title.
 *    
 */
public function getGenusOfProject($project)
```

```
/**
 * Get all genus that have been configured (traits, unit, method, database and crop ontology).
 *
 * @return array
 *   An array of genus names, sorted alphabetically.
 */
public function getActiveGenus() 
```

PR #28, opened 2023-06-30
G4.20 genus project service

reynoldtan:
**Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/10** - [Genus Project Service](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/20)

## Motivation
This service is the upgraded version of [Analyzed Phenotypes Genus API](https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/7.x-3.x/api/analyzedphenotypes.genus.api.inc) with a focus on the new D10 functionality.

## What does this PR do?
Service will handle setting of genus to project and resolving genus of a project. Mapping of genus is done using projectprop table where type_id is the configuration value of the term genus.

### Automated Testing

ClassName->ServiceGenusProjectTest: 
- Create test records in chado.organism and chado.project.
- Create relationship record in chado.projectprop between genus and project.
- Create genus configuration entry (genus: trait, unit, method, db and crop ontology).
- Tests that service can retrieve active genus - genus that have been configured.
- Test to retrieve the genus of a project.
- Test to replace/update a project with genus.

### Manual Testing

1. Install and enable Tripal Cultivate Phenotypes module and switch branch to g4.01-Genus-Project-Service
Note: make sure to have a test organism and project records and an entry in projectprop table (of type the term in configuration corresponding to genus) relating a genus to a project, in your instance.
2. Configure terms and ontology of the module and create a genus-ontology configuration entry.
3. Using php in your host/devel/php
4. test: 
  $service = \Drupal::service('trpcultivate_phenotypes.genus_project');
  $service->getActiveGenus(); // Should list all genus that have been configured.
  $service->getGenusOfProject(project id number); // Should return a genus of a project.
  $service->setGenusToProject(project id number, genus); // Should set a genus to a project.


trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        // Do nothing if maintain the same genus.
+      } 
+      else {
+        // Not set yet, no record in projectprop.
+        // Create a relationship regardless to replace or not.
+        $sql = "INSERT INTO {1:projectprop} (project_id, type_id, value) VALUES ('%s', '%s', '%s')";
+        $query = sprintf($sql, $project, $this->sysvar_genus, $genus);
+        $this->chado->query($query);
```
laceysanderson: You should use placeholders here instead of sprintf.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -0,0 +1,154 @@
...
+      }
+    }
+    else {
+      $error = 1;
+      $this->logger->error('Error. Could not replace genus. Invalid project or genus');
```
laceysanderson: It would be more helpful here to be more specific -check whether the problem is the project or genus.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  /**
+   * Get genus of an experiment/project.
+   * 
+   * @param int $project
+   *   Project (project_id number) to search.
+   *
+   * @return array
+   *   Key is genus/organism id number and value is the genus name/title.
```
laceysanderson, line 126: This should only return the genus. That is because there may be more than one organism with the same genus. Here you are arbitrarily taking the first organism entry that postgresql gives you 😬

laceysanderson: Additionally, for performance reasons I think we select directly from project prop without confirming the genus exists or is configured.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -0,0 +1,154 @@
...
+    return ($genus_project) ? ['id' => $genus_project->id, 'genus' => $genus_project->genus] : 0;
+  }
+
+  /**
+   * Get all genus that have been configured (traits, unit, method, database and crop ontology).
+   *
+   * @return array
+   *   An array of genus names, sorted alphabetically.
+   */
+  public function getActiveGenus() {
```
laceysanderson: We should not be touching the genus ontology configuration in this service as that is the goal of trpcultivate_phenotypes.genus_ontology. Instead you should have trpcultivate_phenotypes.genus_ontology as a dependancy of this service and use it here.

For example, if the `trpcultivate_phenotypes.genus_ontology` is stored in `$this->genusOntologyService` then this method might be:

```php
$genus = $this->genusOntologyService->getConfiguredGenus();
return array_values($genus);
```

Then you would add the following to the GenusOntologyService:

```php
/**
 * Retrieve a list of all genus for which configured values have been registered.
 * 
 * @return array
 *   An array where the key is the formatted genus and the value is the 
 *   original one from Chado.
 */
public function getConfiguredGenus();

/**
 * Retrieve the original genus as formatted in Chado.
 * 
 * @param $formatted_genus
 *   The genus as formatted by formatGenus().
 * 
 * @return string
 *   The genus in Chado that this genus maps to.
 */
public function deformatGenus($formatted_genus);
```

I would also suggest adding to the genusOntologyService a protected variable which stores the mapping between formatted and original which is set by defineGenusOntology().

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -0,0 +1,208 @@
...
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Set test environment.
+    \Drupal::state()->set('is_a_test_environment', TRUE);
+
+    // Install module configuration.
+    $this->installConfig(['trpcultivate_phenotypes']);
```
laceysanderson, line 74: OPEN THOUGHTS: We may want to discuss when to mock the config (as we did in Carolyns PR) and when to install it as you did here.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -0,0 +1,208 @@
...
+    // Test Chado database.
+    // Create a test chado instance and then set it in the container for use by our service.
+    $this->chado = $this->createTestSchema(ChadoTestKernelBase::PREPARE_TEST_CHADO);
+    $this->container->set('tripal_chado.database', $this->chado);
+
+    // Prepare by adding test records to genus, project and projectproperty
+    // to relate a genus to a project.
+    $project = 'Project - ' . uniqid();
+    $sql = "INSERT INTO {1:project} (name, description) VALUES('%s', '%s')";
+    $query = sprintf($sql, $project, $project . ': Description');
```
laceysanderson: Never use sprintf for queries. We should instead used named placeholders.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -0,0 +1,208 @@
...
+    $sql = "INSERT INTO {1:organism} (genus, species, type_id) VALUES('%s', '%s', '%s')";
+    $query = sprintf($sql, $genus, 'Wild Species', 1); // Adding as null organism type.
+    $this->chado->query($query);
+    
+    $genus_id =$this->chado->query(
+      "SELECT organism_id FROM {1:organism} WHERE genus = :genus LIMIT 1", [':genus' => $genus]
+    )
+      ->fetchField();
+    
+    $this->ins['genus_id'] = $genus_id;
```
laceysanderson, line 106: Also, if you use the insert query builder it will return the id of the record it creates when executed.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -0,0 +1,208 @@
...
+      ]
+    ];
+
+    $this->config->set('trpcultivate.phenotypes.ontology.cvdbon', $genus_ontology_config);
+
+    // Term service.
+    $this->service = \Drupal::service('trpcultivate_phenotypes.genus_project');
+  }
+
+  public function testGenusProjectService() {
```
laceysanderson, line 166: It would also be good in these tests to select from the project prop table directly to ensure the record is where we expect.

laceysanderson (changes requested):
I've done a code review of this PR and left comments accordingly.
Can you also update the PR description with instructions for manual testing? Perhaps use the new template as guidance even though this PR was created before the template.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
reynoldtan: modified query to use placeholder text 👍

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
reynoldtan: modified method to be specific to which parameter is invalid 👍

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
reynoldtan: The query already restricts genus based on the genus set in projectprop. I have updated the query to only consider genus that are configured.  Similary, I added check in the set genus method to also check the genus (that it is configured) before pairing it to a project. Both measures will improve accuracy of returned genus and prevent inactive genus from being used.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
reynoldtan: makes perfect sense, I have evaluated the purpose of the method and I think that it can be omitted since it will be used mainly for populating select field and as discussed and agreed that for populating select field we could just straight up write a query or if Tripal 4 provides.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
reynoldtan: as we did in Carolyns PR?? please share. I use the install if test needs to write to a configuration (dunno how to do this with mock) if its only to retrieve configuration values I go ahead with mocking it. If the test offers the install I prefer to use it over mocking considering my config has gotten complicated at this stage. These are my thoughts.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
reynoldtan: noted

laceysanderson (approved):
✅ Code review comments were addressed
✅ Functional testing shows the service works as expected!

Issue #6, opened 2023-04-17
G0.06 Configuration Form for Ontology and Terms


reynoldtan:
This configuration will define vocabulary terms and controlled vocabulary terms used to house traits and by many processes in the phenotypes module.

![Ontology Terms - Terms](https://user-images.githubusercontent.com/15472253/232533132-3129c06c-9bbe-47b5-bbe5-dac1cea27d90.jpg)
![Ontology Terms](https://user-images.githubusercontent.com/15472253/232533208-c802d85b-94d8-4f08-bcba-373c3e9cc787.jpg)

laceysanderson:
Merged

PR #32, opened 2023-07-07
G0.06 configuration ontology and terms

reynoldtan:
**Issue #[6] G0.06 Configuration Form for Ontology and Terms**
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/6
## Motivation

Create genus ontology and terms configuration page that will define vocabulary terms and controlled vocabulary terms used to house traits and by many processes in the phenotypes module.

## What does this PR do?

1. Create configuration page
2. Functional test that tests page access permission, existence of fields in the page/form, set value to field and submit the form and posts warning message when Tripal Job to install terms and genus ontology was not executed.

### Automated Testing

1. Asserts page is accessible with administrative/tripal access permission.
2. Asserts page loads with response status code of 200.
3. Asserts page posts warning when terms and genus ontology are not installed through Tripal job setup during install.
4. Asserts page load when terms and genus ontology are installed (manually created).
5. Asserts for all sample genus and terms have correct field.
6. Asserts setting a value to all fields for all sample genus and terms and submitting the form, that values set are saved.
7. Asserts that retrieving configuration values for all sample genus and terms match the value set.
8. Asserts allow new traits field saves true or false values.

- 
- 
- 

### Manual Testing

1. Install Tripal Cultivate Phenotypes - trpcultivate_phenotypes module.
2. If your build has no genus and accessing the configuration page - my site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology will instruct to execute tripal job. A link to to tripal job page and the job title are encoded into the warning message.
3. Create sample Genus and execute the Tripal Job mentioned.
4. Clear cache and refresh Genus Ontology and Terms configuration page.


trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
```diff
@@ -92,6 +92,7 @@ trpcultivate:
             id: TAXRANK:0000005
             name: genus
             definition: 'The genus'
```
laceysanderson, line 94: Sorry, I missed this in a previous PR. These definitions should be the term definition as defined by that ontology.

trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
```diff
@@ -92,6 +92,7 @@ trpcultivate:
...
+            field_label: 'This term indicates that a given property is the associated "genus". For example, this module save experiments producing phenotypic data as projects and the organism that experiment assays is indicated by adding a "genus" property to it.'
```
laceysanderson: This key makes me think of a title/label. Instead maybe we should call it `help_text` as that's what it appears to be?

trpcultivate_phenotypes/templates/trpcultivate-phenotypes-template-header-instructions.html.twig
```diff
@@ -0,0 +1,69 @@
...
+ * Theme ontology configuration header instructions.
+ * 
+ * Available variables:
+ * - data.section: which section of the page/form to render the header instruction.
+ * - data.link_01: internal/external link number 1. (embed a link to instruction).
+ * - data.link_02: internal/external link number 2. (embed a link to instruction).
+ */
+#}
+
+{% if data.section == 'ontology' %}
```
laceysanderson, line 13: The new twig formats want us to limit control structures and calculation as much as possible. I would be tempted to make two different twig templates: 1) ontology header, 2) terms header.

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
```diff
@@ -0,0 +1,161 @@
...
+    ];
+
+    $ins_genus = "
+      INSERT INTO {1:organism} (genus, species, type_id)
+      VALUES
+        ('$test_insert_genus[0]', 'culinaris', 1),
+        ('$test_insert_genus[1]', 'arientinum', 1)
+    ";
+
+    $this->chado->query($ins_genus);
```
laceysanderson: I would not use chado->query() for inserts but rather the query builder always. That said, if you do keep chado query then definitely used named placeholders and arguments --not concatenate in values directly.

laceysanderson: I would also suggest adding two organisms with the same genus and one without for this test so you catch the case shown in my functional review/test.

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
```diff
@@ -0,0 +1,161 @@
...
+    $values_genusontology = [];
+    
+    foreach($genus_ontology as $genus => $vars) {
+      foreach($vars as $i => $config) {
+        $fld_name = $genus . '_' . $config;
+        // Test if each genus has a trait, unit, method, db and crop ontology field.
+        $session->fieldExists($fld_name);
+
+        // set to Null (id: 1) all genus ontology configuration.
+        $values_genusontology[ $fld_name ] = $null_value;
```
laceysanderson: I would set each of these to different values -incremental is easiest.
With grid forms such as these it's easy to not uniquely be able to specify a particular cell so we want to confirm here that there is no cross cell interaction.

I also like to do a quick submit with no values filled in to see if that has errors as sometimes we assume the user would have made a change and they may not.

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
```diff
@@ -0,0 +1,161 @@
...
+    // Terms.
+    $null_value = 'null (null:local:null)';
+    $terms = $service_terms->defineTerms();
+    $values_terms = [];
+
+    foreach($terms as $config => $prop) {
+      // Test each term has an autocomplete field.
+      $session->fieldExists($config);
+
+      $values_terms[ $config ] = $null_value;
```
laceysanderson: Same goes for this one (i.e. do not set them all to the same thing) although it's less important in a list then it is in a grid.

laceysanderson (changes requested):
## Functional Review/Test

### Confusing warning

Screenshot 2023-07-18 at 1 51 22 PM
This message has a link that sends me to the job page but the job page does not tell me how to execute a job (Tripal core's fault 🙈) and there is no longer an execute button on this page. Should we change this message to tell them the command to run the job and to run it on the command-line?

Hmmm... we need separate messages for detecting when the job needs to be run and when there are no organisms. I ran the job and then refreshed the page to be presented with the same message and same link to the job which is showing complete. The second message should indicate no genus in the default chado and link to the organism add page.

### Doesn't support multiple organisms of the same genus

I then entered three organisms as follows:
 - Tripalus databasica
 - Tripalus postgresquelus
 - Lens examplus

This resulted in. large number of errors as shown in the linked screenshot and displayed an element for Lens but not for Tripalus.

[Large Screenshot showing errors](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/1566301/f6bfe51d-8e97-4a1b-a809-2016b6ae9f97)

### Crop ontology should not be required

Screenshot 2023-07-18 at 3 29 44 PM

The crop ontology should not be required.

## Code Review

I reviewed all files except for the form itself. I would like to see the problems in the functional review addressed before I code review the form.

trpcultivate_phenotypes/src/Form/TripalCultivatePhenotypesOntologySettingsForm.php

All comments for code review are in-line comments.

trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
reynoldtan: 👍 Let's have a discussion about the terms, maybe go over them one by one as I added a few (relating to raw) into the list. The definition for this one I got from AP https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/e21d7f0ba1a53a6364886fc403ed90f025603c79/api/analyzedphenotypes.ontology.api.inc#L62

trpcultivate_phenotypes/templates/trpcultivate-phenotypes-template-header-instructions.html.twig
reynoldtan: I agree completely but it is very very very limited to just 1 sad lonely if else 😁 and both are related to header in the same page.

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
reynoldtan: for colab please... need more info

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
reynoldtan: Used generated terms in the test schema.

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
reynoldtan: same above, used generated terms.

trpcultivate_phenotypes/config/install/trpcultivate_phenotypes.settings.yml
reynoldtan: yes help_text for the field 👍

reynoldtan:
I have modified warning message into 2 levels where genus is top priority and when resolved will instruct user to execute Tripal Job. Please see updated message in the screen captures below.

No genus/organism detected in Tripal instance
no-genus

Genus requirements have been resolved
with-genus

I did not get a lengthy error message with the 3 test organism in your review. Please try again 🤞 . Crop ontology is now optional field in the recent commit.

form

trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
reynoldtan: I added another organism with a shared genus 👍

laceysanderson (approved):
✅ Changes discussed in previous review were implemented and functionality works.

However, I was able to reproduce the error mentioned before... I added an error message for the user once I determined how it was being triggered. I am approving this PR and posting the bug in a separate issue #35 for follow-up there :-)

We will also follow up on the terms in another issue.

Issue #35, opened 2023-08-10
[Bug] Unable to find configuration for genus added after terms installed + form loaded once before


laceysanderson:
### Drupal Version

10.0

### Tripal Version

4.x dev (most recent commit)

### Module Version

4.x dev (most recent commit)

### Site Operating System

Debian Linux

### Installation Method

Docker image built from this module

### Browsers

Firefox

### Observations

I receive the following error (indicating that `$this->service_genusontology->getGenusOntologyConfigValues($genus);` returned NULL when trying to configure multiple genus. This seems to be dependant on the process by which the organisms were added.

![image](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/1566301/6561bced-e6e7-48a9-8ea8-bf201c481caa)

### Expected Behaviour

I expect to be able to configure all my genus' ;-p

### Steps to Reproduce

1. Create a docker image and container from this module. This was first observed in PR #32 but that one will soon be merged.

Execute the following from within a cloned version of this repository:
```
docker build --tag=trpcultivate-pheno:g0.06 ./
docker run --publish=80:80 -tid --name=phenog0.06 --volume=`pwd`:/var/www/drupal9/web/modules/contrib/TripalCultivate-Phenotypes trpcultivate-pheno:g0.06
docker exec phenog0.06 service postgresql restart
```

2. Go to the configuration form: http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
3. Create an organism (http://localhost/tripal/add/organism)
4. Go back to the [configuration form](http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology). You will be asked to load the terms for this module by executing the tripal jobs -do so using `docker exec phenog0.06 drush trp-run-jobs --username="drupaladmin"`
5. Once the terms are loaded, go back to the [configuration form](http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology) -you should see a row in the configuration form for the genus you added ✅ this is good :-) 
6. Now go create a new organism with a different genus that the original one.
7. Return to the [configuration form](http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology) and you should get an error that it could not find configuration for the second genus you created.

laceysanderson:
Fixed by merged PR

PR #36, opened 2023-08-14
Fixes missing genus post-install

reynoldtan:
[**Issue #[35]**](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/35) BUG: genus missing past install.

## What does this PR do?

1. This  PR fixes a bug in configuration for genus ontology form where inserted/created organism past install is missing from the list of genus.

### Automated Testing

### Manual Testing

1. Create organism
2. Install Tripal Cultivate Phenotypes module
3. Execute job requirements for Term and genus ontology install
4. Access genus ontology configuration form [ my tripal site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology ]
At this point, it will only show the genus created in step 1.
5. Insert or create another organism
6. Refresh configuration page to update genus list with the newly inserted genus.


laceysanderson (approved):
Looks good to me 👍

Issue #37, opened 2023-10-11
G4.02 Validation API


reynoldtan:
### Branch

g4.02-Validation-API

### Groups

Group 4 - API | Services | Plugins

### Describe

This is a Drupal Plugin-based validation API that will be used to validate imported data file in all levels: PROJECT, GENUS, FILE, TRAITS and VALUES.

### Design

In contrast to Tripal 3 and Drupal 7 version where the validation API is hook-based, this version will implement a Drupal 9 Plugin for each validation levels. Components are as follows:

1. Plugin manager and plugin manager class.
2. A service for the plugin manager
3. A validator plugin base class with methods to:
  a. loadAssets - this will initialize class properties pertaining to project, genus and file id. All validator plugin instance will have access to these values.
  b. getValidatorName - will return the name of the validator plugin as set in the plugin annotation definition.
  c. getValidatorScope - will return the scope/level the validation is required. Scope can be PROJECT, GENUS, DATA_FILE, TRAITS or VALUES.
4. Plugins titled with the level it is required to validate (ie. Project for PROJECT validator) will be a class that extends the validator base class and will have a validate() method that contains the validation logic required. 

Plugin factory will be in the formValidate() of the Importer.
This is an example of executing Projects validator plugin.
```
$scope = 'PROJECT'; 
$manager = \Drupal::service('plugin.manager.trpcultivate_validator');
$plugins = $manager->getDefinitions();
$plugin_definitions = array_values($plugins);

// Get Project plugin validator definitions.
$plugin_key = array_search($scope, array_column($plugin_definitions, 'validator_scope'));
$validator = $plugin_definitions[ $plugin_key ]['id'];
$instance = $manager->createInstance($validator);

$instance->loadAssets($project, $genus, $file);
$validation_result[ $scope ] = $instance->validate();
...
```

PR #38, opened 2023-10-12
G4.37 validation api

reynoldtan:
**Issue #37** Validation API

## Motivation

This is a Drupal Plugin-based validation API that will be used to validate imported data file in all levels: PROJECT, GENUS, FILE, TRAITS and VALUES.

For this PR it is confined to validating project (PROJECT level) only.

## What does this PR do?

1. Validate project as entered in stage 1 of the importer for empty field, non-existent project and project that does not have a genus paired to.
2. Setup a Drupal Plugin type (validator), plugin factory (in the formValidate() method) and an instance in /Plugin/validator/Project.php that performs validation mentioned in item 1.
3. Store validation result using Drupal Storage system and deliver to validation result window in stage 1
4. Kernel test - tests plugin instance and perform validation of project.
5. Functional test - tests the Importer can navigate from one stage to another with pertinent stage elements enabled/disabled/expanded based on validation result.

### Automated Testing

Functional Test:
1. Navigate from one stage to the next. A pre-configured project and genus information used to trigger a passing validation result to allow from stage 1 to the next stage.
2. Caching of stage is verified in every stage.
3. Stage is correctly styled to indicate the current stage, completed stage or upcoming stage.

Kernel Test:
1. Create instance of the plugin for validating Project.
2. Test the plugin for empty project value, non-existent project and project without genus configured.

### Manual Testing

1. Create test genus record (ie. Lens).
2. Install module and execute Tripal Job required.
3. Configure genus ontology in my site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
4. Create test project records. One with genus set and the other without.
```
in my site/devel/php

// Set the genus of the project. Use configured genus as the method 
// has checks to make sure only configured genus can be paired to a project.
$service = \Drupal::service('trpcultivate_phenotypes.genus_project');
$service->setGenusToProject(PROJECT ID, 'Lens');

```
6. Load page my site/admin/tripal/loaders/trpcultivate-phenotypes-share
and test Project/Experiment field with:

A. Empty project, non-existent project and project without genus set
B. Project with genus set.

A - should trigger validation error and keep you in the same stage (stage 1)
B - should take you to the next stage and no validation error

### DIFF
--- a/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
+++ b/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
@@ -8,6 +8,8 @@
 namespace Drupal\trpcultivate_phenoshare\Plugin\TripalImporter;
 
 use Drupal\tripal_chado\TripalImporter\ChadoImporterBase;
+use Drupal\tripal_chado\Controller\ChadoCVTermAutocompleteController;
+
 
 /**
  * Tripal Cultivate Phenotypes - Share Importer.
@@ -43,6 +45,10 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase {
   // each stage accordingly (form, validation, etc.).
   private $current_stage = 'current_stage';
 
+  // Reference the validation result summary values in Drupal storage
+  // system using this variable.
+  private $validation_result = 'validation_result';
+
   // Headers required by this importer.
   private $headers = [
     'Header 1' => 'Header 1 Description',
@@ -72,6 +78,19 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase {
 
     // Cacheing of stage number:
     // Cache current stage and id field to allow script to reference this value.
+
+    // Account for failed validation.
+    // Refer to Drupal $storage system for validation result values saved.
+    $storage = $form_state->getStorage();
+    // Full validation result.
+
+    $validation_result = [];
+    $has_fail = FALSE;
+
+    if (isset($storage[ $this->validation_result ])) {
+      $has_fail = $this->hasFailedValidation($storage[ $this->validation_result ]); 
+    }
+
     $triggering_element = $form_state->getTriggeringElement();
     $valid_triggering_element = [
       'Validate Data File', // Stage 1
@@ -79,7 +98,7 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase {
       'Skip' // Stage 2
     ];
 
-    $stage = ($form_state->getValue('trigger_element') && in_array($triggering_element['#value'], $valid_triggering_element))
+    $stage = (!$has_fail && $form_state->getValue('trigger_element') && in_array($triggering_element['#value'], $valid_triggering_element))
       ? (int) $form_state->getValue( $this->current_stage ) + 1
       : 1;
     
@@ -95,7 +114,7 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase {
     // stage accordion layout. Each stage is a method titled stage + stage no (ie. stage1).
     $stage_methods = get_class_methods(get_class($this));
     $total_stages = 0;
-
+  
     foreach ($stage_methods as $method) {
       if (preg_match('/stage([1-9])/', $method, $matches)) {
         if ($stage_no = $matches[1]) {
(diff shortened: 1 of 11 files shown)


trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -293,6 +331,45 @@ public function formSubmit($form, &$form_state) {
    */
   public function formValidate($form, &$form_state) {
     $form_state_values = $form_state->getValues();
+    
+    // Current stage.
+    if ($stage = $form_state_values[ $this->current_stage ]) {
```
laceysanderson: What is your if meant to check here, exactly?

$stage has not been set yet so I assume it's not meant to be comparing that the $stage variable == the current stage in the form state values?

This may be better expanded for easier long term maintenance. Maybe something like the following (if I'm interpreting it right 😬 )

```php
// Get the cached state state from the form state values.
// Question: what is $this->current_stage? might be good to say in one sentence... like...
// Note: $this->current_stage is the name of the key used to cache the current stage.
if (array_key_exists($this->current_stage, $form_state_values)) {
  $stage = $form_state_values[ $this->current_stage ];
```

trpcultivate_phenotypes/src/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+namespace Drupal\trpcultivate_phenotypes;
+
+use Drupal\Component\Plugin\PluginInspectionInterface;
+
+/**
+ * Defines an interface for data validator plugin.
+ */
+interface TripalCultivatePhenotypesValidatorInterface extends PluginInspectionInterface {
```
laceysanderson, line 17: You should add `validate` to this interface.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -293,6 +331,45 @@ public function formSubmit($form, &$form_state) {
...
+        $project = $form_state_values['project'];
+        // @TODO: this will be a select field.
+        $genus = '#';
+        $file = $form_state_values['file_upload'];
+
+        if ($stage == 1) {
+          $scope = 'PROJECT';
+
+          $plugin_key = array_search($scope, array_column($plugin_definitions, 'validator_scope'));
+          $validator = $plugin_definitions[ $plugin_key ]['id'];
```
laceysanderson: I would actually add a method to your plugin manager:

```php
/**
 * Retrieve validator implementation with a specific scope.
 * 
 * @param string $scope
 *   The validator_scope you are interested in.
 * @return string
 *   The id of the validator with that scope based on it's annotation.
 */
public function getValidatorIdWithScope ( string $scope ) {

    $plugins = $this->getDefinitions();
    $plugin_definitions = array_values($plugins);

    $plugin_key = array_search(
      $scope, 
      array_column($plugin_definitions, 'validator_scope')
    );

    $validator = $plugin_definitions[ $plugin_key ]['id'];

    return $validator;
}
```

laceysanderson (dismissed):
✅ Functional review -worked exactly as expected! <3 
✅ Code Review is good! I had 3 small suggestions to improve but nothing was wrong or concerning :-) 

As such I am marking this approved although feel free to make the changes before merging if you want. I do not think I need to retest if you decide to add the changes :-)

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: revised the if condition and added more description/comments

trpcultivate_phenotypes/src/TripalCultivatePhenotypesValidatorInterface.php
reynoldtan: this is causing an error that I need to define the function in the validator base but the validate() method is local to individual validator plugin - Project.php in this case.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: added the method.. this is a good method 👍

laceysanderson (approved):
Code changes look good 👍 Still works as expected.

Issue #39, opened 2023-10-20
G2.39 - Data Collection Template File Generator 


reynoldtan:
### Branch

g2.02-Template-Generator

### Groups

Group 2 - Data Importing

### Describe

Create a data collection template file in tab-separated format (tsv). The headers encoded in the template file are the headers defined by the headers property in the importer instance.

Template file can be downloaded by clicking on the the link **Download a template file** located just below the list of headers of the importer.

### Design
Defined a route that will become the value of the href attribute of the download link mentioned in the Describe section.

```
trpcultivate_phenotypes.template_generator:
  path: '/tripalcultivate-phenotypes/download/template-file'
  defaults:
    _title: 'Download: Data Collection Template File'
    _controller: '\Drupal\trpcultivate_phenotypes\Controller\TripalCultivatePhenotypesTemplateFileController::generateFile'
  requirements:
    _permission: 'administer tripal'

```

![image](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/15472253/a333d4d7-db4f-4937-8d98-cc32f8455363)

The controller that handles the file will have the following components/logic:

1. Retrieve the configuration value relating to directory for storing template files.
2. Access headers defined in the importer. This set of headers will be transformed into tsv string (tab separated).
3. Create a file in the specified directory.
4. Write the headers into the file created in the previous step.
5. Start a download action for user to download the template file.

For discussion:
Passing the headers from the importer to the controller.
Options:
1. Using Routing parameter - query string can be hacked 
```
example.user:
  path: '/example/{user}'
```
3. Drupal Session Storage - is client-side and can be susceptible to XSS attack.
```
$session = \Drupal::service('session')->set('column_headers', $headers);
```
4. Drupal State Storage - for information about system state.
```
\Drupal::state()->set('column_headers', $headers);
```
5. Drupal Tempstore - Temporary buffer to save data.
```
\Drupal::service('tempstore.private')->set('column_headers', $headers);
```
6. Other

PR #40, opened 2023-10-25
G2.39 template generator

reynoldtan:
**Issue #39** Data Collection Template File Generator

## Motivation

Create a data collection template file in tab-separated format (tsv). The headers encoded in the template file are the headers defined by the headers property in the importer instance.

Template file can be downloaded by clicking on the the link Download a template file located just below the list of headers of the importer.

## What does this PR do?

1. Resolve the value for the href attribute of the Download template file link in the header listing of the importer.
image
2. The link is the generated template file by a service.
3. Headers defined in the Importer through the headers property are then encoded into the file as the headers column. Headers are converted to required format of tab-separated format. 

### Automated Testing

1. Functional tests included as part of the Importer API functional test and check if the template file was created.
2. Kernel tests test the template generator service to create a file.

### Manual Testing

1. Switch to branch g2.39-Template-Generator.  Install and enable the module
3. Access Tripal page that displays all implementation of Importer plugin in admin/tripal/loaders
4. Click on Tripal Cultivate: Open Science Phenotypic Data link
5. In stage 1, under the list of required column headers, click the link Download a template file link.


laceysanderson (approved):
✅ Code looks good
✅ Manual testing worked perfectly :-)

Unrelated Concern: During manual testing, the job to setup the module (i.e. trpcultivate_phenotypes_install_ontologyterms) took 12 minutes on my mac...

reynoldtan:
I noticed it too, no specific time as I did not time it but there's about that much wait. I will note this.

Issue #41, opened 2023-11-01
G4.41 - Genus Validator


reynoldtan:
### Branch

g4.41-Genus-Validator

### Groups

Group 4 - API | Services | Plugins

### Describe

Create a validator plugin that validates Genus value for
1. No value supplied - Empty value
2. Is a value that exits in chado.organism table, genus field.
3. Is an active genus in that it had been configured.
4. If it is used in combination with Project/Experiment field, that it is the Genus set for the project.

### Design

Identical to the Project validator plugin, this method will implement validator plugin with the following method().
```
public function validate() {
  Check if user left genus field empty.
  If it has a value, check the following:
    Check if it is an active genus - This check will cover both checks to ensure
    that genus is active and that it exits in chado.organism
  
  Final check, but first see it genus was used in combination with project, in this case 
  check to see if genus-project pair matched.
}
```

PR #42, opened 2023-11-07
G4.41 genus validator

reynoldtan:
**Issue #41 Genus Validator**

## Motivation

This PR adds another validation scope or level. The importer form will contain a genus select field that works in sync with the project autocomplete field in that it will autoselect the genus the project had been paired with.

Screen Shot 2023-11-07 at 3 12 13 PM

## What does this PR do?

1. Upgrade the formValidate() of the importer method to support a sequence of validation scopes/levels.
2. Upgrade the formValidate() of the importer to allow to skip a validation scope or level and mark it as upcoming/todo. This is used when a prior scope failed and allow user to rectify the issue before proceeding.
3. Add a genus field that autoselects a genus based on the project entered.
4. Create a route to respond to fetch the genus of the project entered (AJAX request)
5. Upgrade the validator plugin system to allow additional properties to:
    a. headers - reference the headers defined in the importer. This for Header scope validation for later.
    b. skip - mark validator as upcoming or to do. 

### Automated Testing

1. Functional test - provide a project and set genus to the genus set for the project. This is an upgrade of the ImporterShareTest.php to test the new genus field added.
2. Kernel test - test genus validator plugin for:
Correct project-genus pair, incorrect genus, unpaired with a project to see if genus is active (Trait Importer where project is not required) and skip validator.
4. Kernel test - AJAX request controller to fetch genus of project.
Create a http request like in the form to fetch the genus of a project.

### Manual Testing

1. Create test data
Create a project
Create a genus

2. Enable the module and run the Tripal Job created by the install.
    If the module had been already setup, go to configuration page 
    mysite/admin/tripal/extension/tripal-cultivate/phenotypes/ontology 
    
    and set the trait, method, unit and db configuration of the genus.

3. Pair the project - genus in project prop using the genus term in the configuration as the type, alternatively with the module properly configured use the service below.

```
```

4. Load Tripal Cultivate PhenoShare Importer page.
5. Key in project and genus field should select the genus - Clicking the Validate Data File should proceed to next stage (all passed)
6. Key in project and genus field should select the genus but choose another genus - Clicking the Validate Data File should trigger a validation error, detailed in the validation result window
7. Key in non-existent project - Clicking the Validate Data File should trigger a project error and set the genus validation as to do/upcoming validation.


trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -56,18 +56,31 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase {
...
+  protected $service_genusproject;
+  // Genus Ontology configuration service.
+  protected $service_genusontology;
 
   /**
    * {@inheritDoc}
    */
   public function form($form, &$form_state) {
+    // Set genus project service.
+    $this->service_genusproject  = \Drupal::service('trpcultivate_phenotypes.genus_project');
```
laceysanderson: You'll want to add these via dependancy injection. Here is a tutorial focused on how to do this in forms: https://www.drupal.org/docs/drupal-apis/services-and-dependency-injection/dependency-injection-for-a-form

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -216,12 +229,52 @@ public function stage1(&$form, $form_state, $stage_status = '') {
     $form[ $fld_wrapper ]['project'] = [
       '#title' => t('Project/Experiment'),
       '#type' => 'textfield',
-      '#description' => t('Type in the experiment or project title your data is specific to.'),
-      '#weight' => -100,      
+      '#weight' => -100,  
+      '#required' => TRUE,   
+      '#description' => t('Type in the experiment or project title your data is specific to.'),
```
laceysanderson:
```suggestion
      '#description' => t('Enter the name of the experiment or project your data was generated as part of.'), 
```

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -216,12 +229,52 @@ public function stage1(&$form, $form_state, $stage_status = '') {
...
+    foreach($all_genus as $obj_genus) {
+      $genus = $obj_genus->genus;
+
+      $genus_config = $this->service_genusontology->getGenusOntologyConfigValues($genus);
+      if ($genus_config && $genus_config['trait']) {
+        $active_genus[ $genus ] = $genus;
+      }
+    } 
+
+    asort($active_genus);
```
laceysanderson: Lines 243-258 might be best moved into an additional method in the genus ontology service. For example,

```php
/**
 * Retrieve a sorted list of all configured genus.
 * 
 * @return array
 *   Array where the key and value are the genus.
 */
public function getConfiguredGenusList();
```

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -233,6 +286,10 @@ public function stage1(&$form, $form_state, $stage_status = '') {
     $form['file'] = [];
 
     // Other relevant fields here.
+    
+    // This importer does not support using file sources from existing field and local/external source.
+    $form[ $fld_wrapper ]['file']['file_upload_existing'] = ['#access' => 'FALSE'];
+    $form[ $fld_wrapper ]['file']['file_local'] = ['#type' => 'hidden'];
```
laceysanderson: These should be set in the annotation of the importer.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
```diff
@@ -357,24 +414,45 @@ public function formValidate($form, &$form_state) {
...
+            
+            // Load values.
+            $instance->loadAssets($project, $genus, $file, $headers, $skip);
+            
+            // Perform Project Level validation.
+            $validation[ $scope ] = $instance->validate();
+
+            // Save validation result.
+            $storage = $form_state->getStorage();
+            $storage[ $this->validation_result ] = $validation;
```
laceysanderson: I feel like this will only store the validation for the last scope 🤔 Since the key would be the same in $storage and this is a loop for multiple scopes... It's looks like $validation is keyed by scope so maybe this just needs to be an array merge if it's already set?

trpcultivate_phenotypes/src/Plugin/Validators/Genus.php
```diff
@@ -0,0 +1,135 @@
...
+      // the project-genus relationship.
+
+      // Additional check if genus is pared with project, ensure that
+      // it is the genus the project is set to.
+      if (isset($this->project) && $this->project > 0) {
+        // Check if it has a genus set.
+        $project_id = ChadoProjectAutocompleteController::getProjectId($this->project);
+        $genus = $this->service_genus_project->getGenusOfProject($project_id);
+
+        if ($genus && $this->genus != $genus['genus']) {
```
laceysanderson, line 126: This doesn't take into account our discussion that if the project does not have a genus set at all then it can be assigned here and it not be an error.

So I think you want an `elseif (!$genus)` where you create the genus/project connection 🤔

laceysanderson (commented):
Comments are from a code review only.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: I get an error, to me, conflicts with the create method of the ChadoImporterBase???

Fatal error: Declaration of Drupal\trpcultivate_phenoshare\Plugin\TripalImporter\TripalCultivatePhenoshareImporter::create(Symfony\Component\DependencyInjection\ContainerInterface $container) must be compatible with Drupal\tripal_chado\TripalImporter\ChadoImporterBase::create(Symfony\Component\DependencyInjection\ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition)

This is very very interesting resource but we don't want to do this, maybe?
https://www.phptutorial.net/php-oop/php-override-method/

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: applied, this is good 👍

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: interesting 🤔 , but then it's unique in that none of the scope is duplicated as reported by validation result

Screen Shot 2023-11-09 at 2 44 23 PM

Maybe a good idea is keep building the array and when completed write/save it once - might improve the speed too.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: Successfully disabled local source 👍  but the existing field persists and it looks like this field will be available whenever there is a file upload field: https://github.com/tripal/tripal/blob/a6c75eee983ff9261076c5d5e41c50c83a3dd2e6/tripal/src/Form/TripalImporterForm.php#L67

trpcultivate_phenotypes/src/Plugin/Validators/Genus.php
reynoldtan: I've updated the project validator not to trigger failed validation if project has no genus set...this will allow for project without genus set be selected. The genus validator will only check if the genus selected is an active genus. Together will allow unset project to select an active genus and further in the import will create the relationship

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
laceysanderson: Based on the error,
- Method in ChadoImporterBase has $container, $configuration, $plugin_id, $plugin_definition
- Method in your importer had only $container.

To solve this you should just have needed to match the parameters used in create in the chadoimporterbase.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
laceysanderson: That is true that you cannot disable the "Existing" form element this way. That said, your importer should still handle it just fine as Tripal processes the file upload fields into a single, consistent variableset for use later.

Why do you feel so strongly that this needs to be removed?

laceysanderson:
I just saw this mentioned in the issue description and wanted to comment on it:

> Experienced AJAX Response: The database connection is not serialized using #ajax key form API. Implemented a script.

That is a common AJAX error caused by dependency injection of a database connection. This is actually discussed and fixes shown in this Tripal PR: https://github.com/tripal/tripal/pull/1674 as "AJAX appears to be broken". 

The fix is to use code like this in your `#ajax`:
```diff
    $form['obo_existing']['obo_id'] = [
      '#title' => t('Ontology OBO File Reference'),
      '#type' => 'select',
      '#options' => $obos,
      '#default_value' => $obo_id,
      '#ajax' => [
-        'callback' =>  [$this, 'formAjaxCallback'],
+        'callback' =>  [$this::class, 'formAjaxCallback'],
        'wrapper' => 'obo-existing-fieldset',
      ],
      '#description' => t('Select a vocabulary to import.'),
    ];
```

And then to make your ajax callback function static:
```php
public static function formAjaxCallback($form, &$form_state);
```

You should probably do this instead of your own script for better maintainability.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: Why do you feel so strongly that this needs to be removed?

I would like a centralized file source in keeping with the success of the current version, the only field that takes file to be the html5 file field, which is amazing and intuitive as is and I enjoyed clicking using each time. I feel the existing file field is inessential and in some sense perplexing as it is suggesting file that already have been uploaded.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: an override, will try

reynoldtan:
This is the current implementation in my draft https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/c99ba951320f2b97e9fb2f57af1dfa81278f74c0/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php#L238

except the static callback... will try

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: I could not get the constructor to stop complaining about mismatch parameter despite providing the exact same parameter type and count. I found other method of injecting service through the user of setter method... works! manual test is good and all tests passing.

Setter injection
https://medium.com/luciushq/dependency-injection-in-drupal-8-an-introduction-8b3bdf044932

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
reynoldtan: I have revised this section of the code

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
laceysanderson: Ok, good enough reasoning for me :-)

laceysanderson:
Definitely, the key is both to 1) pass in the class name instead of the full $this object and 2) use a static method for the ajax callback.

I've confirmed that your changes related to my other code comments look good!

reynoldtan:
works like a charm 👍👍👍

laceysanderson (approved):
Works beautifully now and code changes look great ✅

Issue #43, opened 2023-11-08
Typing errors 


reynoldtan:
### Drupal Version

9.4

### Tripal Version

4.0.0-alpha1

### Module Version

4.x dev (most recent commit)

### Site Operating System

Debian Linux

### Installation Method

TripalDocker docker

### Observations

Rectify typographical errors in comments section. 

### Expected Behaviour

 n/a

### Steps to Reproduce

n/a

PR #44, opened 2023-11-08
Removes unecessary slash char in namespace

reynoldtan:
**Issue [#43] Typing Error**

## Motivation
Extra forward slash characters found in the namespace that are unnecessary.

## What does this PR do?

This PR removes extra unnecessary slash character. 

### Automated Testing

n/a

### Manual Testing

n/a


laceysanderson (approved):
Looks good 👍 Not critical but doesn't hurt to change :-)

Issue #47, opened 2023-11-16
G4.47 Data File Validator


reynoldtan:
### Branch

g4.47-Data-file-validator

### Groups

Group 4 - API | Services | Plugins

### Describe

Create a validator plugin that validates Data File for

1. Is not empty/0 file id.
1. Has Drupal File Id assigned/created.
2. File Id can loaded using Drupal file system API.
3. A valid TSV or TXT file extension and file MIME type.
4. File size is greater than 0.
5. File can be opened/read.

### Design

 Similar to other validator plugin, this method will implement validator plugin with the following method().

```
public function validate() {
   Check file id is > 0
   
   Load file id
   Check if file can be loaded through Drupal File API
   Use MIME information of the file to validate file type.
   Use the File size information to validate file size is > 0
  
   Open/Read File 
   Check if file can be opened/read
}
```

laceysanderson:
Completed with above merged PR

PR #50, opened 2023-11-22
G4.47 datafile validator

reynoldtan:
**Issue #47 Data File Validator** 

## Motivation

This PR adds another validation scope or level. The Data File validator (could not simplify it it to just File since File is a Drupal reserved keyword/word) validates files provided to the file field in the importer.

This PR is dependent on (https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/42) Genus Validator

## What does this PR do?

1. Updates the scopes array in the validateForm() method of the Importer to include the key/scope 'FILE'.
2. Adds a plugin into the plugin directory that corresponds to the FILE validator scope.

### Automated Testing

1. Kernel Test: Test files for:
  - File is of the expected file type/format of tsv (default file type) or txt (alternative). 
  - A file with 0 file size.
  - A valid extension of tsv file format but is with incorrect file format/mime.
  - Skip validation, set validator as todo/upcoming.

### Manual Testing

1. Install/Enable Tripal Cultivate Phenotypes Module.
2. Create test data
Create a project
Create a genus
Pair the project - genus in project prop using the genus term in the configuration as the type, alternatively with the module properly configured use the service below.

```
```
3. Execute Tripal Job created by the install.
4. Configure module.
5. Load Tripal Cultivate PhenoShare Importer page.
mysite/admin/tripal/loaders/trpcultivate-phenotypes-share

6. Download a template file provided in the importer and use it as test file. Alternatively, sample test files are in this attachment.

[Test Files.zip](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/13444248/Test.Files.zip)

### DIFF
--- a/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
+++ b/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoshareImporter.php
@@ -463,7 +463,7 @@ class TripalCultivatePhenoshareImporter extends ChadoImporterBase implements Con
         $headers = [];
 
         if ($stage == 1) {
-          $scopes = ['PROJECT', 'GENUS'];
+          $scopes = ['PROJECT', 'GENUS', 'FILE'];
 
           // Array to hold all validation result for each level.
           // Each result is keyed by the scope.
--- a/trpcultivate_phenotypes/src/Plugin/Validators/DataFile.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/DataFile.php
@@ -1,0 +1,102 @@
+<?php
+
+/**
+ * @file
+ * Contains validator plugin definition.
+ */
+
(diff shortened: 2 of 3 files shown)


laceysanderson (approved):
✅ Code looks good
✅ Main functionality works great

Specifically I tested:
- Each of the test files provided in this PR (refreshing the form completely between each test) --output was as expected
- Download template and re-upload --output as described (pass because header counts as non-empty)
- Uploaded image (caught by HTML5 element so I didn't see the error but approved because it is caught 👍 )

I also noticed these concerns but felt there were not big enough to stop approval of this PR.

- Uploaded PDF with extension renamed from `.pdf` to `.txt`. This passed as apparently mac changes the mime/type to text when you change the extension 🤔 See issue #59
- Uploaded file that passes validation, then upload file that fails validation WITHOUT REFRESHING THE FORM. This allows the validation to pass even though the file is not compliant 😬 See Issue #60
- Automated testing isn't actually covering the use cases you expect based on the tests written. See Issue #61

Issue #48, opened 2023-11-20
G4.48-Header Validator 


reynoldtan:
### Branch

g4.48-Header-Validator

### Groups

Group 4 - API | Services | Plugins

### Describe

Create a validator plugin that validates Header for:

1. Header line is the data file is not empty.
2. All headers defined by headers property in the Importer match the headers in the file.

    

### Design

Similar to other validator plugin, this method will implement validator plugin with the following method().

```
public function validate() {
  Use drupal file system and load the file using the file_id import assets.
  Use the file uri of the loaded file to read the contents of the file
    Check the first line of the file if it is empty
    Refer to the headers import asset and compare with the header row (first line in the file)
}
```

laceysanderson:
Completed with above merged PR

PR #51, opened 2023-11-22
G4.48 column headers validator

reynoldtan:
**Issue #48 Column Headers Validator** 

## Motivation

This PR adds another validation scope or level. The Column Headers Validator validates column headers in the file to matches the headers defined by the Importer.

This PR is dependent on (https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/50) Data File Validator

## What does this PR do?

1. Updates the scopes array in the validateForm() method of the Importer to include the key/scope 'HEADERS'.
2. Adds a plugin into the plugin directory that corresponds to the HEADER validator scope.

### Automated Testing

1. Kernel Test: Tests headers for:
  - Headers row (line no.1 in the data file) is missing.
  - Mismatch column headers: Extra header and Missing header
  - Skip validation, set validator as todo/upcoming.

### Manual Testing

1. Install/Enable Tripal Cultivate Phenotypes Module.
2. Create test data
Create a project
Create a genus
Pair the project - genus in project prop using the genus term in the configuration as the type, alternatively with the module properly configured use the service below.

```
```
3. Execute Tripal Job created by the install.
4. Configure module.
5. Load Tripal Cultivate PhenoShare Importer page.
mysite/admin/tripal/loaders/trpcultivate-phenotypes-share

6. Download a template file provided in the importer and use it as test file.
7. Alter header row (line no.1 of the template file) to test validation.

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
@@ -78,6 +78,7 @@ class Headers extends TripalCultivatePhenotypesValidatorBase implements Containe
     // Headers:
     //   - Header line is not empty.
     //   - No missing headers.
+    $file_column_headers = [];
 
     // Open file and read the first line, the column headers line.
     $file = File::load($this->file_id);
--- a/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
@@ -98,6 +98,10 @@ class Headers extends TripalCultivatePhenotypesValidatorBase implements Containe
       }
     }  
     
+    // First remove any empty array elements.
+    $file_column_headers = array_filter($file_column_headers);
+
+    // Then check if there are any.
     if (empty($file_column_headers)) {
       // No file has been uploaded into the data file field.
       $validator_status['status']  = 'fail';
--- a/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
@@ -1,0 +1,125 @@
+<?php
+
+/**
+ * @file
+ * Contains validator plugin definition.
+ */
+
+namespace Drupal\trpcultivate_phenotypes\Plugin\Validators;
+
+use Drupal\trpcultivate_phenotypes\TripalCultivatePhenotypesValidatorBase;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\file\Entity\File;
+use Drupal\Core\File\FileSystemInterface;
+
+/**
+ * Validate Required Headers.
+ * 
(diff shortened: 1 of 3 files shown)


trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
```diff
@@ -0,0 +1,125 @@
...
+    // This happens when other prior validation failed and this validation could only proceed
+    // when input values in the failed validator have been rectified.  
+    if ($this->skip) {
+      $validator_status['status'] = 'todo';
+      return $validator_status;
+    }
+
+    // Headers:
+    //   - Header line is not empty.
+    //   - No missing headers.
```
laceysanderson, line 80: Define the variable before using it. Plus this way if you do wall into the `else` cases you are not catching here at least you wont get a variable not defined error.

```suggestion
    //   - No missing headers.
    $file_column_headers = [];
```

trpcultivate_phenotypes/tests/src/Kernel/PluginValidatorTest.php
```diff
@@ -378,6 +387,63 @@ public function testDataFilePluginValidator() {
     $this->assertEquals($validation[ $scope ]['status'], $status);
   }
 
+  /**
+   * Test Headers Plugin Validator.
+   */
+  public function testScopePluginValidator() {
```
laceysanderson:
```suggestion
  public function testColumnHeaderPluginValidator() {
```

trpcultivate_phenotypes/src/Plugin/Validators/Headers.php
```diff
@@ -0,0 +1,125 @@
...
+          // Get the first line - column headers only.
+          $file_column_headers = fgets($file_handle);
+          fclose($file_handle);
+
+          $file_column_headers = str_getcsv($file_column_headers, "\t");
+        }
+      }
+    }  
+    
+    if (empty($file_column_headers)) {
```
laceysanderson, line 105: This isn't actually catching the case you want it to since lines 92-95 result in the following if there is an empty line in the file:
```
Array
(
    [0] => ''
)

```

The following suggestion fixes this:
```suggestion
    // First remove any empty array elements.
    $file_column_headers = array_filter($file_column_headers);

    // Then check if there are any.
    if (empty($file_column_headers)) {
```

laceysanderson (dismissed):
✅ Manual testing with the suggested changes works perfectly
✅ Code review looks good **once the following suggestions are committed**.

I am marking this approved but that is **assuming that the following suggestions are committed first.**

trpcultivate_phenotypes/tests/src/Kernel/PluginValidatorTest.php
reynoldtan: good catch

laceysanderson: This one doesn't seem to have been applied.

laceysanderson (approved):
Adding back in the approved review as my other one seems to have been dismissed 🤔  All the changes suggested have now been applied.

Issue #59, opened 2023-12-08
[Bug] Changing the extension on a PDF to TXT is not caught by the Importer if OS changes mime/type too


laceysanderson:
### Drupal Version

10.1

### Tripal Version

4.x dev (most recent commit)

### Module Version

4.x dev (most recent commit)

### Site Operating System

Debian Linux

### Installation Method

TripalDocker docker

### Browsers

Firefox

### Observations

When I upload a file that is a PDF but has the file ending `.txt` I expect the importer to give me a validation error but it does not on Mac OS. It seems this is because Mac OS X changes the mime type automatically when you edit the extension 🤦‍♀️ 

<img width="968" alt="Screenshot 2023-12-08 at 2 06 12 PM" src="https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/1566301/eb44b5ae-5b77-4fbb-ac63-7aee63cd335f">

<img width="392" alt="Screenshot 2023-12-08 at 2 07 06 PM" src="https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/1566301/dd8e91d7-0280-4617-98fd-255df1b31823">

### Expected Behaviour

I expected to be told that the file was of an unsupported file type.

### Steps to Reproduce

1. Create a fresh docker (including the code in PR #50 if it's not yet merged).
2. Create an organism
3. Create a Project
4. Configure the ontologies for the organism here: http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
5. Use drush php:cli to associate the organism with the project
```php
\Drupal::service('trpcultivate_phenotypes.genus_project')
  ->setGenusToProject(PROJECT ID, 'GENUS', TRUE);
```
6. Navigate to the Pheno Share importer: http://localhost/admin/tripal/loaders/trpcultivate-phenotypes-share (you may need to clear the cache if you get page not found)
7. Download this [sample.pdf](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/13620417/sample.pdf) or use your own. Open up finder, edit the filename and change the extension from `.pdf` to `.txt`
8. Upload this modified file in the importer form you already opened by clicking the browse, selecting it and then clicking upload.
9. Click the "Validate Data File" button and notice that the validation passed...

laceysanderson:
Fixed by PR

PR #64, opened 2023-12-15
Fixes failed test: pdf as tsv

reynoldtan:
**Issue #59 - PDF as tsv file

## Motivation
Data file fails to trigger validation when pdf file converted into a tsv file is uploaded.

## What does this PR do?

1. Add additional check to inspect the file content for pdf signatures and trigger the correct validation error..

## Testing
Below is a tsv file from a pdf file where the pdf extension was replace with tsv.
[sample.txt](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/13687671/sample.txt)

Provide this file into the importer file field and validate to trigger an incorrect file format/extension error.


reynoldtan:
The fix to this issue is very specific to pdf file. It needs to be a check of file content to have content structure (by tabs or by comma) and infer the validity of the file.  I will revise the rule for this one

carolyncaron:
We already discussed this at our latest meeting, but to summarize it here: we think this approach of checking for a disguised pdf is a bit too specific. A broader approach would be to validate that the file is tab-delimited by parsing the first line. Ideally, we would go so far as to implement a method in the Tripal Importer class that can do checks for tab-delimited text and the number of columns, since this functionality is needed by most (maybe all) importers written for Tripal.

laceysanderson (approved):
Code changes look perfect and a manual testing provides the error message I expected :-) 
✅ Non-tsv file pretending to be one gets an error
✅ Correct TSV file passes validation.

Issue #61, opened 2023-12-08
G5.61 Automated testing for DataFile Validator is not covering what we expect it to based on the test code.


laceysanderson:
### Branch

g5.61-datafileValidatorTesting

### Groups

Group 5 | Automated Testing | Performance

### Describe

According to the PR the testDataFilePluginValidator Kernel Test should cover these cases:

- File is of the expected file type/format of tsv (default file type) or txt (alternative).
- A file with 0 file size.
- A valid extension of tsv file format but is with incorrect file format/mime.
- Skip validation, set validator as todo/upcoming.

A reading of the code falls in line with these expectations. However, when I run the code coverage and check that validator, this does not appear to be the case:

![image](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/1566301/9fbbf2fc-d136-4d5f-9ccd-706db6eb7633)

### Design

Improve the tests so they actually cover:

1. No file attached
2. Cannot load file
3. File is empty

PR #65, opened 2023-12-15
Creates more test cases for file validator

reynoldtan:
[**Issue #[61]**](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/61) Expand test cases for Data File validator

## Motivation
The test cases for data file validator failed to include test when file:
1. No file provided
2. Could not load a file
3. Empty file.

## What does this PR do?
This PR add test cases for file:
1. An empty file or file size is 0
2. Could not load a file (filed does not exists)
5. File is not provided


laceysanderson (approved):
The tests you added here were good and caught a few extra cases :-)
However, there were still two cases that were not caught and quite important. I've added tests to catch these as well.

Specifically,
- the case where the file exists but is made unreadable due to permissions. This is now tested by file-7.
- the case where the file exists + is readable but is not actually a tsv file. This is now tested by file-8.

Once I was able to test when a file is not readable via permissions, I found a small bug which I have now fixed. This is that when you try to open a file that is not readable using fopen, it throws an E_WARNING which breaks the flow for good feedback to the user. To fix I just added as `is_readable()` check first.

Once you've taken a look at my changes and if you agree @reynoldtan, this PR can be merged :-) 🎉

Issue #58, opened 2023-12-04
G4.58 - Phenotypes Share Values Validator


reynoldtan:
### Branch

g4.58-Phenoshare-Values-Validator

### Groups

Group 4 - API | Services | Plugins

### Describe

Focuses on validating values in a data file submitted to Tripal Cultivate Phenotypes Share Importer.

There is a Plugin Type for validators which was created as part of #37. Each type of validation we want to do is an instance of this plugin. A scope is assigned to each validator instance to indicate what part of the file it validates and the order it should be in.

This issue is to design validator instances focused on validating the new phenotypes file format. The first attempt of this by Reynold, validated all values/columns in the file in a single validator. However, we would like to move to a different model.

### Design

Create an instance of validator plugin: PhenoShareImportValues

```
 * @TripalCultivatePhenotypesValidator(
 *   id = "trpcultivate_phenotypes_validator_share_values",
 *   validator_name = @Translation("Phenoypes Share Importer Values Validator"),
 *   validator_scope = "SHARE IMPORT VALUES",
 * )
```

The plugin will validate the following:
1. **Trait Name** - Trait name exists in the cv configured for trait in the genus selected. This column is required.
2. **Method Name** - Method name exists in cv configured for method in the genus selected and is one of the methods paired to the trait name. This column is required.
3. **Unit** - Unit must exists in the cv configured for the unit in the genus selected. This column is required.
4. **Germplasm Accession** - Must exists in chado.stock table. This column is required.
5. **Germplasm Name** - Must exists in chado.stock table. This column is required
6. **Year** - A four digit value and must not exceed the current year. This column is required.
7. **Replicate** - an integer value. This column is required
8. **Value** - Use the unit data type to determine the data type for this column. A quantitative unit will require numeric value whereas qualitative value will be descriptive text. This column is required
9. **Data Collector** - String, name of a person, institute or organization. This column is required.

laceysanderson:
Here are the files from Reynolds first attempt at this:

- [PluginShareValuesValidatorTest.txt](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/14225777/PluginShareValuesValidatorTest.txt)
- [PhenoShareImportValues.txt](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/14225778/PhenoShareImportValues.txt)
[...]

PR #66, opened 2023-12-21
G4.58 phenoshare values validator

reynoldtan:
**Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/58** Phenotypes Share Values Validator

## Motivation

This PR is an instance of the validator plugin that will validate data (rows and column) in a phenotype data file uploaded into Tripal Cultivate Phenotypes Share Importer.

image

## What does this PR do?

1. Updates trait service with method to fetch a trait, a trait method, method unit and unit data type (Qualtitative and Quantitative)
2. Updates validator base to to reference configuration value set to allow new traits.
3. Create a plugin validator helper that can be reused to validate values in other importer.
4. Validate the following values:
  - Trait name, method name and unit name  must exists (if allow new is set to false)
  - Germplasm Accession/Germplasm Name must exists
  - Year is a four digit value and must be between 1900 to the present year.
  - Replicate is  number greater than 0
  - Value should match the unit data type.
  - All columns must have value and and column with empty value will trigger a validation error.
  
  This validator is based on the following column header set:
  
**Trait Name:** The full name of the trait as you would like it to appear on a trait page. This should not be abbreviated (e.g. Days till one open flower).

**Method Name:** A short (<4 words) name describing the method. This should uniquely identify the method while being very succinct (e.g. 10% Plot at R1).

**Unit:** The unit the trait was measured with. In the case of a scale this column should defined the scale. (e.g. days)

**Germplasm Accession:** The stock.uniquename for the germplasm whose phenotype was measured. (e.g. ID:1234)

**Germplasm Name:** The stock.name for the germplasm whose phenotype was measured. (e.g. Variety ABC)

**Year:** The 4-digit year in which the measurement was taken. (e.g. 2020)

**Location:** The full name of the location either using “location name, country” or GPS coordinates (e.g. Saskatoon, Canada)

**Replicate:** The number for the replicate the current measurement is in. (e.g. 3)

**Value:** The measured phenotypic value. (e.g. 34)

**Data Collector:** The name of the person or organization which measured the phenotype.

### Automated Testing

Kernel test - Plugin Values Validator
1. Valid data file with data in every column header (pass, no validation error).
2. Valid data file with non-existent Trait name, method name and unit with allow new traits set to true (pass, no validation error).
3. Every column header is set to an empty value in a loop where each iteration is validated (failed, with validation error that column should have a value).
4. With allow new configuration set to false (Not to allow new trait), test non-existent trait name, method name and unit (failed, trait, method and unit must exist)
5. Test non-existent germplasm/stock (failed, germplasm does not exist).
6. Test invalid year (text, 2 digit value and future year).
7. Test invalid replicate (0, text and negative value).

Kernel test - Traits service
1. Get trait by trait id and by trait name.
2. Get trait method by trait id and by trait name.
3. Get method unit
4. Get unit data type.

### Manual Testing

1. Create test data in
- Genus (chado.organism)
- Project (chado.project)
- Stock (chado.stock)
- Create trait, method and unit and relationship. (Use trait importer)
2. Install and enable Tripal Cultivate Phenotypes module
3. Configure Genus Ontology Configuration for the genus created
4. Note the value for Allow New Trait as this will allow non-existent trait, method and unit in the data file when set to TRUE (checked). It will trigger a validation when uploading the same when this configuration is set to FALSE (unchecked - not to allow)
5. Access mysite/admin/tripal/loaders/trpcultivate-phenotypes-share and use the Download a template file to create a tsv file.
6. Test the following:
- Upload the template file without alteration
- Update the template file and remove the comment before uploading (test an empty data row)
- Provide some values leaving some column headers empty
- With the the allow new set to TRUE (checked) Add a value to Trait Name, Method Name and Unit and use the stock created for Germplasm Accession and Germplasm Name. NOTE: make sure to abide by the Unit data type. Quantitative type must have a numeric value whereas Qualitative expects a text value.
- Repeat the same with the allow new set to FALSE (unchecked)
- Provide a non-existent germplasm, invalid year and invalid replicate.


laceysanderson (commented):
The tests were failing due to some changes made to the order in which chado was accessed in the testing environment. I've made the following changes:
- constructors should only set dependencies. I removed the call to defineGenusOntology() because it was using one of the dependencies it had just set.
- ensure the functional tests specifically set chado and that any services using chado were reset prior to the test after chado was made available.
- fixed multiple typos in the trait test.

The tests are now passing locally for me. 🙏 that they now pass on GitHub.

laceysanderson (changes requested):
This PR adds in A Lot... given that and that the column headers are being changed drastically, this PR will **not** be merged in it's current state.

I'm marking it as changes required to ensure it is not merged. I will chat with Reynold about this tomorrow but I'm currently thinking the best plan of action would be for me to move chunks of this functionality into a separate PR based on what I believe should be merged. Not everything here will be eventually merged.

laceysanderson:
Okay, I've stripped the values validator and the trait importer out of this PR so that the remaining changes can be mergable. The values validator classes have been backed up in the linked issue so that parts of them can return once we rework the approach for validating the values in the file and upgrade to the new columns

laceysanderson (approved):
After removing the phenotypes share values validator, helper and trait importer, I tested the module manually:

✅ module installs
✅ pheno share importer renders beautifully
✅ pheno share importer existing validation works good
✅ ontology configuration form works as expected

That means the remaining changes are ready to merge! 🎉

Issue #60, opened 2023-12-08
[Bug] When first file passes validation, subsequent uploads without page refresh do not trigger errors.


laceysanderson:
### Drupal Version

10.1

### Tripal Version

4.x dev (most recent commit)

### Module Version

4.x dev (most recent commit)

### Site Operating System

Debian Linux

### Installation Method

TripalDocker docker

### Browsers

Firefox

### Observations

I tested the Pheno Share importer within the same page session due in part to laziness and in part since this is how a curator might function (import file, get error, immediately import another fixed file into the same form). However, I started with a file that did pass validation and then followed up with one that should not. In this case, the second file did not show validation error! 

This was not enough to stop the PR I was reviewing since the order is not what you would expect from a user... but it should be addressed as it can happen in this order if a user passes validation and then sees traits they didn't expect in the next step of the form and wants to re-upload the newest version.

### Expected Behaviour

I expected to see the same error message in these two cases:
1. Upload a file failing validation from a fresh page reload
2. Upload the same file failing validation but immediately after one already passed.

### Steps to Reproduce

1. Create a fresh docker (including the code in PR #50 if it's not yet merged).
2. Create an organism
3. Create a Project
4. Configure the ontologies for the organism here: http://localhost/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
5. Use drush php:cli to associate the organism with the project
```php
\Drupal::service('trpcultivate_phenotypes.genus_project')
  ->setGenusToProject(PROJECT ID, 'GENUS', TRUE);
```
6. Navigate to the Pheno Share importer: http://localhost/admin/tripal/loaders/trpcultivate-phenotypes-share (you may need to clear the cache if you get page not found)
7. Upload a file that we know will pass validation (e.g. [valid-not-empty-readable-alt.txt](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/13620578/valid-not-empty-readable-alt.txt))
8. Confirm that it passes validation as expected.
9. Now immediately (without refreshing the page, simple scroll up and open the Stage 1 again) upload a file you expect to fail validation (e.g. [not-valid-file-size-zero.txt](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/files/13620588/not-valid-file-size-zero.txt))
10. Confirm that it still passes validation and pushes you to the next stage 😬 
[...]

PR #67, opened 2024-01-02
G bug.60 subsequent uploads

reynoldtan:
**Issue #60 When first file passes validation, subsequent uploads without page refresh do not trigger errors

## Motivation

After uploading a data file with passing validation result, attempt to re-upload a file (with error) does not trigger any validation error.

## What does this PR do?

This PR adds a check in formValidate() method to check for which stage the form is currently at and what triggering element caused to submit the form.

When the stage is greater than 1 (no longer in stage 1 where file upload is required) and the form triggering element is to Validate Data File, then the stage is reset to stage 1 and cache the value to update stage cacheing mechanism of the form.

### Automated Testing

### Manual Testing

1. Create test data in
- Genus (chado.organism)
- Project (chado.project)
- Stock (chado.stock)
2. Install and enable Tripal Cultivate Phenotypes module
3. Configure Genus Ontology Configuration for the genus created
4. Ensure allow new trait is set to true (checked), this will allow non-existent trait to be uploaded (to skip creating trait, method and unit).
5. Access mysite/admin/tripal/loaders/trpcultivate-phenotypes-share and use the Download a template file to create a tsv file.
6. Using test data, fill out the columns. Make sure that this file will pass validation by providing correct data values.
7. Proceed to upload stage and upload the file. With passing validation, the current stage is now either stage 2 or stage3.
8. Alter the data file so that it will fail validation (ie: empty column, invalid year or rep set to 0)
9. Enable or expand stage 1 and re-upload file.
This should trigger validation and stage is accounted for so that it shows the correct stage despite previous file upload moved the stages to either stage 2 or 3.
10. Repeat steps 6-9 while in stage 3.


laceysanderson (approved):
Confirmed that this fixes the bug specified.

Issue #70, opened 2024-02-08
G5.70 Improve tests and to use test traits.


laceysanderson:
### Branch

g5.70-updateExistingTests

### Groups

Group 5 | Automated Testing | Performance

### Describe

This task is a general updating of existing tests to use some new features, versions and ot test a few more cases.

### Design

This is just a general, as needed task so does not need a design.

laceysanderson:
Partially complete with above PR.

Remaining is to 

- [x] improve the tests for better coverage, 
- [x] less functional tests, 
- [ ] use the test trait in PR #54

reynoldtan:
Summary of files that may require improvements:

/Functional

- [ ] ConfigOntologyTermsTest.php
- [ ] ConfigRRulesTest.php
- [ ] ConfigWatermarkTest.php
- [ ] InstallTest.php

/Kernel/Traits

- [ ] ValidTraitTest.php

/Kernel

- [x] ServiceGenusOntologyTest.php
- [x] ServiceGenusProjectTest.php
- [x] ServiceTemplateGeneratorTest.php
- [x] ServiceTermTest.php 

/Traits

- [ ] PhenotypeImporterTestTrait.php

reynoldtan:
> * [ ]  use the test trait in PR [G2.53 trait importer #54](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/54)

Test trait has been implemented in tests in PR G2.53

tests/src/Kernel/TripalImporter
trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormTest.php
trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterRunTest.php

PR #71, opened 2024-02-09
Updates our testing to include PHP 8.3 and Drupal 10.2

laceysanderson:
**Issue #70**

## Motivation

There was just a PR merged in core that updates Tripal core to support PHP 8.3 and Drupal 10.2. Obviously we want to test on these versions as well to confirm our support. This PR does that :-) 

## What does this PR do?

1. Updates the existing testing workflows to include PHP 8.3 and Drupal 10.2
2. Updates workflows to use checkout action v4 to remove nodejs version warning
3. Updates workflows to use test-tripal-action v1.3 so that we can have better control over versions of php used.
4. Expands the testing grid in the readme to include the new version combinations.
5. Makes the dockerfile default to php 8.3 and Drupal 10.2 since these are the versions we should be targeting now.
6. Drupal 10.2 added a new permission to access help pages and became more stringent on permissions for pages. We made fixes to our existing tests with this in mind.

## Testing

1. Make sure the automated testing passes on all versions.
2. Build the docker image both with and without phpversion build-args to ensure it pulls the correct tripal docker file.

### DIFF
--- a/.github/workflows/ALL-PHPUnit.yml
+++ b/.github/workflows/ALL-PHPUnit.yml
@@ -16,11 +16,20 @@ jobs:
         php-version:
           - "8.1"
           - "8.2"
+          - "8.3"
         pgsql-version:
           - "13"
         drupal-version:
           - "10.0.x-dev"
           - "10.1.x-dev"
+          - "10.2.x-dev"
+        exclude:
+          - php-version: "8.3"
+            pgsql-version: "13"
+            drupal-version: "10.0.x-dev"
+          - php-version: "8.3"
+            pgsql-version: "13"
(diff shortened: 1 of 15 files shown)


reynoldtan:
Confirmed resolved. 
I was able to build a local docker without any errors. 

I Noticed that the sub-menu is now on the left hand side (all the time), but this is unrelated and is about Drupal.

reynoldtan (approved):

reynoldtan:
Silly me, this little button puts it back beneath top navigation
Screen Shot 2024-02-09 at 10 58 38 AM 1

Issue #53, opened 2023-11-27
G2.53 - Phenotypic Trait Importer


reynoldtan:
### Branch

g2.53-Trait-Importer

### Groups

Group 2 - Data Importing

### Describe

Upgrade Phenotypes Traits Importer

Current implementation: https://github.com/UofS-Pulse-Binfo/analyzedphenotypes/blob/7.x-3.x/includes/TripalImporter/TraitImporter.inc

![image](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/assets/15472253/a9246e98-5a64-4eab-b975-eba432ec5937)

### Design

1. Create an instance of the Tripal Importer
2. Use Tripal Cultivate Phenotypes Validator Plugin to validate trait import file for
  - Selecting an active genus (genus that have been configured) (GENUS validator plugin)
  - File is in the indicated/expected file type of tsv and txt (FILE validator plugin)
  - All headers defined by the headers property in the trait importer form are present (HEADERS plugin validator)
  - Each column header has a value, no empty column. (TRAIT IMPORTER VALUES validator plugin)
3. Create a traits service that will handle creation of traits, method and unit cvterms as well as required relationships between trait-method, method-unit and supplemental metadata about the unit.

PR #77, opened 2024-05-13
G4.76 STEP 1 --Divides trait import value plugin instance into 3 separate instances

### LINKED ISSUE #76: G4.76 - Divide Trait Import Value plugin instance into 3 separate instances
### Branch

g4.76-divideTraitImportValue

### Groups

### Describe

After spending time thinking about/discussing how to organize plugin instances for validating the phenotype share importer, we realized that the plugin instance to validate columns in the trait importer could be split up into 3 separate instances:
1. Empty values
2. That the combination of Trait Name + Method Short Name + Unit is not duplicated within the datafile **or the database**
3. The column "Type" contains either the value "Qualitative" or "Quantitative"

See issue #53 for the original design of the current validators

### Design

The current validator class TraitImportValues (`TraitImportValues.php`) will be divided into the following 3 classes:
1. EmptyCell
2. DuplicateTraits
3. TraitTypeColumn

These 3 classes will all be under the scope of "FILE ROW" and the "TRAIT IMPORT VALUES" scope will be removed.

**Empty Cell:** Currently checks for individual empty cells and if the file is missing data rows (ie. only has a header). We will want to limit this plugin instance to check individual cells, and relocate the functionality check for data rows to another pre-existing plugin or new one altogether. Additionally, we will want a way to specify which cells are _allowed_ to be empty, as this will be the case for other importers that will use this plugin instance.

**Duplicate Traits:** Currently only checks for duplicate traits in the file. Since this is provided the scope of a row in the file, it will need to keep track of traits that were already processed before it in addition to checking if the trait + method + unit combo already exists at the database level.

**TraitTypeColumn:** Our plan is to make this validator highly generic in the hopes it can be re-used for the phenotype importers. Renamed to something like "ValidValueFromList"? In addition to being provided with a single row from the input file, the $context array will contain:
- indices: an array of indices for which the value is being checked
- valid_values: an array of values that are valid for the cell located in any of the indices in the indices array

carolyncaron:
**Issue #76**

## Motivation

After spending time thinking about/discussing how to organize plugin instances for validating the phenotype share importer, we realized that the plugin instance to validate columns in the trait importer could be split up into 3 separate instances:

1. Empty values
2. That the combination of Trait Name + Method Short Name + Unit is not duplicated within the datafile or the database
3. The column "Type" contains either the value "Qualitative" or "Quantitative"

See issue #53 for the original design of the current validators

## What does this PR do?
This PR creates 3 new validators:
1. **EmptyCell** - Given values from a single row in an input file, and a list of indices to check for, this validator reports `pass` if the values at the specified indices are not empty, and status `fail` if any are.
2. **ValueInList** - Given values from a single row in an input file, a list of indices to check for, and a list of valid input values, this validator reports `pass` if the values at the specified indices contain any one of the specified valid values. Otherwise, a status of `fail` is returned.
3. **DuplicateTraits** - Given values from a single row in an input file, and an array mapping the index for each of: trait name, method name and unit. Checks for duplicates at both the file level and database level, returns `pass` if no duplicates found, `fail` if the combination of trait + method + unit is found.

Additionally, 2 new methods were added to the Validator Base class:
1. **validateRow($row_values, $context)**: The 3 validators above implement this method instead of validate(), in order to pass in the values of the row and any relevant context to that row (ex: indices within the row that the validator should act on).
2. **checkIndices($row_values, $indices)**: Given values from a single row in an input file, and a list of indices to check for, throws an error if the list of indices does **not** make sense in the following contexts:
   - If no indices are provided
   - If the number of indices surpasses the number of values in $row_values
   - If any of the keys in indices is outside of the range of keys for $row_values

**IMPORTANT NOTE**: This PR does NOT remove the class **TraitImportValues** nor does it change the Trait Importer to call these new validators instead of the TraitImportValues class. This will be addressed in a separate PR. The purpose was to flesh out these validators first, and test them independently before moving on to having them utilized by an importer. Some tests were marked as skipped which will also be addressed in the next PR.

### Automated Testing
- All validator tests are now located in `Kernel/Validators`
- New test classes: `ValidatorBaseTest`, `ValidatorTraitImporterTest` and `ValidatorFileRowScopeTest`

**ValidatorBaseTest->testValidatorBaseCheckIndices**
- [x] Provide an empty list of indices
- [x] Provide too many indices for the number of row values
- [x] Provide invalid indices (out of bounds for the indices in row_values array)

**ValidatorFileRowScopeTest->testValidatorValueInList**
- [x] Check for valid value in a single column (1 index)
- [x] Check for valid value in multiple columns (3 indices)
- [x] Check for column with invalid value (1 index)
- [x] Check for 1 column with a valid value, 1 with invalid (2 indices)

**ValidatorFileRowScopeTest->testValidatorEmptyCell**
- [x] Provide a list of indices for cells that are not empty
- [x] Provide a list of indices for cells that are empty
- [x] Provide a list of indices for the entire row (mixture of empty and non-empty cells)

**ValidatorTraitImporterTest->testValidatorDuplicateTraitsFile**
- [x] Enter a single row and check it is in the $unique_traits array
- [x] Re-enter the previous row (duplicate)
- [x] Provide a key in $context['indices'] that is not one of: trait, method, unit
- [x] Enter a 2nd unique row
- [x] Enter a third row with same trait + method as row 1, same unit as row 2 (not a duplicate)

**ValidatorTraitImporterTest->testValidatorDuplicateTraitsDatabase**
- [x] Validate a single row and check against an empty database
- [x] Enter a trait into the database and then try to validate the same trait details
- [x] Validate trait details where trait name and method name already exist in the database, but unit is unique

### Manual Testing
2. Go to localhost (or whichever port you're using for this container) and sign in
3. Add a new organism by navigating to `Tripal` > `Content` > `Add Tripal Content` and select `Organism` from the list
4. Configure the module by going to `Tripal` > `Extensions` > `Tripal Cultivate Phenotypes Configuration`
5. Open  Execute PHP by going to `Devel` > `Execute PHP`. You may need to add it as an option by going to `Configure` and checking `PHP Execute` in the list first.

The following are templates for testing each validator individually. Play around with different values in $file_row, or $context['indices']. You can also try to trigger an exception by the checkIndices() method!

**EmptyCell**
```
// Set the genus variable to a genus you've already created
$genus = 'Tripalus';

$traits_service = \Drupal::service('trpcultivate_phenotypes.traits');
$traits_service->setTraitGenus($genus);

// Create the plugin manager

// An array that contains the contents of a row in a file for the Trait Importer
$file_row = [
  'My trait',
  '',
  'My method',
  '',
  'My unit',
  ''
];

$validator_id = 'trpcultivate_phenotypes_validator_empty_cell';

// We need a context array to specify which indices are NOT allowed to be empty
$context['indices'] = [ 0, 2, 4 ];

// Validate our file_row. Expected validation status = pass
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);

// Set our required columns to be ones that are empty. Expected status = fail
$context['indices'] = [ 1, 3, 5 ];

$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);
```

**ValueInList**
```
// Set the genus variable to a genus you've already created
$genus = 'Tripalus';

$traits_service = \Drupal::service('trpcultivate_phenotypes.traits');
$traits_service->setTraitGenus($genus);

// Create the plugin manager

// An array that contains the contents of a row in a file for the Trait Importer
$file_row = [
  'My trait',
  'My trait description',
  'My method',
  'My method description',
  'My unit',
  'Quantitative'
];

// A list of valid inputs
$valid_inputs = ['Quantitative', 'Qualitative'];

$validator_id = 'trpcultivate_phenotypes_validator_value_in_list';

// We need a context array to specify which indices to check for valid input
$context['indices'] = [ 5 ];
// Set the context array of valid inputs
$context['valid_values'] = $valid_inputs;

// Validate our file_row. Expected validation status = pass
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);

// Try checking for valid inputs at other indices. Expected status = fail
$context['indices'] = [ 1, 3 ];
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);
```

**DuplicateTraits: file level**
```
// Set the genus variable to a genus you've already created
$genus = 'Tripalus';

$traits_service = \Drupal::service('trpcultivate_phenotypes.traits');
$traits_service->setTraitGenus($genus);

// Create the plugin manager

// An array that contains the contents of a row in a file for the Trait Importer
$file_row = [
  'My trait',
  'My trait description',
  'My method',
  'My method description',
  'My unit',
  'Quantitative'
];

// Test the DuplicateTraits validator at the file level
// Each call of validateRow() is treated like a new row within the same file, unless the plugin gets re-instantiated
$validator_id = 'trpcultivate_phenotypes_validator_duplicate_traits';
// We need a context array to specify which indices are the name of the trait, method and unit
$context['indices'] = [ 'trait' => 0, 'method' => 2, 'unit' => 4 ];

// Validate our file_row. Expected validation status = pass
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);

// Re-validate the same row. Expected status = fail
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);
```

**DuplicateTraits: database level**
```
// Set the genus variable to a genus you've already created
$genus = 'Tripalus';

$traits_service = \Drupal::service('trpcultivate_phenotypes.traits');
$traits_service->setTraitGenus($genus);

// Create the plugin manager

// An array that contains the contents of a row in a file for the Trait Importer
// Keys are set to the column headers as the insertTrait() method expects an associative array
$file_row_insert = [
  'Trait Name' => 'My trait',
  'Trait Description' => 'My trait description',
  'Method Short Name' => 'My method',
  'Collection Method' => 'My method description',
  'Unit' => 'My unit',
  'Type' => 'Quantitative'
];

// Test the DuplicateTraits validator at the database level
$validator_id = 'trpcultivate_phenotypes_validator_duplicate_traits';
// We need a context array to specify which indices are the name of the trait, method and unit
$context['indices'] = [ 'trait' => 0, 'method' => 2, 'unit' => 4 ];

// Insert the values of our row into the database
$combo_ids = $traits_service->insertTrait($file_row_insert);

// Now validate our row. Expected validation status = fail
$file_row = array_values($file_row_insert);
$validation_status = $instance->validateRow($file_row, $context);
print_r($validation_status);
```


trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,157 @@
...
+    // Grab our traits service
+    // *************************************************************************
+    // * @TODO: The following code assumes only one record at a time is returned
+    // * by each of the getter methods, but in reality we want to accomodate multiple
+    // * potential methods for a trait, and multiple units for a method. Thus,
+    // * each result will need to be iterated through to find the right one if this
+    // * is a duplicate. Issue #78 should address this in the getter functions
+    // *************************************************************************
+    $service_traits = \Drupal::service('trpcultivate_phenotypes.traits');
+    $trait_record = $service_traits->getTrait(['name' => $trait]);
```
reynoldtan: The duplicate check on db level will always search for all lowercase form of the trait name due to case conversion in line 87. Reference/use the original value of the array element $row_values[$context['indices']['trait']] and pass it as parameter to the getter should solve this.

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,157 @@
...
+    if ($trait_record) {
```
reynoldtan: This will be replaced with the combo getter we were working on, right? which will simplify this block.

reynoldtan:
**Manual Test:**
First and foremost, impressive work @carolyncaron 

I performed manual test outlined above and the method/logic works as described. Suggestions/Improvements are as follow:

1. The validation error report references index number of the column header instead of the column header itself. By using the readable column header, saves the task of cross referencing of the index to column header.
image

Information surrounding the file being uploaded such as the file id, genus, project and the column headers are readily available through the validator base loadAssets() method. 

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/217e1ca8402a8358bcfdd70356cb2aa9e706c706/trpcultivate_phenotypes/src/TripalCultivatePhenotypesValidatorBase.php#L55

In this case, the column headers are available in this property

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/217e1ca8402a8358bcfdd70356cb2aa9e706c706/trpcultivate_phenotypes/src/TripalCultivatePhenotypesValidatorBase.php#L31

$this->column_headers[0] // is the Trait Name
$this->column_headers[1] // is the Trait Description 
and so on.

There is also no reference what line number the error occurred, but I am assuming this will all be addressed in another PR.

2. In empty row check, given this row with double spaces for trait name cell, the validator returns a pass status.
```
$file_row = [
  '  ',
  'My trait description',
  'My method',
  'My method description',
  'My unit',
  'Quantitative'
];

$context['indices'] = [ 0 ];
$validation_status = $instance->validateRow($file_row, $context);

Try the trim() function and trim the value for empty value check.
```

3. When an empty row is provided, it triggers an error which is a good catch but the error is about indices more than the empty row.
 
image

4. In value in list validator, the validator disregards the case of the valid value, so for example when I set the valid value context array to ['Quantitative'], any case of the word Quantitative (quantitative, QUANTITATIVE, Quantitative) will pass. I think the case is important since in the instruction for the Import file is 

Type: One of "Qualitative" or "Quantitative". 

**Source code:**
DuplicateTraits.php
1. We can omit this line: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/7151772284c4334f7751b10aa35f4efdb16a3b09/trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php#L13
2. This service should be D. Injected
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/7151772284c4334f7751b10aa35f4efdb16a3b09/trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php#L107

I have example here:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/217e1ca8402a8358bcfdd70356cb2aa9e706c706/trpcultivate_phenotypes/src/Plugin/Validators/Genus.php#L100

EmptyCell.php
1. File class can go
2. The comment mentions of a unrelated variable $row 
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/7151772284c4334f7751b10aa35f4efdb16a3b09/trpcultivate_phenotypes/src/Plugin/Validators/EmptyCell.php#L50

ValueInList.php
1. File class can go
2. This should be using the key valid_values instead of indeces
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/7151772284c4334f7751b10aa35f4efdb16a3b09/trpcultivate_phenotypes/src/Plugin/Validators/ValueInList.php#L52

checkIndices()
1. Kindly expand more on the @param indices, I am confused since sometime it can be one dimensional array and sometimes an associative array.

carolyncaron:
Thank you Reynold 😊  I have addressed your suggestions specific to the source code in commit [ebede3a](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/77/commits/ebede3a7cef2567c11cf5852f269689a814ba9fb). Specifically:
1. I removed the use of the file class in all 3 validators (good catch!) and I also found I was including it unnecessarily in my test files as well (in addition, my test files used TripalLogger which also wasn't needed).
2. Switched to use drupal injection for the Traits service - thanks for your example, it made it easy!!
3. Both doc blocs in EmptyCell and ValueInList referred to undeclared variable $row, I changed this instead to be $row_Values
4. I didn't change the key of 'indices' at line 52 of ValueInList, but I tried to make the wording a bit more clear
5. I also tried to clarify the param $indices for `checkIndices()`. I can see how it can get confusing, but I think it does make more sense here that checkIndices() expects a one dimensional array, versus the associative array expected by each of the validators, since checkIndices only ever cares about the indices. Since checkIndices() is only ever called by a validator directly, the indices array gets passed in using the param $context['indices'] given to the validator. Let me know if this doesn't make sense!

carolyncaron:
Here are my comments to @reynoldtan's suggestions based on his manual test:
1. We have collectively decided to not address this in this PR, instead see Issue #82 for the start of our discussion regarding the future of the`loadAssets()` method.
2. Addressed in commit [ca00acc](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/77/commits/ca00acc36b5f28315016779b5a71eb51e614c504). Thanks for your suggestion! I also changed my unit test to include one of each in my file row: No whitespace, a single whitespace, and double whitespace.
3. This one took me a bit of time to think about, but I actually don't think I need to make any changes here. Here are my reasons:
    - I could include a quick check for an empty row in the `checkIndices()` method, but I think this method should strictly only be checking the indices, it only needs the row values to evaluate whether the indices make sense in context
    - I actually think this case should be handled at the importer level - because then what is the point of calling the validators on it? For example, in the Traits Importer currently, an empty file row gets skipped and thus none of the validators will see it
    - If anyone feels strongly that there should be a check somewhere for empty row (I guess in case the developer forgets to check in an importer 😅 ), then my suggestion is to create a new method for it in ValidatorBase. Potentially there might be other checks we haven't thought about yet specifically for $row_values? I may have talked myself into this option TBH, any thoughts by either @reynoldtan or @laceysanderson ?
4. Addressed in commit [9472dde](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/77/commits/9472dde4985c67ad68232d860775783186d89ee9) . Now, the ValueinList validator will do a straight check for valid values that is case sensitive. If there's no match, it does a second check for any case insensitive matching, which will set a flag for the purpose of formatting the title in validation_status. This was tricky to do well with the knowledge that return values for the validators will soon be changing (see issue #86). Let me know if you have any alternative suggestions! I also changed my unit tests to be case sensitive, and added a new test that checks for a case insensitive match and that the validation_status indeed reflects that.

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,151 @@
...
+ *   validator_name = @Translation("Duplicate Traits Validator"),
+ *   validator_scope = "FILE ROW",
+ * )
+ */
+class DuplicateTraits extends TripalCultivatePhenotypesValidatorBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * A nested array of already validated values forming the unique trait name +
+   *   method name + unit combinations that have been encountered by this
+   *   validator so far within the input file
```
laceysanderson: For arrays I always find it really helpful to fully lay out the keys and values. Especially when they are nested!

```suggestion
   *   validator so far within the input file. More specifically,
   *   - TRAIT-NAME: array of methods associated with this trait.
   *     - METHOD-NAME: array of units associated with this trait-method combo
   *       - UNIT-NAME: 1 (indicates this full trait-method-unit combo exists)
   *
   *   NOTE: All capital words are intended to be replaced by the actual full
   *   name of the term.
```

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,151 @@
...
+      $configuration,
+      $plugin_id,
+      $plugin_definition,
+      $container->get('trpcultivate_phenotypes.traits')
+    );
+  }
+
+  /**
+   * Validate the values within the cells of this row.
+   * @param array $row_values
```
laceysanderson, line 68: There should always be an empty line after the single line description and before the parameters.

```suggestion
   *
   * @param array $row_values
```

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,151 @@
...
+   *
+   * @return array
+   *   An associative array with the following keys:
+   *   - title: string, section or title of the validation as it appears in the result window.
+   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
+   *   - details: details about the offending field/value.
+   */
+  public function validateRow($row_values, $context) {
+
+    // Check our inputs - will throw an exception if there's a problem
```
laceysanderson:
```suggestion
    // Check the indices provided are valid in the context of the row.
    // Will throw an exception if there's a problem.
    // Typically checkIndices() doesn't take an associative array but
    // because it checks the values not the keys, it will work in this
    // case as well.
```

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -0,0 +1,151 @@
...
+    ];
+
+    return $validator_status;
+  }
+
+  /**
+   * Getter for the $unique_traits array.
+   *
+   * @return $unique_traits
+   *   - The array of unique trait name + method name + unit combinations
```
laceysanderson:
```suggestion
   *   The array of unique trait name + method name + unit combinations.
   *   More specifically, the array is as follows with capitalized words replaced 
   *   by the term name.
   *   - TRAIT-NAME: array of methods associated with this trait.
   *     - METHOD-NAME: array of units associated with this trait-method combo
   *       - UNIT-NAME: 1 (indicates this full trait-method-unit combo exists)

```

trpcultivate_phenotypes/src/Plugin/Validators/EmptyCell.php
```diff
@@ -0,0 +1,95 @@
...
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition
+    );
+  }
+
+  /**
+   * Validate the values within the cells of this row.
+   * @param array $row_values
```
laceysanderson, line 45:
```suggestion
   *
   * @param array $row_values
```

trpcultivate_phenotypes/src/Plugin/Validators/EmptyCell.php
```diff
@@ -0,0 +1,95 @@
...
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *   - title: string, section or title of the validation as it appears in the result window.
+   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
+   *   - details: details about the offending field/value.
+   */
+  public function validateRow($row_values, $context) {
+
+    // Check our inputs - will throw an exception if there's a problem
```
laceysanderson:
```suggestion
    // Check the indices provided are valid in the context of the row.
    // Will throw an exception if there's a problem
```

trpcultivate_phenotypes/src/Plugin/Validators/ValueInList.php
```diff
@@ -0,0 +1,112 @@
...
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition){
+    return new static(
+      $configuration,
+      $plugin_id,
+      $plugin_definition
+    );
+  }
+
+  /**
+   * Validate the values within the cells of this row.
```
laceysanderson, line 43:
```suggestion
   * Validate the values within the cells of this row.
   *
```

trpcultivate_phenotypes/src/Plugin/Validators/ValueInList.php
```diff
@@ -0,0 +1,112 @@
...
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *   - title: string, section or title of the validation as it appears in the result window.
+   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
+   *   - details: details about the offending field/value.
+   */
+  public function validateRow($row_values, $context) {
+
+    // Check our inputs - will throw an exception if there's a problem
```
laceysanderson:
```suggestion
    // Check the indices provided are valid in the context of the row.
    // Will throw an exception if there's a problem
```

trpcultivate_phenotypes/src/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -79,14 +79,59 @@ public function getValidatorName() {
...
    *   The validator plugin scope annotation definition value.
    */
   public function getValidatorScope() {
     return $this->pluginDefinition['validator_scope'];
   }
 
+  /**
+   * {@inheritdoc}
+   */
+  public function validateRow($row_values, $context) {
```
laceysanderson, line 93: This should throw an exception since it's a developer caused issue.

NOTE: No need to change this as it will be changed in issue #92 😜

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -55,7 +55,7 @@ class ValidTraitTest extends ChadoTestKernelBase {
    * Term config key to cvterm_id mapping.
    * Note: we just grabbed some random cvterm_ids that we know for sure exist.
    */
-  protected array $terms = [
+  protected $terms = [
```
laceysanderson: Just removes this change as I seem to remember it being editor focused and not needed.
```suggestion
  protected array $terms = [
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormTest.php
```diff
@@ -21,15 +23,20 @@ class TraitImporterFormTest extends ChadoTestKernelBase {
...
+  /**
+   * Chado connection
+   */
+  protected $connection;
+
   /**
    * Saves details regarding the config.
    */
-  protected array $cvdbon;
+  protected $cvdbon;
```
laceysanderson:
```suggestion
  protected array $cvdbon;
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormTest.php
```diff
@@ -21,15 +23,20 @@ class TraitImporterFormTest extends ChadoTestKernelBase {
...
 
   /**
    * The terms required by this module mapped to the cvterm_ids they are set to.
    */
-  protected array $terms;
+  protected $terms;
```
laceysanderson:
```suggestion
  protected array $terms;
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
```diff
@@ -21,15 +23,20 @@ class TraitImporterFormValidateTest extends ChadoTestKernelBase {
...
+  /**
+   * Chado connection
+   */
+  protected $connection;
+
   /**
    * Saves details regarding the config.
    */
-  protected array $cvdbon;
+  protected $cvdbon;
```
laceysanderson:
```suggestion
  protected array $cvdbon;
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
```diff
@@ -21,15 +23,20 @@ class TraitImporterFormValidateTest extends ChadoTestKernelBase {
...
 
   /**
    * The terms required by this module mapped to the cvterm_ids they are set to.
    */
-  protected array $terms;
+  protected $terms;
```
laceysanderson:
```suggestion
  protected array $terms;
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterRunTest.php
```diff
@@ -21,15 +23,25 @@ class TraitImporterRunTest extends ChadoTestKernelBase {
...
+  /**
+   * Config factory
+   */
+  protected $config_factory;
+
   /**
    * Saves details regarding the config.
    */
-  protected array $cvdbon;
+  protected $cvdbon;
```
laceysanderson:
```suggestion
  protected array $cvdbon;
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterRunTest.php
```diff
@@ -21,15 +23,25 @@ class TraitImporterRunTest extends ChadoTestKernelBase {
...
 
   /**
    * The terms required by this module mapped to the cvterm_ids they are set to.
    */
-  protected array $terms;
+  protected $terms;
```
laceysanderson:
```suggestion
  protected array $terms;
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/PluginValidatorTest.php
```diff
@@ -517,7 +518,9 @@ public function testColumnHeaderPluginValidator() {
    * Test TRAIT IMPORT VALUES Plugin Validator.
    */
   public function testTraitImportValuePluginValidator() {
-    $scope = 'TRAIT IMPORT VALUES';
+
+    $this->markTestSkipped();
```
laceysanderson:
```suggestion
    $this->markTestSkipped('Skipping this test until it can be replaced by the new validators testing.');
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorFileRowScopeTest.php
```diff
@@ -0,0 +1,173 @@
...
+  protected $chado;
+
+  /**
+   * Configuration
+   *
+   * @var config_entity
+   */
+  private $config;
+
+    /**
```
laceysanderson:
```suggestion
  /**
```

laceysanderson (dismissed):
Code-wise this looks great 😍 
All the suggestions are comment focused.

carolyncaron:
Thanks for your suggestions Lacey! I committed all of them in 3 batches 😄

reynoldtan:
All good, DI of trait service looks beautiful. One last thing I noticed when I do a file level duplicate check the message indicate the that the duplication is in the database not the in-file.

image

carolyncaron:
> All good, DI of trait service looks beautiful. One last thing I noticed when I do a file level duplicate check the message indicate the that the duplication is in the database not the in-file.
> 
> image

Thanks for catching this, Reynold! In this case, if a trait combo has been inserted into the database, ideally the validator should still recognize when a trait combo has been validated before (ie. it is a duplicated row in the input file) and tell the user about that too. I've updated the DuplicateTrait validator to account for this and added a test for this situation in commit [503879e](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/77/commits/503879e939ced6dc727e75b2700adea02c434a40)

laceysanderson (approved):
Code updates look good!

reynoldtan (approved):

PR #81, opened 2024-05-30
G4.78 fix trait service getters

### LINKED ISSUE #78: G4.78 - The get functions in the Traits Service do not handle more than one relationship between trait and method, or method and unit
### Branch

g4.78-fixTraitServiceGetters

### Groups

### Describe

The get functions in TripalCultivatePhenotypesTraitsService.php are not accommodating situations when we have multiple methods relating to a trait, or multiple units for a method. For example, the following Trait + Method + Unit combinations are all valid in the database at the same time:

> Trait: Plant height
> Method: Ruler
> Unit: cm
> 
> Trait: Plant height
> Method: Ruler
> Unit: inches
> 
> Trait: Plant height
> Method: barcode reader
> Unit: serial number

Thus, we need a way to get all method records for a single trait, and all units for a single method.

### Design

In general, situations in which more than one record is returned should be handled (including when it shouldn't be possible, which warrants an exception being thrown) rather than `LIMIT` being used in an SQL query. These are the lines in question:

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/526caaf5f97054c69b99449db564f40f90b89a8c/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php#L337

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/526caaf5f97054c69b99449db564f40f90b89a8c/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php#L388

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/526caaf5f97054c69b99449db564f40f90b89a8c/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php#L411

reynoldtan:
**Issue #78 **

## Motivation
Given the trait name, method short name and unit name. allow trait service to pull record for each item.

## What does this PR do?

1. Fetch the record for trait, method and unit by the trait name, method name and unit name (alternatively, by trait id, method id and unit id).

### Automated Testing

Includes a unit test to test:
1. Any of the parameters is set to an empty string.
2. Any of the parametesrs is set to 0.
3. A non-existent trait
4. A non-existent method
5. A non-existent unit
6. All parameters of the same type (string - name)
7. All parameters of the same type (integer - id number)
8. A mix of string and integer.

### Manual Testing
Add and configure a genus in your clone and in a PHP console or devel/php:
```
$service = \Drupal::service('trpcultivate_phenotypes.traits');
$service->setTraitGenus('Lens');

// Create trait, method and unit.
$trait = 'Your Trait Name';
$method = 'Your Method';
$unit = 'cm';

$combo = [
  'Trait Name' => $trait,
  'Trait Description' => 'Trait Description',
  'Method Short Name' => $method,    
  'Collection Method' => 'Pull the plant from the ground and measure', 
  'Unit' => $unit,
  'Type' => 'Quantitative'
];

$ins = $service->insertTrait($combo);
$my_combo = $service->getTraitMethodUnitCombo($trait, $method, $unit);
print_r($my_combo);

// print_r($ins);
// The insert trait method will return the inserted id number keyed by trait, method 
// and unit and can be used to test the combo method parameters type (fetch by id).
// Make sure to comment out the insert line to re-test using id numbers.

```
All 3 method parameters could take a string (name) or integer (id) values.

### DIFF
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
@@ -442,7 +442,7 @@ class TripalCultivatePhenotypesTraitsService {
     // Make sure that each parameter has a valid value.
     foreach([$trait, $method, $unit] as $param) {
       if (empty($param) || $param <= 0) {
-        return 0;
+        throw \Exception('The trait, method and unit passed into getTraitMethodUnitCombo() is required; however, one of them was empty.');
       }
     }
 
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
@@ -27,9 +27,6 @@ class TripalCultivatePhenotypesTraitsService {
   // Chado database connection.
   protected $chado;
 
-  // Tripal logger.
-  protected $logger;
-
   // Genus ontology configuration values.
   private $config = [];
 
@@ -44,15 +41,13 @@ class TripalCultivatePhenotypesTraitsService {
   /**
    * Constructor.
    */
-  public function __construct(TripalCultivatePhenotypesGenusOntologyService $genus_ontology, TripalCultivatePhenotypesTermsService $terms, ChadoConnection $chado, TripalLogger $logger) {
+  public function __construct(TripalCultivatePhenotypesGenusOntologyService $genus_ontology, TripalCultivatePhenotypesTermsService $terms, ChadoConnection $chado) {
     // Genus ontology service.
     $this->service_genus_ontology = $genus_ontology;
     // Terms service.
     $this->service_terms = $terms;
     // Chado connection.
     $this->chado = $chado;
-    // Tripal logger service.
-    $this->logger = $logger;
 
     // Terms configurations.
     // Terms focused on adding traits, required by this service.
@@ -161,7 +156,7 @@ class TripalCultivatePhenotypesTraitsService {
    *   trait, method, unit.
    *
    * @dependencies
-   *   getTraitAsset(), getMethodUnitDataType().
+   *   getPhenoCvterm(), getMethodUnitDataType().
    */
   public function insertTrait($trait, $schema = NULL) {
     // Configuration settings of the genus.
@@ -193,7 +188,7 @@ class TripalCultivatePhenotypesTraitsService {
     // Create trait: Trait, Method and Unit.
     foreach($arr_trait as $type => $values) {
       // Inspect cvterm to see if the trait asset already existed.
-      $trait_asset_rec = $this->getTraitAsset($values['name'], $type);
+      $trait_asset_rec = $this->getPhenoCvterm($values['name'], $type);
 
       if ($trait_asset_rec) {
         // Trait asset found, reference the record and re-use.
@@ -211,11 +206,9 @@ class TripalCultivatePhenotypesTraitsService {
         $ins = chado_insert_cvterm($rec, [], $schema);
         if (!$ins) {
           // Could not insert cvterm.
-          $this->logger->error('Error. Failed to insert term @type : @term.',
-            ['@type' => $type, '@term' => $values['name']],
-            ['drupal_set_message' => TRUE]
-          );
-          throw new \Exception(t('A database error occurred while inserting a term.'));
+          throw new \Exception(t('A database error occurred while inserting a term. Failed to insert term @type : @term.',
+            ['@type' => $type, '@term' => $values['name']]
+          ));
         }
 
         $arr_trait[ $type ]['id'] = $ins->cvterm_id;
@@ -235,29 +228,22 @@ class TripalCultivatePhenotypesTraitsService {
         $object  = $arr_trait['method']['id'];
 
         // Fetch all method(s) linked to the trait.
-        $asset_rec = $this->getTraitMethod($subject);
+        $related_terms = $this->getTraitMethod($subject);
       }
       else {
         $subject = $arr_trait['method']['id'];
         $object  = $arr_trait['unit']['id'];
 
         // Fetch all unit(s) linked to the method.
-        $asset_rec = $this->getMethodUnit($subject);
+        $related_terms = $this->getMethodUnit($subject);
       }
 
       $exists = FALSE;
-      if ($asset_rec) {
-        if (is_array($asset_rec)) {
-          foreach($asset_rec as $rec) {
-            if ($rec->cvterm_id == $object) {
-              $exists = TRUE;
-              break;
-            }
-          }
-        }
-        else {
-          if ($asset_rec->cvterm_id == $object) {
+      if (is_array($related_terms)) {
+        foreach($related_terms as $rec) {
+          if ($rec->cvterm_id == $object) {
             $exists = TRUE;
+            break;
           }
         }
       }
@@ -272,11 +258,10 @@ class TripalCultivatePhenotypesTraitsService {
           ->execute();
 
         if (!$ins_rel) {
-          $this->logger->error('Error. Failed to create term relationship @type : subject id - @subject object id - @object.',
-            ['@type' => $type, '@subject' => $subject, '@object' => $object],
-            ['drupal_set_message' => TRUE]
-          );
-          throw new \Exception(t('A database error occurred while inserting a term relationship.'));
+          throw new \Exception(t('A database error occurred while inserting a term relationship. 
+            Failed to create term relationship @type : subject id - @subject object id - @object.',
+            ['@type' => $type, '@subject' => $subject, '@object' => $object]
+          ));
         }
       }
     }
@@ -293,10 +278,9 @@ class TripalCultivatePhenotypesTraitsService {
         ->execute();
 
       if (!$ins_type) {
-        $this->logger->error('Error. Failed to insert unit data type @unit : @data_type.',
-          ['@unit' => $type, '@data_type' => $trait['Unit']], ['drupal_set_message' => TRUE]
-        );
-        throw new \Exception(t('A database error occurred while inserting a unit data type.'));
+        throw new \Exception(t('A database error occurred while inserting a unit data type. Failed to insert unit data type @unit : @data_type.',
+          ['@unit' => $type, '@data_type' => $trait['Unit']]
+        ));
       }
     }
 
@@ -321,10 +305,10 @@ class TripalCultivatePhenotypesTraitsService {
    *   Matching record/line in cvterm table or 0 if trait record Was not found.
    *
    * @dependencies
-   *   getTraitAsset()
+   *   getPhenoCvterm()
    */
   public function getTrait($trait) {
-    $trait_rec = $this->getTraitAsset($trait, 'trait');
+    $trait_rec = $this->getPhenoCvterm($trait, 'trait');
     return $trait_rec;
   }
 
@@ -336,17 +320,17 @@ class TripalCultivatePhenotypesTraitsService {
    *   is the trait id number (cvterm.cvterm_id).
    *
    * @return array
-   *   All matching records (object) in an array. 0 if no methods were found.
+   *   All matching records (object) in an array. An empty array if no methods were found.
    *   If there is only one result (method) returned access the value using index 0.
    *
    * @dependencies
-   *   getTraitAsset()
+   *   getPhenoCvterm()
    */
   public function getTraitMethod($trait) {
-    $trait_rec = $this->getTraitAsset($trait, 'trait');
+    $trait_rec = $this->getPhenoCvterm($trait, 'trait');
     if (!$trait_rec) {
       // Trait was not found.
-      return 0;
+      return [];
     }
 
     // Inspect the relationship table where the trait has a trait - method relationship.
@@ -370,7 +354,7 @@ class TripalCultivatePhenotypesTraitsService {
         ->fetchAll();
     }
 
-    return ($methods) ? $methods : 0;
+    return ($methods) ? $methods : [];
   }
 
   /**
@@ -385,13 +369,13 @@ class TripalCultivatePhenotypesTraitsService {
    *   If there is only one result (unit) returned access the value using index 0.
    *
    * @dependencies
-   *   getTraitAsset()
+   *   getPhenoCvterm()
    */
   public function getMethodUnit($method) {
-    $method_rec = $this->getTraitAsset($method, 'method');
+    $method_rec = $this->getPhenoCvterm($method, 'method');
     if (!$method_rec) {
       // Method was not found.
-      return 0;
+      return [];
     }
 
     // Inspect the relationship table where method has a method - unit relationship.
@@ -415,7 +399,7 @@ class TripalCultivatePhenotypesTraitsService {
         ->fetchAll();
     }
 
-    return ($units) ? $units : 0;
+    return ($units) ? $units : [];
   }
 
   /**
@@ -430,10 +414,10 @@ class TripalCultivatePhenotypesTraitsService {
    *   and was not set a data type.
    *
    * @dependencies
-   *   getTraitAsset()
+   *   getPhenoCvterm()
    */
   public function getMethodUnitDataType($unit) {
-    $unit_rec = $this->getTraitAsset($unit, 'unit');
+    $unit_rec = $this->getPhenoCvterm($unit, 'unit');
     if (!$unit_rec) {
       // Unit was not found.
       return 0;
@@ -451,13 +435,11 @@ class TripalCultivatePhenotypesTraitsService {
       ->fetchCol();
 
     if (count($data_type) > 1) {
-      // Unit appears to multiple data types.
-      $this->logger->error(
-        'Error. Failed to retrieve data type for unit : @unit in cv : @cv. Multiple data types found for the same unit.',
-        ['@unit' => $unit, '@cv' => $genus_config['unit']['name']],
-        ['drupal_set_message' => TRUE]
-      );
-      throw new \Exception(t('A multiple data type error occurred while retrieving a unit data type.'));
+      // Unit appears to have multiple data types.
+      throw new \Exception(t('A multiple data type error occurred while retrieving a unit data type. 
+        Failed to retrieve data type for unit : @unit in cv : @cv. Multiple data types found for the same unit.',
+        ['@unit' => $unit, '@cv' => $genus_config['unit']['name']]
+      ));
     }
 
     return ($data_type) ? reset($data_type) : 0;
@@ -480,25 +462,25 @@ class TripalCultivatePhenotypesTraitsService {
    *   0 if any one of trait, method or unit did not return any record.
    *
    * @dependencies
-   *   getTraitAsset(), getMethodUnitDataType().
+   *   getPhenoCvterm(), getMethodUnitDataType().
    */
   public function getTraitMethodUnitCombo(string|int $trait, string|int $method, string|int $unit) {
     // Trait.
-    $trait_rec = $this->getTraitAsset($trait, 'trait');
+    $trait_rec = $this->getPhenoCvterm($trait, 'trait');
     if (!$trait_rec) {
-      return 0;
+      return [];
     }
 
     // Method.
-    $method_rec = $this->getTraitAsset($method, 'method');
+    $method_rec = $this->getPhenoCvterm($method, 'method');
     if (!$method_rec) {
-      return 0;
+      return [];
     }
 
     // Unit.
-    $unit_rec = $this->getTraitAsset($unit, 'unit');
+    $unit_rec = $this->getPhenoCvterm($unit, 'unit');
     if (!$unit_rec) {
-      return 0;
+      return [];
     }
 
     // Append unit data type to the unit data object.
@@ -514,18 +496,18 @@ class TripalCultivatePhenotypesTraitsService {
   }
 
   /**
-   * Get trait asset - trait, method or unit.
+   * Get cvterm record for a trait, method or unit.
    *
    * @param string|int $key
    *   A string value is the name (cvterm.name), whereas an integer value
    *   is the id number (cvterm.cvterm_id).
    * @param string $type
-   *   trait, method or unit asset type. Trait is the default.
+   *   trait, method or unit type. Trait is the default.
    *
    * @return object
-   *   Trait asset record object or 0 if not found.
+   *   A trait, method or unit cvterm object.
    */
-  public function getTraitAsset($key, $type = 'trait') {
+  protected function getPhenoCvterm($key, $type = 'trait') {
     // Configuration check.
     $genus_config = $this->config;
     if (!$genus_config) {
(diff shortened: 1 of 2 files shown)


carolyncaron (commented):
To remove the deprecation notices from your unit tests, you can add the $config variable at line 33 of the ValidTraitTest class:
```
  /**
   * Configuration
   *
   * @var config_entity
   */
  protected $config;
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -381,16 +381,20 @@ public function getMethodUnit($method_id) {
     ];
 
     // Query unit.
-    $unit_id = $this->chado->query($sql, $args)
-      ->fetchField();
+    $unit_ids = $this->chado->query($sql, $args);
+    $sql = "SELECT * FROM {1:cvterm} WHERE cvterm_id = :id AND cv_id = :c_id LIMIT 1";
```
carolyncaron: The original issue brings up that "LIMIT 1" is very rarely a good idea. I suggest removing it here and also on the other 2 lines that were mentioned.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -414,4 +418,97 @@ public function getMethodUnitDataType($unit_id) {
...
+   * @return array
+   *   An associative array where the keys are trait, method and unit and the values
+   *   are the cvterm records for each key.
+   * 
+   *   0 if any one of trait, method or unit did not return any record.
+   * 
+   * @dependencies
+   *   getTrait(), getTraitMethod(), getMethodUnit() and getMethodUnitDataType().
+   */
+  public function getTraitMethodUnitCombo(string|int $trait, string|int $method, string|int $unit) {
```
carolyncaron, line 516: These parameters which are set up to take 2 different types is great! As we discussed in the office, it'd be really helpful to have the individual getter methods take these parameters as well. So:

- `getTrait(string|int $trait)`
- `getTraitMethod(string|int $trait)`
- `getMethodUnit(string|int $method)`

This way, `getTraitMethodUnitCombo` doesn't need to deal with creating an array to pass into each of the methods, or detecting int vs string, and all the getters are kept consistent.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -244,55 +244,164 @@ public function testTraitsServiceGetters() {
...
+      // Get trait method by trait id.
+      $m = $this->service_traits->getTraitMethod(['id' => $trait_id]);
+      $this->assertEquals($method_name, $m[0]->name, 'Trait method not found (by trait id).');
+
+      // Get trait method by trait name.
+      $m = $this->service_traits->getTraitMethod(['name' => $trait_name]);
+      $this->assertEquals($method_name, $m[0]->name, 'Trait method not found (by trait name).');
+
+      // Units - UnitABC and AnotherUnit.
+      $method_id = $method_id = $m[0]->cvterm_id;
```
carolyncaron: Variable setting seems a bit excessive here 😅 . Also, does `$method_id` need to be set on line 291 in this case?

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -244,55 +244,164 @@ public function testTraitsServiceGetters() {
...
+      ['Not Found Trait', $method_id, $unit_id], // Non-existent trait (mix string and id).
+      [$trait_id, 'Not found method', $unit_id], // Non-existent method (mix string and id).
+      [$trait_id, $method_id, 'Not found unit'], // Non-existent unit (mix string and id).
+    ];
 
-    $unit_name = $trait['Unit'];
-    $u = $this->service_traits->getMethodUnit($method_id);
-    $this->assertEquals($unit_name, $u->name, 'Trait method unit not found.');
-    $unit_id = $u->cvterm_id;
+    foreach($test_missing_param as $test) {
```
carolyncaron, line 543: I thought this was a really neat way to test this! 👍

carolyncaron (changes requested):

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
reynoldtan: This will require a major overhaul of the service and not to mention the need to revise all reference to each method in many places to comply with the new parameter setup. I will start a separate issue that will update this service that includes the limit clause pointed out. in the meantime, if you accept this method, your source code will not be affected by update.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
reynoldtan: That is a good catch, in both cases

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: I don't think this needs to be done in a different PR. Especially as it is only these 3 getters we are suggesting to change right now and they are only used within the service and the ValidTraitTest so it will not affect as much as you worry 💞

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: Yes, you should remove the limit in all cases in an API query. Also, if your query restricts by the primary key (i.e. `FROM {1:cvterm} WHERE cvterm_id = :id`) then there should not be any other parameters (i.e. no `AND cv_id = :c_id` in the above query).

In cases where you remove the limit and the primary key is not in the where clause, then you should use `fetchAll()` or `fetchCol()` to get all the results and then confirm there is only 1.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -414,4 +418,97 @@ public function getMethodUnitDataType($unit_id) {
...
+    // Make sure that each parameter has a valid value.
+    foreach([$trait, $method, $unit] as $param) {
+      if (empty($param) || $param <= 0) {
+        return 0;
```
laceysanderson: Anywhere there is an error reached in an API, an exception with a meaningful message should be thrown. I'm seeing a lot of `return 0;` throughout these getters and it makes debugging very difficult.
```suggestion
        throw \Exception('The trait, method and unit passed into getTraitMethodUnitCombo() is required; however, one of them was empty.');
```

reynoldtan: This one makes sense. Good catch!

laceysanderson: 🥰 Please update all the other `return 0;` instances in the getter functions 🙂

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -189,62 +211,93 @@ public function insertTrait($trait, $schema = NULL) {
...
+      if ($asset_rec) {
+        if (is_array($asset_rec)) {
+          foreach($asset_rec as $rec) {
+            if ($rec->cvterm_id == $object) {
+              $exists = TRUE;
+              break;
+            }
+          }
+        }
+        else {
```
laceysanderson: This should be removed as it will no longer be triggered. Instead, combine ln 249 and 250 into a single if with no else. This is because if we had anything returned from the getters it is an array and if not, then the method was not found in a reasonable way.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -189,62 +211,93 @@ public function insertTrait($trait, $schema = NULL) {
...
 
-    // Create relationships.
     foreach($arr_rel as $type => $rel) {
       // Check if relationship exists.
       if ($type == 'method-trait') {
         $subject = $arr_trait['trait']['id'];
         $object  = $arr_trait['method']['id'];
+
+        // Fetch all method(s) linked to the trait.
+        $asset_rec = $this->getTraitMethod($subject);
```
laceysanderson: This variable name is not very informative. Perhaps use something like `$related_terms` to keep that generic could be a method or a unit.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -168,17 +190,17 @@ public function insertTrait($trait, $schema = NULL) {
...
 
-    // Create trait.
+    // Create trait: Trait, Method and Unit.
     foreach($arr_trait as $type => $values) {
-      // Check if a record exists in the same Genus.
-      $value = ($type == 'method') ? $values['description'] : $values['name'];
-      $id = $this->chado->query($sql, [':value' => $value, ':id' => $genus_config[ $type ]['id']])
-        ->fetchField();
+      // Inspect cvterm to see if the trait asset already existed.
+      $trait_asset_rec = $this->getTraitAsset($values['name'], $type);
```
laceysanderson: As mentioned previously I find this naming very confusing and not very informative. I think it would greatly improve readability and maintainability to refer to the fact that this is returning a cvterm.
```suggestion
      $trait_cvterm = $this->getCvterm($values['name'], $type);
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -189,62 +211,93 @@ public function insertTrait($trait, $schema = NULL) {
         $ins = chado_insert_cvterm($rec, [], $schema);
         if (!$ins) {
           // Could not insert cvterm.
-          $this->logger->error('Error. Could not insert unit @key.', ['@key' => $values['name']]);
+          $this->logger->error('Error. Failed to insert term @type : @term.',
```
laceysanderson: There is no point logging an error if an exception is thrown IMO. Logging is meant for users and Exceptions are meant for developers to catch. Instead I would make the exception more informative personally.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +308,286 @@ public function insertTrait($trait, $schema = NULL) {
...
-   *   Matching record/line in cvterm table.
+   *   Matching record/line in cvterm table or 0 if trait record Was not found.
+   *
+   * @dependencies
+   *   getTraitAsset()
    */
   public function getTrait($trait) {
-    // Configuration settings of the genus.
-    $genus_config = $this->config;
+    $trait_rec = $this->getTraitAsset($trait, 'trait');
```
laceysanderson: Just adding a comment here to make it more clear why this function literally just calls another. 🤪 
```suggestion
    // Since the trait is simply a cvterm, we can use our helper method to retrieve it.
    $trait_rec = $this->getTraitAsset($trait, 'trait');
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +308,286 @@ public function insertTrait($trait, $schema = NULL) {
...
-      // Genus is not configured.
+  /**
+   * Get trait method.
+   *
+   * @param string|int $trait
+   *   A string value is the trait name (cvterm.name), whereas an integer value
+   *   is the trait id number (cvterm.cvterm_id).
+   *
+   * @return array
+   *   All matching records (object) in an array. 0 if no methods were found.
```
laceysanderson: I don't like that these methods return 0 but rather think it should just return an empty array. This also makes it easier to check when using these functions IMO. For example, returning an empty array from these getters shrinks your code in the insert when checking if the method/unit already exists substantially. For example,

```diff
- if ($asset_rec) {
-   if (is_array($asset_rec)) {
-     foreach($asset_rec as $rec) {
-       if ($rec->cvterm_id == $object) {
-         $exists = TRUE;
-         break;
-       }
-    }
- }
+ $exists = FALSE;
+ foreach ($asset_rec as $rec) {
+   if ($rec->cvterm == $object) {
+     $exists = TRUE;
+     break;
+
+ }
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +308,286 @@ public function insertTrait($trait, $schema = NULL) {
...
+
+    return [
+      'trait' => $trait_rec,
+      'method' => $method_rec,
+      'unit'  => $unit_rec
+    ];
+  }
+
+  /**
+   * Get trait asset - trait, method or unit.
```
laceysanderson: As mentioned above, I think it would be really helpful if this new method were clear that it is returning cvterms. I would switch any mention of trait asset or record in here to say cvterm instead since that is what it is :-)

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +308,286 @@ public function insertTrait($trait, $schema = NULL) {
...
+   * @param string|int $key
+   *   A string value is the name (cvterm.name), whereas an integer value
+   *   is the id number (cvterm.cvterm_id).
+   * @param string $type
+   *   trait, method or unit asset type. Trait is the default.
+   *
+   * @return object
+   *   Trait asset record object or 0 if not found.
+   */
+  public function getTraitAsset($key, $type = 'trait') {
```
laceysanderson: This method should be protected not public as it should not be used outside this service.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +308,286 @@ public function insertTrait($trait, $schema = NULL) {
...
+
+      if ($asset_rec && (int) $asset_rec[0]->cv_id != $genus_config[ $type ]['id']) {
+        // The id number seems to be of a trait asset that is outside of
+        // the cv: asset type (trait, method or unit) the genus was configured.
+        throw new \Exception(t('The requested trait asset @type : id @key CV value does not match the CV the genus was configured.',
+          ['@type' => $type, '@key' => $key]
+        ));
+      }
+    }
+    else {
```
laceysanderson: We had discussed first assuming an integer was a cvterm but if no cvterm had that id then treating the integer like a potential name. This block doesn't support that right now.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -236,63 +243,435 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    // name or id number as parameter using getTraitAsset() method.
+
+    // Test that by using id number and the id provided is an id for a trait asset that belongs
+    // to a cv not the cv set for the genus.
+
+    // term null with id 1 not in the genus.
+    /* Does not seem to catch and would kill the tests.
+    $exception_message = '';
+    try {
+      $this->service_traits->getTraitAsset(1, 'trait');
```
laceysanderson: This sounds like you have an outstanding bug as this test case should work, just as you say in the comment.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -236,63 +243,435 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    */
+
+    foreach($test_combo as $combo) {
+      foreach($keys as $key => $title) {
+        // Each combo, retrieve assets - trait, method and unit.
+
+        // By string parameter.
+        $asset = $this->service_traits->getTraitAsset($combo[ $title ], $key);
+        $asset_bystring_id = (int) $asset->cvterm_id;
+        $this->assertNotEquals($asset_bystring_id, 0, 'Failed to fetch trait asset' . $key . ' : ' . $combo[ $title ] . ' (by string name parameter).');
```
laceysanderson: It would be better to check that it returns the right id here instead. You should have this information from when it was inserted above, you just need to keep track above and then use it here. I like to keep an `$expected_cvterms` type array where the key is the name and the value is the id.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -236,63 +243,435 @@ public function testTraitsServiceDatabaseExpectations() {
...
+
+        // By id number parameter.
+        $asset = $this->service_traits->getTraitAsset($asset_bystring_id, $key);
+        $asset_byinteger_id = (int) $asset->cvterm_id;
+        $this->assertNotEquals($asset_byinteger_id, 0, 'Failed to fetch trait asset' . $key . ' : ' . $combo[ $title ] . ' (by integer id parameter).');
```
laceysanderson: Same comment as above.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -236,63 +243,435 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    $this->assertMatchesRegularExpression('/The requested trait asset (trait|method|unit)/', $exception_message,
+      'Trait asset id provided returned an asset not in the cv the genus was configured.');
+    */
+
+    foreach($test_combo as $combo) {
+      foreach($keys as $key => $title) {
+        // Each combo, retrieve assets - trait, method and unit.
+
+        // By string parameter.
+        $asset = $this->service_traits->getTraitAsset($combo[ $title ], $key);
```
laceysanderson: The getTraitAsset function should have been protected which means we don't test it directly but rather test the results of the other getters.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -236,63 +243,435 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    $a_trait_methods_byid   = $this->service_traits->getTraitMethod($a_trait->cvterm_id);
+
+    // Assert that in both cases, the returned set of methods was the same.
+    // Then proceed to assert that methods returned are the expected methods.
+    $this->assertEquals($a_trait_methods_byname, $a_trait_methods_byid,
+      'A Trait methods returned by methods getter with name and id as parameter do not match.');
+
+    // 3 methods in the set?
+    $methods_count = count($a_trait_methods_byid);
+    $this->assertEquals($methods_count, 3, 'A Trait methods returned by methods getter does not match expected count (3).');
```
laceysanderson: It is better to use `assertCount` here. That way if it fails the output shows the array that did not have 3 elements which is more informative.

laceysanderson (changes requested):

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: To be clear, I think the getTraitMethods() and getMethodUnits() should return empty array instead of 0. That is because they return sets of items.

carolyncaron:
***UPDATE***: Reynold suggested that I may have set Trait, Method and Unit to be all from the same CV during configuration of the module. It turns out that YES, I indeed had 🤦‍♀️ Once I set up a new docker instance and configured the three to be from different CVs, the problem was resolved. Reynold will make an issue to check for this in the config form, but we all agreed that nothing needs to be added to this PR except perhaps automated tests to try these scenarios 😄 

----

I re-tested my previous manual test cases and they are all working as expected :+1: However, I tried a new case that involved giving the trait, method and unit to getTraitMethodUnitCombo() in a different order than expected. It happily returned the records in the wrong order, as seen in this screenshot:

localhost_devel_php

Perhaps @laceysanderson could also weigh in on this, but I think I would expect to get nothing returned in this situation (as opposed to the trait, method, unit returned in the correct order since they do all have relationships between them). I think the Combo Getter should check if the first parameter is actually a trait, second parameter is actually a method, and the last parameter is actually unit. 

Also, I tried giving it the same string ($unit) for all 3 parameters and was also returned the same 3 records 😢 :
Screenshot 2024-06-07 at 12 20 52 PM

laceysanderson (commented):
I went through my original comments and all are resolved [except one which seems to have been missed](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/81#discussion_r1631496286).

Specifically, the `getPhenoCvterm()` method currently makes the assumption that if the key parameter passed is an integer then it must be the cvterm_id. While this is the most likely case, it would be good to check that it might be the name if there is not a term with that id.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +292,289 @@ public function insertTrait($trait, $schema = NULL) {
...
+    // Method.
+    $method_rec = $this->getPhenoCvterm($method, 'method');
+    if (!$method_rec) {
+      return [];
+    }
+
+    // Unit.
+    $unit_rec = $this->getPhenoCvterm($unit, 'unit');
+    if (!$unit_rec) {
+      return [];
```
carolyncaron: @laceysanderson and I were wondering whether it would be more appropriate for these if statements to return null rather than an empty array.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +292,289 @@ public function insertTrait($trait, $schema = NULL) {
...
+   * @param string|int $unit
+   *   A string value is the method unit name, whereas an integer value is the unit id number.
    *
-   * @return string
-   *   The data type of the unit, either quantitative or qualitative. False if not set.
+   * @return array
+   *   An associative array where the keys are trait, method and unit and the values
+   *   are the cvterm records for each key.
+   *
+   *   0 if any one of trait, method or unit did not return any record.
```
carolyncaron: Needs to be updated since it currently returns an empty array, not 0. See below comment!

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -255,163 +292,289 @@ public function insertTrait($trait, $schema = NULL) {
...
+   * Get cvterm record for a trait, method or unit.
+   *
+   * @param string|int $key
+   *   A string value is the name (cvterm.name), whereas an integer value
+   *   is the id number (cvterm.cvterm_id).
+   * @param string $type
+   *   trait, method or unit type. Trait is the default.
+   *
+   * @return object
+   *   A trait, method or unit cvterm object.
```
carolyncaron: Does not mention the return value if a record was not found

carolyncaron (changes requested):

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: Changed to NULL in most recent commit

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: Updated doc blocks to be very clear about return value and exceptions throw.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: Updated doc blocks to be very clear about return value and exceptions thrown.

laceysanderson (approved):
✅ Code review looks good now

I updated the code to
 - better document the return value and exceptions thrown
 - consistently return either null or [] for getters depending on whether they should return one or more values respectively.
 - added a check to ensure the key is either a string or a number for getPhenoCvterm()
 - updated test $chado property to $connection to match rest of test and test trait. NOTE: We will want to create another issue to use $chado throughout the testing environment for consistency but the change touched too many files and feels beyond the scope of this PR.

carolyncaron (approved):
I redid all of my testing on the latest changes, and it all works great! I also read through the documentation and all looks good to me as well.

Issue #83, opened 2024-06-10
[BUG] CV (trait, method and unit) configuration value should be unique for genus 


reynoldtan:
### Drupal Version

9.4

### Tripal Version

4.0.0-alpha1

### Module Version

4.x dev (most recent commit)

### Site Operating System

Debian Linux

### Installation Method

TripalDocker docker

### Browsers

Firefox

### Observations

Genus ontology configuration form allows for setting up the controlled vocabulary used for trait, method and unit to be used when creating cvterms. The form lacks a check to make sure that each CV value  used for trait, method and unit is unique controlled vocabulary. When trait, method and unit are set to the same CV trait service could return unexpected records as describe in this in this review comment.

**https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/81#issuecomment-2155314262**

### Expected Behaviour

See comments: **https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/81#issuecomment-2155314262**

### Steps to Reproduce

 See comments: **https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/81#issuecomment-2155314262**

PR #84, opened 2024-06-10
Fixes bug in ontology configuration form

reynoldtan:
**Issue #83 **

## Motivation
This bug was identified when fixing getter methods in Traits service as describe in this review comment.
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/81#issuecomment-2155314262

## What does this PR do?

1. Add a validation to inspect the CV values provided for Trait, Method and Unit field for each Genus and ensure that each one is a unique CV value.

### Automated Testing

Includes a test where CV for trait, method and unit field was set to the same value.

### Manual Testing

1. Access the Ontology and terms configuration form.
my site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
2. Set the CV for Trait, Method and Unit to the same CV value. It can be all 3 fields or any two fields.
image
3. A form error massage will trigger when all 3 fields or at least 2 fields have the same cv value.
image


carolyncaron (approved):
This works as expected! 🎉

laceysanderson (approved):
Code changes look good 👍

Issue #82, opened 2024-06-07
G4.82 Revisit the use of scope by the Validator API


carolyncaron:
### Branch

g4.82-[shortDescriptor]

### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Describe

@laceysanderson and I have had discussions regarding the use of scope by the validators. Prior to PR #77 and #79, each validator was assigned its own scope, but breaking up the previous TraitImportValues validator led to 3 validators under the same scope, which was renamed from `'TRAIT IMPORT VALUES'` to `'FILE ROW'`. While this change works for the purpose of those PRs, we'd like to evaluate our use of scope in general, and if we need a new way to categorize validators to streamline their use by the Traits and Phenotype importers.

Here are some things to consider:
- Changes to scope would impact the ValidatorBase class, and multiple tests
- The validators not touched by PR# 77 & 79 use the class method `validate()` which handles the file for each validator. The new validators instead use `validateRow()` which does not interact with the file directly. It would make the most sense to not do any file iteration within any of the validators, and instead do it one time only in the formValidate() method of the importer. But we don't want to implement validateRow() for every validator that doesn't apply to every row of the input file.
- We still want to categorize the validators based on the input they receive. The goal of this is the insulate each validator to access only the data it needs. This should negate the need for the loadAssets() method in the Validator Base class to provide some things that are unnecessary to every validator, such as Genus and Project.
- Finally, if loadAssets() is removed, we need a way to replace the purpose of the $skip parameter. This is what allows the Trait Importer to skip the validators that follow that are in a different scope. (Example: if there is an issue with the type of file uploaded, we don't want validation to continue on the file contents.)

### Design

Current scopes:
1. `GENUS`
2. `FILE`
3. `HEADERS`
4. `FILE ROW`

laceysanderson:
## Regarding my original vision for scope:

When I first brought this up I was picturing scoping as a grouping mechanisms where importers could give the same context to all validators in the same scope and new ones could be easily added without changes being made in the importer per se. For example, I pictured a row level scope where all validators of this type would just be given a row and know what to do with it. In this way, if we later developed a new row level validator we could just add its identifier to an array and the importer would happily add it to the list of things to check.

## My thoughts on the design I would like to see:

In hindsight, this concept might better be described as two variables. First an inputType which indicates what the importer should pass to the validator and perhaps a general category to group them further 🤔 

[...]

PR #94, opened 2024-06-27
G4.92 Update validator base, annotation and interface

### LINKED ISSUE #92: G4.92 Update annotation, interface and validator base with new design
### Branch

g4.92-updateValidatorBase

### Groups

### Describe

Issue #82 describes changes to the design of the validator api. This issue brings those changes into the annotation, interface and base validator class. It also marks the old design components as deprecated.

### Design

Do be expanded later.

laceysanderson:
**Issue #92**

## Motivation

Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/82 describes changes to the design of the validator api. This issue brings those changes into the annotation, interface and base validator class. It also marks the old design components as deprecated.

## What does this PR do?

1. Moves the validator plugin base into a dedicated directory and update namespace and class references.
2. Adds validateMetadata(), validateFile() and validateRow() to interface and base class.
3. Documents new return value and old one.
4. Adds inputTypes to annotation.
5. Adds getSupportedInputTypes() and checkInputTypeSupported() to support inputType usage.
6. Mark all deprecated functions and parameters with `@deprecated`

### Automated Testing

In the ValidatorBaseTest all of my new test methods create a validator developed specifically for testing the base class. This validator is found in `trpcultivate_phenotypes/tests/src/Kernel/Validators/FakeValidators/BasicallyBase.php` and simply extends the base class without implementing or overriding anything. As such, it is essentially the base class with added annotation 😜 

ValidatorBaseTest->testBasicValidatorGetters: tests retrieving the name and the configuration allow_new which indicates if new traits can be added.
- retrieves the name of the BasicallyBase validator using getValidatorName()and compares it to what I wrote in the annotation block.
- retrieves the status of the allow_new configuration and confirms it matches the default which is TRUE. Note that this is independent from the importer but already existed in the base class.

ValidatorBaseTest->testInputTypeValidatorGetters: tests the two new methods I added related to the inputType in the annotation. There's an example of the inputType annotation being used in the BasicallyBase and ValueInList validators.
- retrieves the value of the inputType in the BasicallyBase validator using getSupportedInputTypes(). This should return an array and we confirm that this array matches the one I wrote in the annotation. I use assertEqualsCanonicalizing to ensure that the order of the array does not cause the tests to fail.
- asks the BasicallyBase validator if it supports the `data-row` inputType and confirms that it knows it does.
- asks the BasicallyBase validator if it supports the `metadata` inputType and confirms that it knows it does NOT.
- asks the BasicallyBase validator if it supports the non-existant `SARAH` inputType and confirms that it knows it does NOT.

ValidatorBaseTest->testValidatorValidateMethods: tests the base implementation of the validate* methods. The base implementation is meant to simply inform the developer that the need to override it to support it. They're also there so that they can all be defined in the interface for consistency without requiring all validators to implement all of them.
- calls validateMetadata with random params matching the type hints and confirms we get an exception specific to validateMetadata.
- calls validateFile with random params matching the type hints and confirms we get an exception specific to validateFile.
- calls validateRow with random params matching the type hints and confirms we get an exception specific to validateRow.
- calls validate with random params matching the type hints and confirms we get an exception specific to validate. Note that the test for this method will change to check for the deprecation method once we get near the end of this design update.

#### Test Coverage

Screenshot 2024-06-28 at 9 55 30 AM

Screenshot 2024-06-28 at 9 55 01 AM

### Manual Testing

As this changes the interface, annotation and abstract base class it's quite hard to test 🤔 

Unfortunately the BasicallyBase plugin is not accessible in the drush php:cli so we can't just use the same approach as in the automated tests. In order to test cases I didn't or different values you will need to edit the automated tests. 

**Thankfully, we can test the ValuesInList validator in drush php:cli and I've updated it so it can be used as an example.**

For the annotation and such you can also test the ValuesInList validator as I updated it to show an example of the inputType. 

```php
$validator_id = 'trpcultivate_phenotypes_validator_value_in_list';

$instance->getValidatorName();
$instance->getSupportedInputTypes();
$instance->checkInputTypeSupported('data-row');
$instance->checkInputTypeSupported('metadata');
$instance->checkInputTypeSupported('SARAH');
```
Screenshot 2024-06-28 at 10 35 28 AM

Testing the validate functions you expect to see the exception for all but the validateValues() which this validator does override.

```php
$validator_id = 'trpcultivate_phenotypes_validator_value_in_list';

$form_values = ['genus' => 'Fred', 'project_id' => 123];
$instance->validateMetadata($form_values);

$filename = 'public://does_not_exist.txt';
$fid = 123;
$instance->validateFile($filename, $fid);

$row_values = ['col1', 'col2', 'col3', 'col4', 'col5'];
$context = [
  'indices' => [0, 2], 
  'valid_values' => ['col3']
];
$instance->validateRow($row_values, $context);

$instance->validate();
```
Screenshot 2024-06-28 at 10 42 33 AM

### DIFF
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
@@ -53,6 +53,6 @@ class TripalCultivatePhenotypesValidator extends Plugin {
    *  - header-row: for validating the first row in the file.
    *  - data-row: for validating all data rows in the file.
    */
-  public array $inputTypes;
+  public array $input_types;
 
 }
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
@@ -2,8 +2,8 @@
 
 /**
  * @file
- * Contains \Drupal\trpcultivate_phenotypes\Annotation\TripalCultivatePhenotypesValidator.
- * 
+ * Contains \Drupal\trpcultivate_phenotypes\TripalCultivateValidator\Annotation\TripalCultivatePhenotypesValidator.
+ *
  * @see Plugin manager in src\TripalCultivatePhenotypesValidatorManager.php
  */
 
@@ -13,23 +13,23 @@ use Drupal\Component\Annotation\Plugin;
 
 /**
  * Defines a data validator annotation object.
- * 
+ *
  * @see Drupal\trpcultivate_phenotypes\TripalCultivatePhenotypesValidatorManager
- * @see Drupal\trpcultivate_phenotypes\TripalCultivatePhenotypesValidatorInterface
- * 
+ * @see Drupal\trpcultivate_phenotypes\TripalCultivateValidator\TripalCultivatePhenotypesValidatorInterface
+ *
  * @Annotation
  */
 class TripalCultivatePhenotypesValidator extends Plugin {
   /**
    * The validator plugin ID.
-   * 
+   *
    * @var string.
    */
   public $id;
 
   /**
    * The validator human-readable name.
-   * 
+   *
    * @var string.
    */
   public $validator_name;
@@ -37,7 +37,7 @@ class TripalCultivatePhenotypesValidator extends Plugin {
   /**
    * The scope a validator will perform a check
    * ie. FILE level check, Project/Genus level check or Data Values level check.
-   * 
+   *
    * @var string.
    */
   public $validator_scope;
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
@@ -38,7 +38,21 @@ class TripalCultivatePhenotypesValidator extends Plugin {
    * The scope a validator will perform a check
    * ie. FILE level check, Project/Genus level check or Data Values level check.
    *
+   * @deprecated Remove in issue #91
+   *
    * @var string.
    */
   public $validator_scope;
+
+  /**
+   * The type of the data this validator supports validating.
+   *
+   * This should be one or more of the following:
+   *  - metadata: for validating the form values of the importer not including the file.
+   *  - file: for validating the file object but not it's contents.
+   *  - header-row: for validating the the first row in the file.
+   *  - data-row: for validating all data rows in the file.
+   */
+  public array $inputTypes;
+
 }
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
@@ -50,7 +50,7 @@ class TripalCultivatePhenotypesValidator extends Plugin {
    * This should be one or more of the following:
    *  - metadata: for validating the form values of the importer not including the file.
    *  - file: for validating the file object but not it's contents.
-   *  - header-row: for validating the the first row in the file.
+   *  - header-row: for validating the first row in the file.
    *  - data-row: for validating all data rows in the file.
    */
   public array $inputTypes;
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
@@ -7,7 +7,7 @@
  * @see Plugin manager in src\TripalCultivatePhenotypesValidatorManager.php
  */
 
-namespace Drupal\trpcultivate_phenotypes\Annotation;
+namespace Drupal\trpcultivate_phenotypes\TripalCultivateValidator\Annotation;
 
 use Drupal\Component\Annotation\Plugin;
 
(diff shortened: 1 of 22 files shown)


trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php
```diff
@@ -0,0 +1,58 @@
...
+   */
+  public $validator_scope;
+
+  /**
+   * The type of the data this validator supports validating.
+   *
+   * This should be one or more of the following:
+   *  - metadata: for validating the form values of the importer not including the file.
+   *  - file: for validating the file object but not it's contents.
+   *  - header-row: for validating the the first row in the file.
```
carolyncaron:
```suggestion
   *  - header-row: for validating the first row in the file.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -0,0 +1,196 @@
...
+   *
+   * @return array
+   *  An array of information about the validity of the data passed in.
+   *  The supported keys are:
+   *  - 'case': a developer code describing the case triggered
+   *      (i.e. no record in chado matching project name). If the data is
+   *      is valid then this is not required but could be 'data verified'.
+   *  - 'valid': a boolean indicating the data is valid (TRUE) or not (FALSE)
+   *  - 'failedIems': an array of the items that failed validation. For example,
+   *      if this validator validates a number of indicies are not empty then
```
carolyncaron:
```suggestion
   *  - 'failedItems': an array of the items that failed validation. For example,
   *      if this validator validates that a number of indices are not empty, then
```

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -0,0 +1,196 @@
...
```
carolyncaron:
```suggestion
   *  - 'failedItems': an array of the items that failed validation. For example,
   *      if this validator validates that a number of indices are not empty, then
```

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -0,0 +1,196 @@
...
```
carolyncaron:
```suggestion
   *  - 'failedItems': an array of the items that failed validation. For example,
   *      if this validator validates that a number of indices are not empty, then
```

carolyncaron (changes requested):
Code Review: 
Looks great - just some minor suggestions in some of the comments/docblocks 😄 

I also think you can throw out any changes made to `TraitImportValues.php` since this file will be deleted in PR #79. The tests for it should now be skipped in the 4.x branch

laceysanderson:
> I also think you can throw out any changes made to TraitImportValues.php since this file will be deleted in PR https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/79. The tests for it should now be skipped in the 4.x branch.

Yes, the tests should also be skipped in this branch as it's up to date with 4.x. It does feel silly but in order to ensure everything is consistent, those files do need to be updated as well until after PR #79 is merged. Plus I did a single find/replace in all files so it would have been more work to exclude.

carolyncaron (dismissed):
I followed the manual testing instructions and it worked for me! 🎉

reynoldtan:
Manual test worked for me too - fantastique. Code review:

**Validator Annotation**
1. There are 2 variable naming conventions used in trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php ie. $validator_name (underscore) and $inputTypes (pascal case). To be consistent and use only one rule moving forward, either convert $validator_name to $validatorName or $inputTypes to $input_type. I suggest the former.

**Validator Interface:**
1. The @file doc is missing.

2. The comments about failedItems for validateMetadata() and validateFile() are incorrect since both are not file-row level (no indices) validation. As discussed, it is a string value - the failed value as entered to that field element. 

2. I think this is no longer necessary - there is a good chance this will be overlooked later.
   *  The old style keys we are deprecating are:
   *  - title: the title of the validation (shown both when passes or fails).
   *  - details: string describing the failure to users with failed items embedded.
   *  - status: one of 'pass' or 'fail'
   
**Plugin Manager**
1. The @file doc is missing.

2.   public function getValidatorIdWithScope($scope) what might be the fate of this method? I think to deprecate

laceysanderson:
Thanks for the review @reynoldtan :-)

## Validator Annotation

> There are 2 variable naming conventions used in trpcultivate_phenotypes/src/TripalCultivateValidator/Annotation/TripalCultivatePhenotypesValidator.php ie. $validator_name (underscore) and $inputTypes (pascal case). To be consistent and use only one rule moving forward, either convert $validator_name to $validatorName or $inputTypes to $input_type. I suggest the former.

🙈 You're right and after confirming with Drupal standards it looks like it should be input_types. I've made that change in the annotation and everywhere the annotation is used.

## Validator Interface

> The `@file` doc is missing.

According to [Drupal standards for `@file`](https://www.drupal.org/docs/develop/standards/php/api-documentation-and-comment-standards#file) "Files that contain a namespaced class/interface/trait, whose file name is the class name with a .php extension, and whose file path is closely related to the namespace (under PSR-4 or a similar standard), SHOULD NOT have a @file documentation block."

> The comments about failedItems for validateMetadata() and validateFile() are incorrect since both are not file-row level (no indices) validation. As discussed, it is a string value - the failed value as entered to that field element.

See my expanded descriptions here: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/94/files/5eefa8b5d1518ef086c79c1a62aaa37580a58bd9..cde132109448e250834724ab0bc966e04d48de83. NOTE: In all cases the failedItems should be an array. The particular "stuff" in the array will be specific to the validator.

> I think this is no longer necessary - there is a good chance this will be overlooked later.

I think this is helpful when converting the old style to the new style. I have mentioned removing this in the issue related to cleanup and used the keyword deprecated to make it easier to find.

## Plugin Manager

> The `@file` doc is missing.

According to [Drupal standards for `@file`](https://www.drupal.org/docs/develop/standards/php/api-documentation-and-comment-standards#file) "Files that contain a namespaced class/interface/trait, whose file name is the class name with a .php extension, and whose file path is closely related to the namespace (under PSR-4 or a similar standard), SHOULD NOT have a @file documentation block."

> public function getValidatorIdWithScope($scope) what might be the fate of this method? I think to deprecate

Yes, it will be deprecated. I've added the deprecation tag to this function too :-)

reynoldtan:
Looking great. very detailed clarification of @file doc 👍

reynoldtan (approved):

laceysanderson:
Awesome! Thank you @reynoldtan and @carolyncaron 🎉

PR #95, opened 2024-06-29
G4.93 Validator Setters

### LINKED ISSUE #93: G4.93 Develop a number of setter traits for use with configuring validators.
### Branch

g4.93-validatorSetters

### Groups

### Dependencies

- #92 

### Describe

Right now there are two approaches born out of this concept of 'configure the validator but pass it the same thing within the same scope'.

First was @reynoldtan's take on this where he used `loadAssets()` to configure the module by giving it all the context it could ever need. This meets the goal of configuring once and allowed us to loop through the validators and call them all the same way. However, it also creates a lot of dependencies that are not needed and encouraged a single validator to move across multiple input types. This makes these validators a lot harder to share across importers.

Second was @carolyncaron's approach to pass in the context to the `validateRow($row, $context)` function. This meets the need of restricting access of the validator to unneeded information for less dependencies and easier re-use. However, it resulted in needing to call each validator within the same scope independently in order to set a different context for each. Hense we could no longer loop through them indiscriminately.

Hindsight makes me think maybe we would be best served by configuring the validator once at the top with a number of setters. Each instance/object of a validator could be configured differently in this way allowing the same validator to be used multiple times in a single importer. It would also mean we may not need to pass in additional context to the `validate*()` methods.

### Design

Lets consider the trait importer as an example usage since we are all familiar with it now.

I picture adding a new method to the importer class which handles configuring the validators. Here is some psudo-code on how that might look.

```php
/**
 * Configure all the validators this importer uses.
 * 
 * @return array
 *   A listing of configured validator objects first keyed by 
 *   their inputType. More specifically,
 *   - [inputType]: and array of validator instances. Not an 
 *     associative array although the keys do indicate what 
 *     order they should be run in.
 */
public function configureValidators() {
  $validators = [];

  // Metadata
  // Genus Exists and is configured
  $instance = $manager->createInstance('genusExists');
  $instance->setGenus('Tripalus');
  $validators['metadata']['genus_exists'] = $instance;
  // Genus matches the configured project
  $instance = use plugin manager to create instance of genusProjectMatch validator
  $instance->setGenus('Tripalus');
  $instance->setProjectId(1234);
  $validators['metadata']['genus_project'] = $instance;

  // ... 

  // File level
  $instance = $manager->createInstance('fileExists');
  $instance->setFileSuffix(['tsv', 'txt']);
  $instance->setMimeType('text/tab-separated-values');
  $validators['file']['file_exists'] = $instance;

  // do the same sort of thing for any other file level validators.
  // each validator would have its own set of setters although some 
  // could share setters across different validators that are 
  // defined in a parent class.

  // Header Level
  // All header row cells are not empty.
  $instance = $manager->createInstance('emptyCell');
  $instance->setIndicies([0,1,2,3,4,5]);
  $validators['header-row']['header_not_empty'] = $instance;

  // ... 

  // Data Row Level
  // All data row cells in columns 0,2,4 are not empty
  $instance = $manager->createInstance('emptyCell');
  $instance->setIndicies([0,2,4]);
  $validators['data-row']['data_not_empty'] = $instance;

  // ...

  $this->validatorObjects = $validators;

  return $validators; 
}
```

laceysanderson:
**Issue #93**

## Motivation

With issue #82 we developed a new design for the validators. This involves moving to configuring each validator in the importer configureValidators() method with a number of setters. Since many validators will use the same setters this PR sets up some php traits with common setters that can be shared among the validators.

## Current Outstanding

- [x] Automated tests for setFileMimeType(), getFileMimeType()
- [x] Header data provider refactoring of tests
- [x] DataFileDelimiter trait needs to go away
- [x] Any changes to file types trait based on Reynolds feedback
- [x] Any changes to header trait based on Lacey/Carolyn Review
- [x] Add todo comments above exceptions that we think should become logged messages and checked by a validator in another PR
- [x] Move FileType::getFileDelimiters() and FileType::mime_to_delimiter_mapping into validator base, make test for getFileDelimiters() and update splitRowIntoColumns to use mime_to_delimiter_mapping.

## What does this PR do?

1. Create various traits for supporting the various validators with providing the context that they need. 
   - [x] **GenusConfigured**: for supporting validators that need a genus configured with the phenotypes module in order to function
~~- [x] **DataFileDelimiter**: set the delimiter the importer will use to encode and split data value in the data file.~~
   - [x] **Project**: reference the project entered in the the Importer.
   - [x] **FileTypes**: for supporting validators that need the supported mime-types of an importer, and also those validators that need the specific mime-type of the input file
   - [x] **ColumnIndices**: for supporting validators that need to know which indices to validate in a file row
   - [x] **ValidValues**: for supporting validators that need to know what values are valid to check against, given a value as input in a file
   - [x] **Headers**: set the headers that will be used to reference a column in the data file. This trait will support validators that require all headers or subset of the headers defined. 
2. ~~Update all tests that use the PhenotypeImporterTestTrait test trait to use chado_connection to meet the new standard and allow me new test to use this new standard ;-p~~ Addressed in PR #102 
3. Create automated tests to cover the new validator traits.

--------------------------------------------------------------
**Setup Project Setter/Getter**
1. Create project setter method
2. Create project getter method
3. Create automated test - test both setter and getter methods.

```
to set:
$this->setProject(123) by project id or $this->setProject('AGILE: Application of Genomics Innovation in Lentil Economy') by project name.
to get:
$project = $this->getProject()

$project is:
array(
  'id' => 123,
  'name' => 'AGILE.....'
)
```

------------------------------------------------------------------
**Setup File Type**
1. Create file type setter method
2. Create file type getter method
3. Create automated test - test both setter and getter methods.

``` 
to set:
$this->setFileTypes(['tsv', 'txt']) to support both tsv and txt

to get:
$file_type = $this->getFileTypes()

$file_type is 
array(
 'tsv' => 'text/tab-separated-values',
 'txt' => 'text/plain'
)
```

------------------------------------------------------------------
**Setup Headers**
1. Create headers setter method
2. Create required header type getter method
3. Create optional header type getter method
4. Create general-purpose header getter by specifying a type or group of types
5. Create automated test - test both setter and getters methods.

```
to set:

// Define headers for the importer.
$headers = [
  [
    'name' => 'Header Name',
    'description' => 'Header Description',
    'type' => 'required' // required or optional
  ]
  ...
];

$this->setHeaders($headers);

to get:

// All headers with required type set
$required_headers = $this->getRequiredHeaders();

// All headers with optional type set
$optional_headers = $this->getOptionalHeaders();

// Get specific type or combination of types

$required_and_optional = $this->getHeaders();

$optional = $this->getHeaders(['optional']);

$optional_and_required = $this->getHeaders['optional', 'required']);

```

### Automated Testing

ValidatorTraitGenusConfiguredTest->testSetter: tests **setConfiguredGenus()**
- [x] With a valid case where the genus has been fully configured with phenotypes and the validator is set up properly.
- [x] Throws an exception if the genus is in chado but not configured.
- [x] Throws an exception if the genus is not even in chado.
- [x] That the expected cv/db configuration was actually saved for the validators use
- Don't worry about these for now, but potentially test for in the future:
  - Throws an exception if the validator doesn't setup the chado_connection property.
  - Throws an exception if the validator doesn't setup the service_PhenoGenusOntology property.

tests **getConfiguredGenus()**
- [x] That the configured genus was actually saved the validators use.

-----------------------------------------------------------------------
~~**Data File Delimiter:**~~
~~ValidatorTraitDelimiterTest - testDelimiterSetterGetter()~~
~~- [x] Test that the getter will fire an exception message prior to calling a setter method to set a delimiter.~~
~~- [x] Test invalid delimiters.~~
~~- [x] Test valid delimiters.~~
~~- [x] Test delimiter getter.~~

-----------------------------------------------------------------------
**Project:**
ValidatorTraitProjectTest - **testProjectSetterGetter()**
- [x] Test that the getter will fire an exception message prior to calling a setter method to set a project.
- [x] Test invalid project.
- [x] Test non-existent project by project id (integer parameter type).
- [x] Test non-existent project by project name (string parameter type).
- [x] Test valid and existing project by project id (integer parameter type).
- [x] Test valid and existing project by project name (string parameter type).

-----------------------------------------------------------------------
**FileTypes:**
ValidatorTraitFileTypesTest - **testSupportedMimeTypes()**
- [x] Test that the getter will fire an exception message prior to calling a setter method to set file types.
- [x] Test no file types are provided (empty array)
- [x] Test combinations of unsupported file types: `[jpg, gif, svg], [png, pdf], [gzip]`
- [x] Test combinations of supported file types: `[tsv], [csv], [txt], [tsv, txt], [csv, txt]`
- [x] Test a combination of supported and unsupported types: `[tsv, 'jpg']`
- [x] Test that getSupportedMimeTypes() returns the correct mime type value for valid types, and throws an exception in the presence of an invalid type.

**testFileMimeType()**
- [x] Test that the getter will fire an exception message prior to calling a setter method to set a file type.
- [x] Test an empty string
- [x] Test the following valid mime-types: `text/tab-delimited-values`, `text/csv`, `text/plain`, and `hello world`
- [x] Test that getFileMimeType() returns the correct mime-type for valid types, and throws an exception in the presence of an invalid type.

-----------------------------------------------------------------------
**ColumnIndices:**
ValidatorTraitColumnIndicesTest - testColumnIndicesSetterGetter()
- [x] Test the getter method throws an exception when called prior to the setter method
- [x] Test the setter method throws an exception when given an empty array of indices
- [x] Test the setter method throws an exception for a combination of invalid indices
  - [x] Given a nested array
  - [x] Given an array value that is an object
  - [x] Given an array of float values
- [x] Test combinations of valid indices

-----------------------------------------------------------------------
**ValidValues:**
ValidatorTraitValidValuesTest - testValidValuesSetterGetter()
- [x] Test the getter method throws an exception when called prior to the setter method
- [x] Test the setter method throws an exception when given an empty array of values
- [x] Test the setter method throws an exception for a combination of invalid indices
  - [x] Given a nested array
  - [x] Given an array value that is an object
  - [x] Given an array of float values
- [x] Test combinations of valid values

-----------------------------------------------------------------------
**Headers:**
ValidatorTraitHeadersTest - testHeadersSetterGetter()
- [x] Test exception thrown when calling any of the getters prior to calling the setter method
- [x] Test parameter to the setter method cannot be an empty array
- [x] Test that each header must pass key and value check
  - [x] Name and Type key must be set.
  - [x] Name and Type key must have a value
  - [x] Type key value must be one of the set or valid header types defined by the trait
 -[x] Test that for a headers with required and optional header types, the setter is able to set the correct context array element and any getter will be able to pull specific or combination of header types.
- [x] Test the headers getter with invalid types
- [x] Test that the setter and getter methods maintain the index (order) of the headers as each appeared in the headers definition. 

### Manual Testing


trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/DataFileDelimiter.php
```diff
@@ -0,0 +1,52 @@
...
+   * @param string $delimiter
+   *   A value used to separate/delimit the values in a data file row or line. 
+   * 
+   * @throws \Exception
+   *   Invalid delimiter provided such as empty string, false and 0 values.
+   * 
+   * @return void
+   */
+  public function setDelimiter(string $delimiter) {
+    // Delimiter must have a value.
```
carolyncaron: I'm worried that if $delimiter is truly an empty string, then the current exception message won't reveal the actual problem. Instead, I would recommend a separate check for an empty string with an explicit exception message. Make sure to also update the test for this case!
```suggestion
    // Delimiter must not be an empty string
    if($delimiter === '') {
      throw new \Exception('The DataFileDelimiter Trait requires a non-empty string as a data file delimiter.');
    }
  
    // Delimiter must not be false, 0 or null
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,82 @@
...
+   * 
+   * @throws \Exception
+   *   types array is an empty array or string.
+   * 
+   * @return void
+   */
+  public function setFileTypes(array $types) {
+    // Types must have a value.
+    if (empty($types)) {
+      throw new \Exception('No file type provided.');
```
carolyncaron: Ideally the exception message would be a bit more informative. I've been using this as a template based on Lacey's exception messages for GenusConfigured:
`The [trait name] Trait requires a non-empty array of [values].`

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,82 @@
...
+   *   file media type (mime), keyed by file type.
+   * 
+   * @throws \Exception
+   *   If the 'file_type' key does not exist in the context array (ie. the file_type
+   *   array has NOT been set).
+   */
+  public function getFileTypes() {
+    // The trait key element file_type should exists in the context property.
+    if (!array_key_exists($this->trait_key, $this->context)) {
+      throw new \Exception("Cannot retrieve file types set by the importer.");
```
carolyncaron: It would be helpful to mention which set method is needed to be called first in order to use the getFileTypes() method. Something like:
```suggestion
      throw new \Exception("Cannot retrieve file types as none have been set by the setFileTypes() method.");
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,82 @@
...
+    }
+    
+    // File extension and file media type (mime type).
+    $file_type_to_mime = [
+      'tsv'  => 'text/tab-separated-values',
+      'csv'  => 'text/csv',
+      'txt'  => 'text/plain',
+
+      // Other file types here, ensure file types
+      // support delimiter.
```
carolyncaron:
```suggestion
      // Add any additional valid file types here, while ensuring file types support use of a delimiter.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/DataFileDelimiter.php
```diff
@@ -0,0 +1,56 @@
...
+   *   The delimiter value set by the importer.
+   * 
+   * @throws \Exception
+   *   If the 'delimiter' key does not exist in the context array (ie. the delimiter
+   *   array has NOT been set).
+   */
+  public function getDelimiter() {
+    // The trait key element delimiter should exists in the context property.
+    if (!array_key_exists($this->trait_key, $this->context)) {
+      throw new \Exception("Cannot retrieve delimiter set by the importer.");
```
carolyncaron: This message should also mention the setDelimiter() method

carolyncaron (commented):
Here are some suggestions for the validator traits that @reynoldtan has been working on. Looks great so far! These are just a few suggestions mostly pertaining to exception messages. The Project trait exceptions should also be updated accordingly.

reynoldtan:
Based on the conversation I had with @carolyncaron, we came up with this solution for the headers setter/getter.

The headers property in the importer that defines the headers will now contain information about the type of header ie. required, optional

```
Current array setup:
$headers = [
  'Header 1' => 'Header 1 text description',
  'Header 2' => 'Header 2 text description'
];

To:

$headers_change_me = [
  [
    'name' => 'Header 1',
    'description' => 'Header 1 text description'
    'type' => 'required',
  ],
  [
    'name' => 'Header 2',
    'description' => 'Header 2 text description'
    'type' => 'optional',
  ],
];

````

The file text description of the importer form can use the type information to indicate requisite headers from optional.

Setters and getters:

``` 
/**
 * Sets headers.
 *
 * @param $headers
 *   The headers array defined by the importer.
 */
public function setHeaders($headers) {
  // Parameter checks.

  // Create a list of types expected, each type will have a dedicated getter below.
  $types = [
    'required',  // All headers required - must have a value. 
    'optional',  // All optional headers - may have a value.
    
    // Add other types here and implement a getter.
   ];
  
  // For each types, pull all of the headers using the type key for comparison.
  // Create an element in context array header using the type as the key ie.
  
  // Construct each array subset and check if required has a value.
  $required_headers = All headers where type is required.
  // Throw an exception if required array is empty.
  $optional_headers = All headers where type is optional.
  
  // With the array, set context array.
  $this->context['headers']['optional'] = $optional_headers;
  $this->context['headers']['required'] = $required_headers

  // Each type array now consists of an array of headers keyed by the index (order) the
  // header appeared in headers ie.
  $optional_headers = [
    4 => 'Header 5',
    5 => 'Header 6'
  ];

  $this->context['headers']['optional'] = $optional_headers;
  
  // Header 5 and Headers 6 were defined in index 4 and index 5 of the headers
  // parameter (defined in the importer).

  // Each subset will omit the type and description information.
  
  // Set a context value to reference all of the headers. Keyed- all
  $this->context['headers']['all'] = All the headers.
}
```

```
// Getters for all required headers.
public function getRequiredHeaders() {
  // Exception if required was not set  

  return $this->context['headers']['required'];
}

// Getters for all optional headers.
public function getOptionalHeaders() {
  // Exception if optional was not set

  return $this->context['headers']['optional'];
}

// Getters for all headers.
public function getAllHeaders() {
  // Exception if all were not set

  return $this->context['headers']['all'];
}

```

The template generator will have to be adjusted to understand the new headers structure.

File: trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitProjectTest.php [hunk 6]
```diff
@@ -0,0 +1,232 @@
...
+   */
+  protected ValidatorProject $instance;
+
+  /**
+   * Test projects with reference to project id and project name
+   * keyed by id and name, respectively.
+   * 
+   * @var array
+   */
+  protected array $test_projects;
```
carolyncaron: There's a typo here which is causing the PHP deprecation notice
```suggestion
  protected array $test_project;
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
```diff
@@ -0,0 +1,61 @@
...
+namespace Drupal\trpcultivate_phenotypes\TripalCultivateValidator\ValidatorTraits;
+
+/**
+ * Provides setters focused on configuring a validator to look at particular
+ * column within a row of a provided file
+ */
+trait ColumnIndices {
+
+  /**
+   * Sets the array of indices for which the validator to do its checks on.
```
laceysanderson: I think more detail explaining this would be helpful. What do you think of this?
```suggestion
   * Sets the array of indices for the validator to do its checks on.
   *
   * Specifically, in this context, an index is the key of the column the
   * validator should check. In the case of lists, this index/key is a sequential
   * integer created by PHP and in the case of associative arrays, this index/key
   * will likely be a meaningful string (such as names of column headers).
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param array $indices
+   *   An array of integers or keys which correspond to columns in a row of
+   *   delimited values that the validator instance should act on
```
laceysanderson: This description is still a bit inclear to me. What do you think of this?

```suggestion
   *   An array where each value is the key of the column the validator instance should
   *   act on. It must be either an integer or string. For more detail, see the
   *   definition of indices in the docblock above.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
```diff
@@ -0,0 +1,61 @@
...
+   *  - If $indices array is multi-dimensional or contains objects
+   */
+  public function setIndices(array $indices) {
+
+    // Make sure we don't have an empty array
+    if(count($indices) === 0) {
+      throw new \Exception('The ColumnIndices Trait requires a non-empty array of indices.');
+    }
+
+    // Check if we have a multidimentsional array or array of objects
```
laceysanderson: It's probably good to have another check here to ensure it is a string or integer :thinking: or maybe check that instead of array/object? That way it follows the docblock in what we expect.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
```diff
@@ -0,0 +1,61 @@
...
+    // Set the indices array
+    $this->context['indices'] = $indices;
+  }
+
+  /**
+   * Returns an array of integers or keys which correspond to columns in a row
+   * of delimited values that the validator instance should act on.
+   *
+   * @return array
+   *   A one-dimensional array containing column indices or column names
```
laceysanderson:
```suggestion
   *   A one-dimensional array containing column indices. If the array to be validated
   *   is a list then these will be sequential integers and if it's an associative
   *   array they will be the string keys set by the developer.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/DataFileDelimiter.php
```diff
@@ -0,0 +1,62 @@
...
+trait DataFileDelimiter {
+
+  /**
+   * The key used by the setter method to create a delimiter element 
+   * in the context array, as well as the key used by the getter method 
+   * to reference and retrieve the delimiter element value.
+   * 
+   * @var string
+   */
+  private string $trait_key = 'delimiter';
```
laceysanderson: I find the usage of trait very confusing in variable names unless we are referring to phenotypic traits. As such, I would suggest using context_key instead.


```suggestion
  private string $context_key = 'delimiter';
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
```diff
@@ -0,0 +1,61 @@
...
+      }
+    }
+
+    // Set the indices array
+    $this->context['indices'] = $indices;
+  }
+
+  /**
+   * Returns an array of integers or keys which correspond to columns in a row
+   * of delimited values that the validator instance should act on.
```
laceysanderson: What about this wording instead to be consistent with what I suggested above?

```suggestion
   * Returns the indices of the columns this validator should act on.
   *
   * Specifically, in this context, an index is the key of the column the  
   * validator should check. In the case of lists, this index/key is a sequential  
   * integer created by PHP and in the case of associative arrays, this index/key  
   * will likely be a meaningful string (such as names of column headers).  
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,90 @@
...
+   * The key used by the setter method to create a file types element 
+   * in the context array, as well as the key used by the getter method 
+   * to reference and retrieve the file types element value.
+   * 
+   * @var string
+   */
+  private string $trait_key = 'file_types';
+
+  /**
+   * Sets a file types with the file media type.
```
laceysanderson: This wording is not very clear to me. What about this?

```suggestion
   * Sets the file extensions and associated mime-types that a validator checking the
   * file metadata should ensure the file is one of.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * @param array $extensions
+   *   An array of file extensions (file types) the importer expects the file extension of the data file.
```
laceysanderson:
```suggestion
   *   An array where the values are supported file extensions (e.g. tsv ) for validating a data file.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,90 @@
...
+   */
+  public function setFileTypes(array $extensions) {
+    
+    // Extensions array must have a element.
+    if (empty($extensions)) {
+      throw new \Exception('The File Types Trait requires an array of file extensions and must not be empty.');
+    }
+    
+    // File extension and file media type (mime type).
+    $file_type_to_mime = [
```
laceysanderson: This should be a private variable near the top of the class to make it easier to add to.

I'm also still not sure about this being a one-to-one mapping. It might be best to be a extension => array of mime types mapping to support future cases where multiple mime types apply.

For example,
```php
private array $mime_type_mapping = [
  'tsv'  => ['text/tab-separated-values'],
  'csv'  => ['text/csv'],
  'txt'  => ['text/plain'],
];
```

laceysanderson (commented):
I've made it through the ColumnIndices, DataFileDelimter and FileTypes traits with comments added in this review.

I'm making a note to chat more about the design behind file extensions + mime types as I'm still not sure this should be a one-to-one relationship.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,90 @@
...
+    $file_types = [];
+    $unresolved = [];
+
+    foreach($extensions as $type) {
+      if (!isset($file_type_to_mime[ $type ])) {
+        array_push($unresolved, $type);
+        continue;
+      }
+      
+      $file_types[ $type ] = $file_type_to_mime[ $type ];
```
laceysanderson: Instead of saving a mapping of supported mime types, what if we save supported mime types separately from supported file extensions. This makes it easier to use these in your validator.

Specifically we would do `$mime_types = array_merge($mime_types, $file_type_to_mime[$type]);` here and then at the end of this setter we would have `$this->context['mime_types'] = $mime_types` and `$this->context['file_extensions'] = $file_types;`

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,90 @@
...
+   *
+   * @return array
+   *   The file types set by the setter method. The types include the file extension (type) and 
+   *   the corresponding file media type (mime), keyed by file extension.
+   * 
+   * @throws \Exception
+   *  - If the 'file_types' key does not exist in the context array
+   *    (ie. the file_types array has NOT been set).
+   */
+  public function getFileTypes() {
```
laceysanderson: I think it would be bedtter to have two separate getters here: getSupportedMimeTypes() and getSupportedFileExtensions(). This would map to the two context keys I suggested above and make it more readible in your validator when checking these.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/DataFileDelimiter.php
```diff
@@ -0,0 +1,62 @@
+<?php
+
+namespace Drupal\trpcultivate_phenotypes\TripalCultivateValidator\ValidatorTraits;
+
+/**
+ * Provides setters focused for setting a delimiter used by the importer
+ * to delimit data/values in the data file, and getter to retrieve the set value.
+ */
+trait DataFileDelimiter {
```
laceysanderson: We just had a great in-person discussion regarding this trait that I am going to summarize here.

Action Items (on hold until after split row PR is merged):
1. This trait is going to disappear and the functionality is going to move into the fileTypes trait.
2. The fileTypes trait will have a mapping of MimeType => delimiter (e.g. 'text/tab-separated-values' => "\t") stored in a static variable.
3. There will be a getSupportedFileDelimiters() method that returns the list of delimiters supported by the set mime-types to use in validators.
4. The splitRowIntoColumns(string $line, string $mime_type) static method will use the mapping in action item 2 to split the line based on the mime_type passed in.
5. There won't be a setDataFileDelimiter() as the supported delimiters will also be set in the setFileTypes() with lookup being based on the mime types.
6. Instead there will be a static method to get the delimiter for a specific mime type (this can be used by both the static `splitRowIntoColumns()` and the validator checking the line format using `validateRawRow()`).

Discussion:
When discussing the fact that we would like to support multiple delimiters, it became apparent that this is also heavily integrated with the file extensions and mime types. In order to ensure that someone doesn't set a delimiter that is not correct for the mimetype/extension set with another setters, we've decided to let the single setter set all 3 (i.e. extensions, mime types, delimiters).

The discussion also passed into how we would use the multiple delimiters to split the row. While we want to use the `str_getcsv()` method to split the row, this only supports a single delimiter. We could loop through the supported delimiter until we find one that "Works" but we're worried that would be error prone and also a performance concern. The solution to this is to have the importer tell the splitRowIntoColumns() the mime type of the current file (which is can grab easily via the Drupal File object) and then this method can split specifically based on the type of delimiter supported for this mime type. Also when we validate the file row format using the validateRawRow() we will be able to use the same approach (i.e. get the delimiter for the specific mime type and then confirm the row uses it).

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ColumnIndices.php
carolyncaron: Thanks for the suggestion, it does make a lot more sense 😅 I made this change in commit [0593edc](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/95/commits/0593edca7a34540a1ed69fda5821a1a6ad3144bf)

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Project.php
```diff
@@ -0,0 +1,94 @@
...
+      // Project id number.
+      if ($project <= 0) {
+        throw new \Exception('The Project Trait requires project id number to be a number greater than 0.');  
+      }
+
+      // Look up the project id to retrieve the project name.
+      $project_rec = ChadoProjectAutocompleteController::getProjectName($project);
+      
+      $set_project = [
+        'id' => $project,       // Id
```
laceysanderson: The key here should be `project_id` not just `id`. That is because id here is ambiguous as it could refer to an entity_id.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ValidValues.php
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Provides setters focused on configuring a validator that processes the columns
+ * in a file row and checks the value in the column against a list of valid values
+ */
+trait ValidValues {
+
+  /**
+   * Sets the array of values for which the validator is checking for in a
+   * specified set of columns (set by another validator trait) within a file row
```
laceysanderson: What about this wording:


```suggestion
   * Sets a list of values that should be allowed for a cell being validated.
   *
   * Note: Often this setter is combined with the setIndices() setter which indicates
   * which columns should have one of these values.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/ValidValues.php
```diff
@@ -0,0 +1,62 @@
...
+      }
+    }
+
+    // Set the valid_values array
+    $this->context['valid_values'] = $valid_values;
+  }
+
+  /**
+   * Returns an array of valid values which are permitted to be in specified
+   * columns of a file row that the validator instance acts on.
```
laceysanderson: What about this wording:


```suggestion
   * Returns a list of allowed values for cell(s) being validated. Specifically, it
   * is expected that the cell must contain one of the values in this list.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Project.php
```diff
@@ -0,0 +1,94 @@
...
+   * The key used by the setter method to create a project element 
+   * in the context array, as well as the key used by the getter method 
+   * to reference and retrieve the project element value.
+   * 
+   * @var string
+   */
+   private string $trait_key = 'project';
+
+  /**
+   * Sets a project.
```
laceysanderson: What about this wording?


```suggestion
   * Sets a single project for use by a validator.
   *
   * Note: This project must exist in the chado.project table already and as long as it
   * does both the project_id and project name will be saved for use by the validator.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Project.php
```diff
@@ -0,0 +1,94 @@
...
+    
+    if ($set_project['id'] <= 0 || empty($set_project['name'])) {
+      throw new \Exception('The Project Trait requires a project that exists in the database.');
+    }
+
+    $this->context[ $this->trait_key ] = $set_project; 
+  }
+
+  /**
+   * Gets the project.
```
laceysanderson:
```suggestion
   * Returns a single project which has been verified to exist by the setter for use
   * by a validator.
```

laceysanderson (commented):
More suggestions. I've now looked through all the traits except for Headers which was not finished at the time I reviewed. Furthermore, the test coverage on all but the header trait is 100% and looks good!

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,201 @@
...
+    'optional', // Header may contain a value.
+  ];
+
+  /**
+   * Sets the headers.
+   * 
+   * @param $headers
+   *   The key-value pair array of headers defined by the importer.
+   *   Each item consists of a header name, text description and type (required, optional)
+   *   keyed by name, description and type, respectively.
```
laceysanderson: No mention of description ;-p

```suggestion
   *   A list of header definitions defined by the importer (usually in a protected
   *   variable at the top of the importer class). Each item in this list is an associative
   *   array defining the specific header and the headers must be listed in order!
   *   Each header item must consist of a header name (key: 'name'), and type (key: 'type',
   *   supported values: 'required', 'optional').
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,201 @@
...
+   *   A list of header types to get. 
+   *   Default to required and optional header types.
+   * 
+   * @return array
+   *   A list or combination of lists of headers of type
+   *   defined by the types parameter.
+   *   
+   *   Returned array is keyed by the index (order) from
+   *   the headers array and header name as the value.
+   *   NOTE: the headers array is zero-based index.
```
laceysanderson: This was not updated after the design change to merge the types into a single array.

```suggestion
   * @return array
   *   A single list of headers matching a type defined by the types parameter.
   *   For example, if the types parameter includes required and optional the the 
   *   resulting array will container all headers in order of index assuming there
   *   are only required and optional types supported.
   *
   *   The returned array is keyed by the index (column order) from
   *   the headers array and header name as the value.
   *   NOTE: the headers array is zero-based index.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Project.php
reynoldtan: id - project id 👍

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,201 @@
...
+   *  - If an unrecognized header type is requested in the types parameter.
+   */
+  public function getHeaders(array $types = ['required', 'optional']) {
+
+    $valid_types = $this->types;
+    
+    // See if the requested types to determine if there are any that are not recognized. 
+    // If any such unrecognized types are detected, throw an exception.
+    $invalid_types = array_filter($types, function($type) use($valid_types) { 
+      return !in_array($type, $valid_types);
```
carolyncaron, line 188: The comment above says it will throw an exception here, but instead it returns a boolean. I think it should be an exception message that perhaps mentions which types are unrecognized?

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,201 @@
...
+        }
+
+        // Type value is one of valid types.
+        if ($key == 'type' && !in_array($header[ $key ], $this->types)) {
+          $str_types = implode(', ', $this->types);
+          throw new \Exception('Headers Trait requires the header key: ' . $key . ' value to be one of [' . $str_types . '].');
+        }
+      }
+
+      // With the header type already verified to be one of the valid types,
```
carolyncaron, line 95: I think we want to avoid gradually building the context array here. Instead, assign to a temporary array and then set the $context array at the very end so that we don't end up with a partially set context array if an exception gets thrown partway through

carolyncaron (commented):
Here's a quick review of the Headers trait for @reynoldtan . Let me know if you have any questions!

trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitHeadersTest.php
```diff
@@ -0,0 +1,516 @@
...
+   *     - required: headers array returned by the getRequiredHeaders() getter method (required type headers).
+   *     - optional: headers array returned by the getOptionalHeaders() getter method (optional type headers).
+   */
+  public function provideHeadersForHeadersSetter() {
+    return [
+      [
+        'headers array is empty',
+        [
+          []
+        ],
```
laceysanderson: This case is not testing what you think it is. According to line 58 in your Headers Trait, and exception should be thrown for empty headers with the message "The Headers Trait requires an array of headers and must not be empty" but that's not the message you are checking for here.

Instead your input_headers should be the following to actually test this case.
```suggestion
        [
        ],
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+  public function setHeaders(array $headers) {
+
+    // Headers must have a value.
+    if (empty($headers)) {
+      throw new \Exception('The Headers Trait requires an array of headers and must not be empty.');  
+    }
+    
+    // Required keys that each header element must posses and 
+    // could not be set to an empty value. Type key value must be
+    // one of the type values defined by the types property.
```
carolyncaron:
```suggestion
    // Required keys that each header element must possess and 
    // cannot be set to an empty value. The 'type' key's value must be
    // one of the 'type' values defined by the types property. 
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+    // Required keys that each header element must posses and 
+    // could not be set to an empty value. Type key value must be
+    // one of the type values defined by the types property. 
+    $required_header_keys = [
+      'name', // Name of the header.
+      'type'  // Type of the header (ie. required or optional).
+    ];
+    
+    // Each header, check required keys and value and place 
+    // in designated type array of the context.
```
carolyncaron:
```suggestion
    // For each header, check that the required keys exist and contain a value.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+  
+  /**
+   * Get required headers.
+   * 
+   * @return array
+   *   All headers of type required, keyed by the index (order) from
+   *   the headers array and header name as the value.
+   * 
+   *   The key required in the context headers array set by the setter method.
+   *   NOTE: the headers array is zero-based index.
```
carolyncaron: I removed `The key required in the context headers array set by the setter method.` because the "key" mentioned here I assume refers to the context or type key, but this is not obvious when the previous paragraph talks about the array being keyed by index. I don't think we need to mention the context key at all since we are supposed to be talking about the return value here, so I suggest just truncating the two paragraphs into one:

```suggestion
   *   All headers of type 'required', keyed by the index (order) from
   *   the headers array and header name as the value.
   *   NOTE: the array is zero-based indexed.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * 
+   * @throws \Exception
+   *  - If the 'headers' key does not exists in the context array
+   *    (ie. the headers element has NOT been set).
```
carolyncaron: Not only the 'headers' key in the context array, but also the 'required' type key. I also like to mention here which setter method needs to be called in order for the array to be available to get.
```suggestion
   *  - If the ['headers']['required'] key does not exist in the context array
   *    (ie. the setHeaders() method has NOT been called).
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+  
+  /**
+   * Get optional headers.
+   * 
+   * @return array
+   *   All headers of type optional, keyed by the index (order) from
+   *   the headers array and header name as the value.
+   * 
+   *   The key optional in the context headers array set by the setter method.
+   *   NOTE: the headers array is zero-based index.
```
carolyncaron:
```suggestion
   *   All headers of type 'optional', keyed by the index (order) from
   *   the headers array and header name as the value.
   *   NOTE: the array is zero-based indexed.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * 
+   * @throws \Exception
+   *  - If the 'headers' key does not exists in the context array
+   *    (ie. the headers element has NOT been set).
```
carolyncaron:
```suggestion
   *  - If the ['headers']['optional'] key does not exist in the context array
   *    (ie. the setHeaders() method has NOT been called).
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+    if (array_key_exists($this->context_key, $this->context)) {
+      return $this->context[ $this->context_key ][ $type_key ];
+    }
+    else {
+      throw new \Exception('Cannot retrieve ' . $type_key . ' headers from the context array as one has not been set by setHeaders() method.');
+    }
+  }
+
+  /**
+   * Get headers of type(s).
```
carolyncaron:
```suggestion
   * Get headers of a specified type(s).
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * 
+   * @return array
+   *   A single list of headers matching a type defined by the types parameter.
+   *   For example, if the types parameter includes required and optional the the 
+   *   resulting array will container all headers in order of index assuming there
+   *   are only required and optional types supported.
+   *
+   *   The returned array is keyed by the index (column order) from
+   *   the headers array and header name as the value.
+   *   NOTE: the headers array is zero-based index.
```
carolyncaron:
```suggestion
   *   A single list of headers matching the type(s) defined by the 'types' parameter.
   *   For example, if the 'types' parameter includes 'required' and 'optional', then the 
   *   resulting array will contain all headers in order of index (this is assuming there
   *   are only 'required' and 'optional' types supported).
   *
   *   The returned array is keyed by the index (column order) from the original headers 
   *   array that was given to the setHeaders() method. 
   *   The value is the header 'name'.
   *   NOTE: the array is zero-based indexed.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * 
+   * @throws \Exception
+   *  - If the 'headers' key does not exists in the context array
+   *    (ie. the headers element has NOT been set).
+   *  - If an unrecognized header type is requested in the types parameter.
```
carolyncaron:
```suggestion
   * If the ['headers'][<one of the types given as a param>] key does not exist in the context array.
   * This could be caused by either:
   *  - The setHeaders() method has NOT been called.
   *  - If an unrecognized header type is requested in the 'types' parameter.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   */
+  public function getHeaders(array $types = ['required', 'optional']) {
+
+    $valid_types = $this->types;
+    
+    // Pull out any unrecognized types by compairing the types paramter
+    // to the types property listing valid types.
```
carolyncaron:
```suggestion
    // This use of array_filter will pull out any unrecognized types by 
    // comparing the 'types' parameter to the types property defined
    // at the top of this class.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * @throws \Exception
+   *  - If the 'headers' key does not exists in the context array
+   *    (ie. the headers element has NOT been set).
+   */
+  public function getRequiredHeaders() {
+    
+    $type_key = 'required';
+
+    if (array_key_exists($this->context_key, $this->context)) {
+      return $this->context[ $this->context_key ][ $type_key ];
```
carolyncaron: What would happen if `$this->context['headers']` exists but NOT `$this->context['headers']['required']`? I think there needs to be a separate check for this case.

See `getGenusConfigured()` and its tests for an example of this!

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+   * @throws \Exception
+   *  - If the 'headers' key does not exists in the context array
+   *    (ie. the headers element has NOT been set).
+   */
+  public function getOptionalHeaders() {
+    
+    $type_key = 'optional';
+
+    if (array_key_exists($this->context_key, $this->context)) {
+      return $this->context[ $this->context_key ][ $type_key ];
```
carolyncaron: Again, `$this->context['headers']['optional']` needs to be checked if it exists separately.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+      $str_valid_types = implode(', ', $valid_types);
+      throw new \Exception('Cannot retrieve invalid header types: ' . $str_invalid_types . '. Use one of valid types: [' . $str_valid_types . ']');
+    }
+
+    if (array_key_exists($this->context_key, $this->context)) {
+      // At this point, types requested are valid.   
+      $headers = [];
+
+      foreach($types as $type) {
+        foreach($this->context[ $this->context_key ][ $type ] as $index => $header) {
```
carolyncaron: Similarly to getRequiredHeaders() and getOptionalHeaders(), we want to check for the presence of each 'type' key in the `$this->context['headers']` array

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,213 @@
...
+          $headers[ $index ] = $header;
+        }
+      }
+      
+      // All headers of type matching all types requested into 
+      // one array key by the index and header name as the value.
```
carolyncaron:
```suggestion
      // All of the header elements with a type matching one of the types 
      // requested, keyed by the column index and 'header name' as the value.
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitHeadersTest.php
```diff
@@ -0,0 +1,514 @@
...
+   * @param array $headers_input
+   *   Headers array input value.
+   * @param array $types_input
+   *   Header types array input value.
+   * @param boolean $has_exception
+   *   Indicates if the test scenario will thrown an exception (TRUE) or not (FALSE).
+   * @param array $exception_message
+   *   An array of exception messages that setter and getters will throw.
+   * @param array $expected
+   *   An array of headers array each getter will produce.
```
carolyncaron: Note to myself: I'm realizing I should have done this in the FileType tests 🙈

carolyncaron (changes requested):
Here is my review of the Headers trait. Don't worry - most of these are suggestions made to documentation!

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,217 @@
...
+        foreach($this->context[ $this->context_key ][ $type ] as $index => $header) {
+          $headers[ $index ] = $header;
+        }
+      }
+      
+      // All of the header elements with a type matching one of the types 
+      // requested, keyed by the column index and 'header name' as the value.
+      if ($unset_types == 0) {
+        return $headers;
+      }
```
laceysanderson: The change you've added to check for unset types is not currently testable. Rather I would suggest the following approach. This ensures we don't make assumptions about keys but is still testable right now :-) 

```suggestion
      $processed_types = 0;

      foreach($types as $type) {
        if (array_key_exists($type, $this->context[ $this->context_key ])) {
          $processed_types++;
          foreach($this->context[ $this->context_key ][ $type ] as $index => $header) {
            $headers[ $index ] = $header;
          }
        }
      }
      
      // All of the header elements with a type matching one of the types 
      // requested, keyed by the column index and 'header name' as the value.
      // Only return if all of the types requested were processed (i.e. were
      // in the headers context array).
      if ($processed_types == sizeof($types)) {
        return $headers;
      }
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
reynoldtan: No longer part of the source code.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
reynoldtan: No longer part of the source code

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
reynoldtan: Implemented as described but as pubic static property

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
reynoldtan: Implemented using the setSupportedMimeTypes() setter method.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
reynoldtan: Both getter methods are implemented

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/DataFileDelimiter.php
laceysanderson: Completed

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,220 @@
...
+   *   supported values: 'required', 'optional').
+   * 
+   *   This list of headers is expected to appear in the data file header row.
+   *   
+   *   NOTE: the set headers array is zero-based index and represents the order
+   *         of the header as each item appears in the headers array parameter
+   *         and is unaltered by setters and getters.
+   * 
+   * @throws \Exception
+   *  - An empty array header.
```
carolyncaron, line 50: Quite a few exceptions get thrown for this method 😉 
```suggestion
   *  - An empty array header.
   *  - When 'name' and/or 'type' key is missing
   *  - When 'name' and/or 'type' key has a missing value
   *  - When the value for 'type' is not one of 'required' or 'optional'
```

laceysanderson (commented):
I did a thorough code review of all changes in this PR and feel it's ready to merge! 🎉 I can't approve officially because technically it's my PR but this is effectively me marking it approved ;-p 

Also, --speechless-- we have 100% coverage of these traits and the base classes! 🤩 

Screenshot 2024-09-03 at 2 17 06 PM
Screenshot 2024-09-03 at 2 17 14 PM

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -0,0 +1,168 @@
...
+   *  - If the 'file_mime_type' key does not exist in the context array
+   *    (ie. the setFileMimeType() method has NOT been called).
+   */
+  public function getFileMimeType() {
+
+    $context_key = 'file_mime_type';
+
+    if (array_key_exists($context_key, $this->context)) {
+      return $this->context[$context_key];
+    } else {
```
reynoldtan: last one, If-else block same as the other.
```suggestion
    }
    else {
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -0,0 +1,220 @@
...
+      // All of the header elements with a type matching one of the types 
+      // requested, keyed by the column index and 'header name' as the value.
+      // Only return if all of the types requested were processed (i.e. were
+      // in the headers context array).
+      if ($processed_types == sizeof($types)) {
+        return $headers;
+      }
+    }
+    
+    throw new \Exception('Cannot retrieve headers from the context array as one has not been set by setHeaders() method.');
```
carolyncaron: I worry this message isn't specific enough to the range of problems that could have triggered it. Even though the most likely scenario is that setHeaders() wasn't called. But I don't want to suggest additional exceptions get thrown since they will be extremely tough to test 😆 How about this?
```suggestion
    throw new \Exception('Cannot retrieve requested headers as the ' . $this->context_key . ' key was not set in the context array, or one or more types were not set by setHeaders() method.'); 
```

carolyncaron (dismissed):
Just a couple of nitpicky suggestions after a bit of discussion with @reynoldtan . But I'm going to approve this now so it can get merged as soon as the changes are made!

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
carolyncaron: Thank you for catching this! 👍

carolyncaron (approved):
Tests are fixed, this looks ready to merge! 🎉

Issue #87, opened 2024-06-26
G4.87 Split GENUS validator into new style validators.


laceysanderson:
### Branch

g4.87-genusValidators

### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Dependencies

- #92 
- #93 

### Describe

This issue is meant to upgrade the existing [Genus](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/src/Plugin/Validators/Genus.php) to the new style described in Issue #82. These will be in inputType `metadata` and will be given the form values. The original [Project](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/src/Plugin/Validators/Project.php) is not needed as it is handled by the drupal form validate by making the field required 🤔 

Does not remove the original validator! Does not update trait importer to use this new validator.

### Design

I think we want two validators:
1. Genus exists and is configured with phenotypes.
2. Project exists and is configured to this genus.

I don't think either one requires any config as they should both have everything they need in the form values. Neither should look at the file in any way ;-p

reynoldtan:
psudo-code for the genusExists() plugin:
```
class GenusExists {
  public function validateMetadata($form_values) {
     // Locate/extract the genus field from the form_values. 
     // This assumes that there is field named ‘genus’.
     
     // If the importer is configured to validate for genus and
     // this validator could not find a genus field.
     1. Throw an exception to the developer: you need/forgot a genus field.

     // If found
     
     Genus is a string value ie. Cicer or Lens.
     1. Check that it exits in chado.organism using the organism.genus.
     
     Failed if not found.
     
     2. Check that the genus is configured, that is the genus 
     trait, method, unit and database are set.

     Failed if not configured.
   } 
}
```

reynoldtan:
For the project validator plugin:
**Proposing** the name projectGenusMatch() - this means that I am validating that project exits and is paired to a genus. The name projectExists() might be good/suited for validator to just check if the project exists alone for Importer that collects just project information.

This plugin assumes the existence of fields named genus and project.

```
class projectGenusMatch {
  public function validateMetadata($form_values) {
    // Locate the project field from $form_values.
    // Can be the project id or project name.
    
    // Locate the genus field from $form_values.
    // Genus is a string value ie. Cicer or Lens.
    
    // Not Found
[...]

PR #96, opened 2024-07-03
G4.87 revise genus validator

reynoldtan:
Issue #87

## Motivation
This draft PR is an installment of the genus+project issue. Currently this draft proposes a genus validator plugin and project validator will follow.

**Update 07/09: projectExists Validator**
project exists validator will check project provided exists in chado.projects table.

**Update 07/09: projectGenusMatch Validator**
project genus match validator will check project provided exists in chado.projects table and inspect the genus set to project matches the genus provided.

## What does this PR do?

1. Check that genus provided in the importer exists
2. Check that genus provided is configured.

**Update 07/09: projectExists Validator**
1. Check that a project id or project name exists in chado.projects table.

**Update 07/11: ProjectGenusMatch Validator**
1. Check that a project with genus set match the value provided in the genus field.

### Automated Testing

Test genus input.
1. Test that an exception is thrown when:
- the $form_state was passed as parameter to the validator instead of doing $form_state->getValues()
- No field with the name/key 'genus' detected in the $form_state

2. Test non-existent genus
4. Test an existing genus but is not configured.
5. Test an existing genus and is configured.

**Update 07/09: projectExists Validator**
Test project input.
1. Test that an exception is thrown when:
- the $form_state was passed as parameter to the validator instead of doing $form_state->getValues()
-  A string value.
- No field with the name/key 'project' detected in the $form_state

2. Test non-existent project
4. Test an existing project using project id and project name.

**Update 07/11: projectGenusMatch Validator**
Test project and genus input.
1. Test that an exception is thrown when:
- the $form_state was passed as parameter to the validator instead of doing $form_state->getValues()
-  A string value.
- No field with the name/key 'project' or 'genus' detected in the $form_state

2. Test non-existent project
4. Test an existing project with a genus set matches  the genus provided in the genus field.

### Manual Testing

**Update 07/09: projectExists Validator**
1. Create a project project record.
2. In your D10/T4 instance, access my site/devel/php
```
$validator_id = 'trpcultivate_phenotypes_validator_project_exists';

// PASS
$form_values = ['project' => 'REPLACE WITH PROJECT ID OR PROJECT NAME'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// FAIL: not found.
$form_values = ['project' => 'out_of_this_world_project'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// Will throw an exception - execute separately
$form_values = 'A string value';
$validation_status = $instance->validateMetadata($form_values);

$form_values = ['genus' => 'Lens', 'file' => 2];
$validation_status = $instance->validateMetadata($form_values);

use Drupal\Core\Form\FormState;
$form_state = new FormState();
$form_state->setValues(['project' => 'abc']);
$validation_status = $instance->validateMetadata($form_state);
```

**Update 07/11: projectGenusMatch Validator**
1. Create a project project record.
2. Create and configure a genus.
3. Setup a project-genus property

```
  ->setGenusToProject(REPLACE WITH PROJECT ID, REPLACE WITH CONFIGURED GENUS);
```

4. In your D10/T4 instance, access my site/devel/php
```
$validator_id = 'trpcultivate_phenotypes_validator_project_genus_match';

// PASS
$form_values = ['project' => 'REPLACE WITH PROJECT ID OR PROJECT NAME', 'genus' => 'REPLACE WITH CONFIGURED GENUS'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// FAIL: not found.
$form_values = ['project' => 'out_of_this_world_project', 'genus' => 'REPLACE WITH CONFIGURED GENUS'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// FAIL: not the correct genus.
$form_values = ['project' => 'REPLACE WITH PROJECT ID', 'genus' => 'Not the genus'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// Will throw an exception - execute separately
$form_values = 'A string value';
$validation_status = $instance->validateMetadata($form_values);

// Missing project field.
$form_values = ['genus' => 'Lens', 'file' => 2];
$validation_status = $instance->validateMetadata($form_values);

// Missing genus field.
$form_values = ['project' => 'The project', 'file' => 2];
$validation_status = $instance->validateMetadata($form_values);

use Drupal\Core\Form\FormState;
$form_state = new FormState();
$form_state->setValues(['project' => 'abc']);
$validation_status = $instance->validateMetadata($form_state);
```

//////////////////////////////////////////////

1. Setup 2 genus where 1 is configured whereas the other is not configured.
2. In your D10/T4 instance, access my site/devel/php

```
$validator_id = 'trpcultivate_phenotypes_validator_genus_exists';

// PASS
$form_values = ['genus' => 'REPLACE WITH CONFIGURED GENUS'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);


// ------------------------------


// FAIL: not configured.
$form_values = ['genus' => 'REPLACE WITH NOT CONFIGURED GENUS'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);


// ------------------------------

// FAIL: not found.
$form_values = ['genus' => 'spurious_genus'];
$validation_status = $instance->validateMetadata($form_values);
print_r($validation_status);

// -----------------------------

// Will throw an exception - execute separately
$form_values = ['project' => 1, 'file' => 2];
$validation_status = $instance->validateMetadata($form_values);

use Drupal\Core\Form\FormState;
$form_state = new FormState();
$form_state->setValues(['genus' => 'abc']);
$validation_status = $instance->validateMetadata($form_state);
```


trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,118 @@
...
+    
+    // Genus.
+    $genus = trim($form_values[ $expected_field_key ]);
+
+    // This method has now curated all genus available in the organism table,
+    // both configured and non-configured genus.
+    $genus_config = $this->service_PhenoGenusOntology->getGenusOntologyConfigValues($genus);
+
+    if (!$genus_config) {
+      // The genus provided does not exist.
```
carolyncaron, line 109: A string for $case should be set here for a non-existing genus

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,118 @@
...
+      $valid = FALSE;
+      $failed_items = $genus . ' (Not found)';
+    }
+    else {
+      // The genus provided does exist, test that it was
+      // fully configured, that is, a cv was set for trait configuration.
+
+      if (empty($genus_config['trait']) || $genus_config['trait'] <= 0) {
+        // Not configured genus.
```
carolyncaron, line 123: A string for $case should be set here for an unconfigured genus

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,118 @@
...
+        $valid = FALSE;
+        $failed_items = $genus . ' (Not configured)';
+      }
+    }
+    
+    return [
+      'case' => 'Genus exists and is configured with phenotypes',
```
carolyncaron: This case string can be set as the default near the top of the class, to indicate that the genus is valid :)

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,180 @@
...
+    // Failed item is the failed genus value. Additional failed information is attached
+    // whether the genus failed because it does not exist or not configured.
+
+    // Test passing $form_state values with a genus field but genus does not exits.
+    $genus = 'genus-' . uniqid();
+    $form_state->setValues(['genus' => $genus]);
+    $form_values = $form_state->getValues();  
+    $validation_status = $instance->validateMetadata($form_values);
+    
+    $this->assertEquals('Genus exists and is configured with phenotypes', $validation_status['case'],
```
carolyncaron: Once the changes are made to genusExists(), each of these assertEquals can now check for unique strings that indicate which case occurred

carolyncaron (commented):
Hi Reynold! This is looking great so far, but there is one thing that jumped out to me. I think we want the return value of 'case' to be different depending on if the validator returns valid or not, and also between the different cases that may cause the input to be invalid. I know that your original validator design kept 'title' as the same text throughout a validator and only 'details' would differ, but here I interpret 'case' as being a more succinct version of 'details'. This way, you can write your kernel tests to check for the specific case that should trigger valid to be false.

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+   *     - valid: either TRUE or FALSE depending on if the genus value is valid or not.
+   *     - failedItems: the failed genus value provided. This will be empty if the value was valid.
+   */
+  public function validateMetadata($form_values) {
+    // This genus validator assumes that a field with name/key genus was
+    // implemented in the Importer form.
+    $expected_field_key = 'genus';
+  
+    // An object was given as parameter to this method.
+    if (is_object($form_values)) {
```
laceysanderson: It would be better to check if it's not an array. For example, if someone passed in a string this wouldn't catch it ;-p

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,180 @@
...
+      ->fields([
+        'genus' => $genus,
+        'species' => 'databasica',
+      ])
+    ->execute();
+
+    $this->assertIsNumeric($organism_id, 'We were not able to create an organism for testing (not configured).');
+    $this->test_genus['not-configured'] = $genus;
+    $config = \Drupal::configFactory()->getEditable('trpcultivate_phenotypes.settings')
+      ->set('trpcultivate.phenotypes.ontology.cvdbon.' . $this->test_genus['not-configured'], ['trait' => 0]);
```
laceysanderson: This should not be needed. The checking of whether a genus is configured or not should not make the assumption that this exists as there are edge cases when the genus is added later where it would not be.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,180 @@
...
+    $instance = $this->plugin_manager->createInstance($validator_id);
+    
+    // The validator expects the form values entered into the form elements
+    // through the $form_state.
+    
+    // Create form fields and set value.
+    // This would have been a form that has been submitted.
+    $form_state = new FormState();
+    // A random field.
+    $form_state->setValues(['project' => uniqid()]);
```
laceysanderson: It would be better for these tests not to mimic the form submission process so closely. For example, do not create a form state here and populate it. Instead for valid case create the array directly and for invalid case pass in a standard object and a string.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,180 @@
...
+    $exception_caught  = FALSE;        
+    try {
+      $instance->validateMetadata($form_state);
+    }
+    catch (\Exception $e) {
+      $exception_caught  = TRUE;
+    }
+    
+    $this->assertTrue($exception_caught, 'Failed to catch exception when passing a $form_state to genus metadata validator.');
+    $this->assertStringContainsString('Unexpected object type was passed', $e->getMessage(),
```
laceysanderson: Using $e here is out of scope. Instead you should have an `$exception_caught` and an `$exception_message` that are set in the catch clause.

laceysanderson (commented):
These are the things that jump out to me in a code review :-)

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
reynoldtan: image

I could not see a string or other single value being passed from the way this will implemented as designed (shown above). I could see the possibility of passing the $form_state object that is why there is this catch.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
reynoldtan: not quite. in the real world, upon install the module will create a config for all genus and set all cv's to 0 - the not configured genus. Other genus added later will not have this configuration setting until set in the config page. Since the install routine was skipped in this test, non configured genus does not have that config item (set to 0) and is always treated as genus does not exists. This will a true test for genus where it was not configured yet.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
reynoldtan: Could you expand more. Based on the design, the values are from field element and I just want to make sure that the test captures that extra step of the $form_state getter.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
reynoldtan: Yikes... strange it did not complain. This is out of scope too https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/d35be827579f35939be3f753aa1ab11fe79d5b5b/trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorTraitImporterTest.php#L151

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
laceysanderson: Yes, I can see why you checked it that way :-) But we shouldn't make assumptions like that in an API as it may be used in other places that you did not expect. We should also always check using the pattern: if this is NOT what we need then throw an error. That way we don't make any assumptions on how it will be used and will catch all invalid input.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
laceysanderson: Exactly, "Other genus added later will not have this configuration setting until set in the config page." So if a genus is added and then this validator is triggered without anyone going to a config page then the module will not have this key set. Again, I just really like tests to make as few assumptions as possible and this feels like an assumption to me 😬 Also, I think it's unnecessary as I believe your validator will catch in even without you setting this in the test. If you want to be really thorough then I suggest adding another case where there is a genus without this set that exists but is not configured. That way you check my case and yours and we will know if the validator works even if your assumption is not true in some weird edge case.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
laceysanderson: Yes, I hear you but that also makes the form state being used first as a potential dependency. Maybe we need to re-word the validator doc block but I meant the note about the form state to be used as hint to the developer but not a requirement. For example, in the future, Tripal Importers may have a drush interface to allow them to be run. In which case, we would want these validators to be useful with a generic array passed in. That is why I made the parameter an array and not just the form state itself --to remove the form state as a requirement.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
laceysanderson: Yes, I found this myself, PHP is very lax on scope so it will let you use it. However, if this test ever fails and an exception isn't thrown then you don't get the nice assert but rather get a variable is not set error 🤪  I know from experience 🙈

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+   *     - failedItems: the failed genus value provided. This will be empty if the value was valid.
+   */
+  public function validateMetadata($form_values) {
+    // This genus validator assumes that a field with name/key genus was
+    // implemented in the Importer form.
+    $expected_field_key = 'genus';
+  
+    // Parameter passed to the method is not an array.
+    if (!is_array($form_values)) {
+      throw new \Exception(t('Unexpected @type type was passed as parameter to genusExists validator.', ['@type' => gettype($form_values)]));
```
laceysanderson: Remember not to use translation in the exception messages as they are meant for the developer. This needs to be fixed in a few other places in this PR as well.
```suggestion
      $type = gettype($form_values);
      throw new \Exception("Unexpected '$type' type was passed as parameter to genusExists validator when we expected an array.");
```

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+    
+    // Failed to locate the genus field element.
+    if (is_array($form_values) && !array_key_exists($expected_field_key, $form_values)) {
```
laceysanderson: No need to check if is array since we already threw an exception if it wasn't.
```suggestion
    if (!array_key_exists($expected_field_key, $form_values)) {
```

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+      throw new \Exception(t('Failed to locate genus field element. genusExists validator expects a form field element name genus.'));
+    }
+  
+    // Validator response values for a valid genus value.
+    $case = 'Genus exists and is configured with phenotypes';
+    $valid = TRUE;
+    $failed_items = '';
```
laceysanderson: This should be an array according to the interface.
```suggestion
    $failed_items = [];
```

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+
+    // This method has now curated all genus available in the organism table,
+    // both configured and non-configured genus.
+    $genus_config = $this->service_PhenoGenusOntology->getGenusOntologyConfigValues($genus);
+
+    if (!$genus_config) {
+      // The genus provided does not exist.
+      $case = 'Genus does not exist';
+      $valid = FALSE;
+      $failed_items = $genus;
```
laceysanderson: Should be an array. Maybe key it by what it is for easier use later?
```suggestion
      $failed_items = ['genus_provided' => $genus];
```

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+    }
+    else {
+      // The genus provided does exist, test that it was
+      // fully configured, that is, a cv was set for trait configuration.
+
+      if (empty($genus_config['trait']) || $genus_config['trait'] <= 0) {
+        // Not configured genus.
+        $case = 'Genus exists but is not configured';
+        $valid = FALSE;
+        $failed_items = $genus;
```
laceysanderson:
```suggestion
        $failed_items = ['genus_provided' => $genus];
```

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
```diff
@@ -0,0 +1,121 @@
...
+      $valid = FALSE;
+      $failed_items = $genus;
+    }
+    else {
+      // The genus provided does exist, test that it was
+      // fully configured, that is, a cv was set for trait configuration.
+
+      if (empty($genus_config['trait']) || $genus_config['trait'] <= 0) {
+        // Not configured genus.
+        $case = 'Genus exists but is not configured';
```
laceysanderson, line 124: CodeClimate says this case is not being covered by automated testing 🤔

trpcultivate_phenotypes/src/Plugin/Validators/projectExists.php
```diff
@@ -0,0 +1,107 @@
...
+   *   Each form element value can be accessed using the field element key
+   *   ie. field name/key project - $form_values['project'].
+   * 
+   *   This array is the result of calling $form_state->getValues().
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *     - case: a developer focused string describing the case checked.
+   *     - valid: either TRUE or FALSE depending on if the project value existed or not.
+   *     - failedItems: the failed project value provided. This will be empty if the value was valid.
```
laceysanderson: Again, this should be an array

trpcultivate_phenotypes/src/Plugin/Validators/projectGenusMatch.php
```diff
@@ -0,0 +1,141 @@
...
+      $failed_items = $project;
+    }
+    else {
+      // Inspect the set genus to the project and see if it matches
+      // the genus provided in the genus field.
+      $project_genus = $this->service_PhenoGenusProject->getGenusOfProject($project_id);
+      
+      if ($genus != $project_genus['genus']) {
+        // Genus does not match the genus paired to the project.
+        $case = 'Genus does not match the genus set to the project';
```
laceysanderson: Codeclimate says this case is not being caught by automated testing.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,441 @@
...
+
+    // Set test environment.
+    \Drupal::state()->set('is_a_test_environment', TRUE);
+
+    // Install module configuration.
+    $this->installConfig(['trpcultivate_phenotypes']);
+
+    // Test Chado database.
+    // Create a test chado instance and then set it in the container for use by our service.
+    $this->connection = $this->createTestSchema(ChadoTestKernelBase::PREPARE_TEST_CHADO);
```
laceysanderson: We should always use `$chado_connection` now as of PR #102.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,441 @@
...
+  * Tests Tripal Cultivate Phenotypes Metadata Validator Plugins.
+  *
+  * @group trpcultivate_phenotypes
+  * @group validators
+  */
+class MetadataInputTest extends ChadoTestKernelBase {
+  use PhenotypeImporterTestTrait;
+
+  /**
+   * A genus that exists and is configured.
```
laceysanderson: This implies it is a single value but looking below it is an array. Make sure to type hint for all variables both in the doc block and before the variable. In the case of an array, also describe the keys expected in the doc block.

```
  /**
   * A Database query interface for querying Chado using Tripal DBX.
   *
   * @var ChadoConnection
   */
  protected ChadoConnection $chado_connection;
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,441 @@
...
+    ->execute();
+
+    $this->assertIsNumeric($organism_id, 'We were not able to create an organism for testing (not configured).');
+    $this->test_genus['not-configured'] = $genus;
+    
+    // Set terms configuration.
+    $this->setTermConfig();
+
+
+    // Create test project.
```
laceysanderson: Would be good to create 3 projects so that you cover the additional cases where 
1. the project exists but is not attached to any genus
2. the project exists but is attached to a different genus

It might also be good to test the case where the project has a genus property that was added without the `trpcultivate_phenotypes.genus_project` service.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,441 @@
...
+
+    // Genus does not exist.
+    $genus = 'genus-' . uniqid();
+    $form_values = ['genus' => $genus];  
+    $validation_status = $instance->validateMetadata($form_values);
+    
+    $this->assertEquals('Genus does not exist', $validation_status['case'],
+      'Genus exists validator case title does not match expected title for non-existent genus.');
+    $this->assertFalse($validation_status['valid'], 'A failed genus must return a FALSE valid status.');
+    $this->assertStringContainsString($genus, $validation_status['failedItems'], 'Failed genus value is expected in failed items.');
```
laceysanderson: This should be an assertEquals rather then string contains string. Also this will change because failed items should always be an array.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -0,0 +1,441 @@
...
+    
+
+    // Genus exists but not configured/recognized by the module.
+    $genus = $this->test_genus['not-configured'];
+    $form_values = ['genus' => $genus];
+    $validation_status = $instance->validateMetadata($form_values);
+    
+    $this->assertEquals('Genus does not exist', $validation_status['case'],
```
laceysanderson: This should be returning `Genus exists but is not configured` according to the code above which is why this test is passing but the actual case where a genus exists and is not configured is not being covered according to codeclimate.

I'm guessing this is due to an assumption/bug in the phenotypes module that getGenusOntologyConfigValues will only be called after the ontology config form has been loaded once and the genus detected by the system?

In my opinion, this should not be fixed by changing the setUp for the test but instead by changing the validator and/or getGenusOntologyConfigValues() method.

laceysanderson (changes requested):
Some suggestions based on code review. I did not do a manual test. These are getting there 😍

carolyncaron (changes requested):
I followed the instructions for manual testing and most tests returned the expected result, with one exception. I already brought this up with @reynoldtan in person but am documenting it here as well. When passing an existing but non configured genus to the genusExists() validator, I received a 'case' of 'Genus does not exist'. See screenshot below:
Screenshot 2024-07-31 at 2 49 08 PM
I would have expected to see the message 'Genus exists but is not configured'. @laceysanderson already pointed out in her review that the automated test was also not picking up this case.

carolyncaron (commented):
I confirmed in a manual test that the genusExists() validator now detects when a genus exists but is not configured 🎉  I will follow up with a code review tomorrow 😄 

Screenshot 2024-08-07 at 4 43 15 PM

trpcultivate_phenotypes/src/Plugin/Validators/genusExists.php
carolyncaron: CodeClimate now shows coverage 👍

trpcultivate_phenotypes/src/Plugin/Validators/projectGenusMatch.php
carolyncaron: CodeClimate now shows coverage 👍

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
carolyncaron: @laceysanderson's comment for context:

> Would be good to create 3 projects so that you cover the additional cases where
> 1. the project exists but is not attached to any genus
> 2. the project exists but is attached to a different genus
>
> It might also be good to test the case where the project has a genus property that was added without the trpcultivate_phenotypes.genus_project service.

I see the automated tests for suggested case 1 & 2 now, and confirmed that manual testing of these also get caught. However, they both return the same message: `Genus does not match the genus set to the project` 

@reynoldtan is it possible to distinguish between these two cases with a different message, or does that require too much logic do you think?

Lastly, I didn't see any automated test for Lacey's last suggestion:
> where the project has a genus property that was added without the trpcultivate_phenotypes.genus_project service.

I also don't know how to go about manually testing this 😅

carolyncaron (dismissed):
Code review looks really good to me - great job! 😄 Please see my reply to one of Lacey's now outdated comments.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
laceysanderson: > Lastly, I didn't see any automated test for Lacey's last suggestion:
> 
> > where the project has a genus property that was added without the trpcultivate_phenotypes.genus_project service.
> 

This was added by setting up the project-genus-not-tru-service project by inserting the genus directly to the projectprop table instead of using the service in the setup() ✅

laceysanderson (approved):
All the changes I suggested have no been addressed -ready to merge! 🎉 

Note: I did add the type hinting that I originally suggested since it hadn't been completed fully. See changes here: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/96/files/695d82dc76724daa87cc1002edef25cb44d72cce..384af230233381ee223710e50d8558211729a099

Issue #88, opened 2024-06-26
G4.88 Split FILE validator into new style validators.


laceysanderson:
### Branch

g4.88-fileValidators

### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Dependencies

- #92 
- #93 

### Describe

This issue is meant to upgrade the existing [DataFile](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/src/Plugin/Validators/DataFile.php) to the new style described in Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/82.

Does not remove the original validator! Does not update trait importer to use this new validator.

### Design

This should be split into two validators.

The first one is a generic check the file is valid without looking it in. This will be in inputType file and be given the filename with full path and the fid (if uploaded). It will check the following:

- Has Drupal File Id assigned/created.
- A valid file extension based on what's configured.
- Matches mime type configured.
- File Id created/assigned can be loaded.
- Is not empty file.

The other one is specific to tsv format and has an inputType of header and will be given the first row of the file as an array. It will implement validateRow and check that the contents of the first row are in TSV format.

reynoldtan:
```
/**
 * input_types: {"file"} 
 */

class validateFile {
  public function validateFileObject($filename, $fid = NULL) {
    // Parameter check:
    Check $filename is a valid path and is accessible (when provided).
    Throw an exception if invalid file path.

    Check $fid is 0 or not integer value (when provided).
    Throw an exception if non-integer value.


    // Has File ID and Id can be loaded check
    // Load $filename (load by uri) or $fid (load by file id)
    Failed if it cannot load the file. File has no file id assigned/created.
    
    // File exists check.
    // Reference file uri and check if it exists in the filesystem.
    // ie: file_exists($file->uri);
    Failed if the file does not exists.
   
    // MIME and file extension check.
    // Get the file mime type provided by the Drupal File System object.
    // ie: $file->getMimeType();
    Failed if type/extension is not the expected mime type configured.
    
    // File is not empty check.
    // Get the file size provided by the Drupal File System object. 
    // ie: filesize($file->uri);
    Failed if file size is 0
  }
}

```

reynoldtan:
```
/**
 * input_types: {"header"} 
 */

class validTsvFile {
  public function validateRow($header_row)
[...]

PR #98, opened 2024-07-24
G4.88 file validators

reynoldtan:
**Issue #88 ** Split File Validator

## Motivation
This PR will update the DataFile validator as describe in the issue.
The update will implement 2 validators, 1 validator is responsible for validating the file for compliance in terms of file type, mime, not empty file and with a File id, whereas the other validator is to check the a raw row in the data file is delimited.

## What does this PR do?
*Please describe each thing this PR does. For example, a PR may 1) solve a specific bug, 2) create an automated test to ensure it doesn't return.*

1. Create a file-level validation check plugin 
2. Create a raw row delimiter validator plugin.
3. Create a setter method to set a validator configuration for:
  A. number of columns: the number of columns to expect
  B. strict: a strict comparison flag.
4. Create a getter method to get the values set in no. 3.

### Automated Testing

**Class: DataFileInputTest - testDataFileInput()**

Automated test for file-level check:
1. Test an empty filename value for the $filename parameter.
2. Test a 0 file id value for the $fid parameter.
3. Test a non-existent file.
4. Test file not managed by Drupal file system.
6. Test by providing an empty file.
7. Test a file with the wrong MIME type.
8. Test a file of an invalid type masquerading as a valid type.
9. Test a valid file of type primary file type (ie. tsv).
10. Test a valid file of type alternative file type (ie txt).

Tests 3-9 contain tests for both parameters filename (uri) and file id number.

**Class: DataFileDelimiterTest - testDataFileRowIsDelimited()**
Automated test for the raw row is delimited:

1. Test with an empty raw row.
2. Test none of the set delimiters was used in the raw row.
3. Test that the line did not return the expected number of columns with the strict flag set to true.
4. Test that the line did not return the expected number of columns with the strict flag set to false.
5. Test raw row has one delimiter to delimit values and another delimiter within a value.
6. Test a valid raw row that will return the expected number of columns with the strict flag set to true.
7. Test a valid raw row that will return the expected number of columns with the strict flag set to false.
8. Test a raw row that has only one column and no delimiter.

**Class DataFileDelimiterTest - testGetExpectedColumns()**

1. This will test the validator-getter method before calling the setter method. In this test, the setter method will throw an error.

**Class DataFileDelimiterTest - testValidatorSetterAndGetter()**

1. Test the number of columns set to a zero value.
2. Test a valid number of columns.
3. Test the getter will return the set values.

### Manual Testing


trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,199 @@
...
+
+use Drupal\trpcultivate_phenotypes\TripalCultivateValidator\TripalCultivatePhenotypesValidatorBase;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Validate a line in a data file is properly delimited.
+ *
+ * @TripalCultivatePhenotypesValidator(
+ *   id = "trpcultivate_phenotypes_validator_valid_delimited_file",
```
carolyncaron: I think we are moving away from these long ids for validators since many of them will be used across multiple modules in the future. Here is my suggestion to keep consistent with the validators I've been working on:
```suggestion
 *   id = "valid_delimited_file",
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,199 @@
...
+ *   validator_name = @Translation("Valid Delimiter File Validator"),
```
carolyncaron:
```suggestion
 *   validator_name = @Translation("Valid Delimited File Validator"),
```

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
+use Drupal\file\Entity\File;
+
+/**
+ * Validate data file.
+ *
+ * @TripalCultivatePhenotypesValidator(
+ *   id = "trpcultivate_phenotypes_validator_valid_data_file",
```
carolyncaron:
```suggestion
 *   id = "valid_data_file",
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/DataFileInputTest.php
```diff
@@ -0,0 +1,376 @@
...
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Set test environment.
+    \Drupal::state()->set('is_a_test_environment', TRUE);
+
+    // Create a plugin instance for this validator
+    $validator_id = 'trpcultivate_phenotypes_validator_valid_data_file';
```
carolyncaron:
```suggestion
    $validator_id = 'valid_data_file';
```

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+ *   validator_name = @Translation("Valid Data File Validator"),
+ *   input_types = {"file"}
+ * )
+ */
+class ValidDataFile extends TripalCultivatePhenotypesValidatorBase implements ContainerFactoryPluginInterface {
```
carolyncaron: I noticed that the filename for this class starts as lowercase instead of uppercase. I don't know how to suggest a change like that through Github though so I'm mentioning it here 😆

carolyncaron (commented):
Just some things I've noticed for now! I think I am similar to Lacey in this case that it'll be easier to review fully once both validators are finished 👍

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
reynoldtan: Git did not allow to push just a filename change but did allow with a bit of code changes plus the filename change.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,211 @@
...
+   * 
+   * @throws \Exception
+   *  - The value 0 as number of columns.
+   */
+  public function setExpectedColumns($number_of_columns, $strict = FALSE) {
+
+    $context_key = $this->validator_context_key;
+
+    if ($number_of_columns <= 0) {
+      throw new \Exception('The setter method in ' . $context_key . ' requires an integer value greater than zero.');
```
carolyncaron: I think it's better to explicitly state which method since there is potential for future additional setter methods in this class.
```suggestion
      throw new \Exception('setExpectedColumns() requires an integer value greater than zero.');
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,211 @@
...
+   */
+  public function getExpectedColumns() {
+
+    $context_key = $this->validator_context_key;
+
+    if (array_key_exists($context_key, $this->context)) {
+      return $this->context[ $context_key ];
+    }
+    else {
+      throw new \Exception('Cannot retrieve the values set by the ' . $context_key . ' setter method.');
```
carolyncaron:
```suggestion
      throw new \Exception('Cannot retrieve the number of expected columns as one hasn't been set by setExpectedColumns().');
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,211 @@
...
+   * Set a number of required columns.
+   * 
+   * @param integer $number_of_columns
+   *   An integer value greater than zero.  
+   * @param bool $strict
+   *   This will indicate how to compare the the number of columns set against
+   *   the number of columns returned by the split line method.
+   * 
+   *   A true value represents an exact match, while a false indicates 
+   *   a minimum number of columns required.
```
carolyncaron:
```suggestion
   *   This will indicate whether the value $number_of_columns is the minimum
   *   number of columns required in an input file's row, or if it is strictly the only
   *   acceptable number of columns.
   *   - FALSE (default) = minimum number of columns
   *   - TRUE = the strict number of required columns
```

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+/**
+ * Validate data file.
+ *
+ * @TripalCultivatePhenotypesValidator(
+ *   id = "valid_data_file",
+ *   validator_name = @Translation("Valid Data File Validator"),
+ *   input_types = {"file"}
+ * )
+ */
+class ValidDataFile extends TripalCultivatePhenotypesValidatorBase implements ContainerFactoryPluginInterface {
```
carolyncaron: Unfortunately the filename is still showing as `validDataFile.php` for me

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+   * Perform validation of data file.
+   * Checks include:
+   *  - Parameter filename or file id is valid.
+   *  - Has Drupal File Id number assigned and can be loaded.
+   *  - Valid file extension and file mime type configured by the file importer instance.
+   *  - File exists and not empty.
+   *  - File can be opened.
+   * 
+   * @param string $filename
+   *   The location of a file within the file system.
```
carolyncaron: I don't know if this needs to be an absolute or relative path, but it would be helpful to mention that here

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+    if (!is_null($fid) && $fid <= 0) {
+      return [
+        'case' => 'Invalid file id number',
+        'valid' => FALSE,
+        'failedItems' => ['file_id' => $fid]
+      ];
+    } 
+
+    // File input.
+    $file_input = (is_null($fid)) ? $filename : $fid;
```
carolyncaron: Is only one parameter (filename or fid) required? That should be documented in the docblock above. Also, what would happen if both filename and fid are provided? It looks like fid is prioritized, but it's the 2nd parameter... is this what we want?

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+      
+        // Check that both the file extension and file MIME type
+        // are supported by the importer.
+        $file_mime_type = $file_object->getMimeType();
+        $file_filename  = $file_object->getFileName();
+        $file_extension = pathinfo($file_filename, PATHINFO_EXTENSION);
+
+        $supported_file_extensions = $this->getSupportedFileExtensions();
+        $supported_mime_types = $this->getSupportedMimeTypes();
+        if (in_array($file_mime_type, $supported_mime_types) && in_array($file_extension, $supported_file_extensions)) {
```
carolyncaron: I think these should be separate checks to do more helpful error reporting. However, I wonder if file extension should be checked for at all here. If anything, it can be used in the case string. For example:
`$case = 'The file type is not supported by this importer. Make sure your file is one of the following: $supported_file_extensions'`

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+        $file_object = File::load($file_id);
+      }
+
+      $input_key = 'filename';
+    }
+    
+
+    // Validate data file:
+
+    if ($file_object) {
```
carolyncaron: I would inverse all of these if statements so that you can set the validation array inside the if block instead of in the else. Currently, it sadly makes it very hard to follow :(

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
```diff
@@ -0,0 +1,197 @@
...
+        $file_extension = pathinfo($file_filename, PATHINFO_EXTENSION);
+
+        $supported_file_extensions = $this->getSupportedFileExtensions();
+        $supported_mime_types = $this->getSupportedMimeTypes();
+        if (in_array($file_mime_type, $supported_mime_types) && in_array($file_extension, $supported_file_extensions)) {
+          
+          // Check that the file can be opened.
+          $file_uri  = $file_object->getFileUri();
+          $file_handle = @fopen($file_uri, 'r');
+          if (!$file_handle) {
```
carolyncaron: Code coverage says this if block is missing an automated test

trpcultivate_phenotypes/tests/src/Kernel/Validators/DataFileDelimitedTest.php
```diff
@@ -0,0 +1,343 @@
...
+use Drupal\Tests\tripal_chado\Kernel\ChadoTestKernelBase;
+use Drupal\Tests\trpcultivate_phenotypes\Traits\PhenotypeImporterTestTrait;
+
+ /**
+  * Tests Tripal Cultivate Phenotypes Data File Delimited Validator Plugins.
+  *
+  * @group trpcultivate_phenotypes
+  * @group validators
+  */
+class DataFileDelimitedTest extends ChadoTestKernelBase {
```
carolyncaron: Lacey and I had a discussion about test names and we've decided to change the standard on you again 🙈 Sorry!! I've added a new README to the tests/src/Kernel/Validators/ folder in my latest PR to help exlpain this.

This class should be `ValidatorValidDelimitedFileTest`. Remember that the filename will need updating as well!

trpcultivate_phenotypes/tests/src/Kernel/Validators/DataFileDelimitedTest.php
```diff
@@ -0,0 +1,343 @@
...
+   *   - Number of expected colum input (first parameter to the setter method).
+   *   - Strict condition flag input (second parameter to the setter method).
+   *   - Expected values set:
+   *     - number_of_columns: the number of columns to expect.
+   *     - strict: strict comparison flag.
+   *   - Expected exception message for both setter and getter:
+   *    - setter: setter exception message.
+   *    - getter: getter exception message.
+   */
+  public function provideValidatorConfigurationValues() {
```
carolyncaron: I would name this more specifically to what it is used for. Something like `provideExpectedColumnsForSetter()`

trpcultivate_phenotypes/tests/src/Kernel/Validators/DataFileDelimitedTest.php
```diff
@@ -0,0 +1,343 @@
...
+    }
+    catch (\Exception $e) {
+      $exception_caught = TRUE;
+      $exception_message = $e->getMessage();
+    }
+
+    $this->assertEquals(
+      $exception_message,
+      $exception['setter'],
+      'Exception message does not match the expected message when trying to set a zero value to number of columns in scenario ' . $scenario
```
carolyncaron: This is testing more than a zero number of columns, so the message should be more generic. How about:
```suggestion
      'Exception message does not match the expected message when trying to call ValidDelimitedFile::setExpectedColumns() in scenario ' . $scenario
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/DataFileDelimitedTest.php
```diff
@@ -0,0 +1,343 @@
...
+    }
+    catch (\Exception $e) {
+      $exception_caught = TRUE;
+      $exception_message = $e->getMessage();
+    }
+
+    $this->assertEquals(
+      $exception_message,
+      $exception['getter'],
+      'Exception message does not match the expected message when trying to get expected number of columns in scenario ' . $scenario
```
carolyncaron:
```suggestion
      'Exception message does not match the expected message when trying to call ValidDelimitedFile::getExpectedColumns() in scenario ' . $scenario
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,211 @@
...
+   * @param string $raw_row
+   *   A line in the data file that can be the headers row (line no. 1) or a data row.
+   * 
+   * @return array
+   *   An associative array with the following keys.
+   *     - case: a developer focused string describing the case checked.
+   *     - valid: either TRUE or FALSE depending on if the raw row is delimited or not.
+   *     - failedItems: the failed raw row. This will be an empty array if the line was delimited.
+   */
+  public function validateRawRow($raw_row) {
```
carolyncaron: Code Climate is telling me this about this method:
Screenshot 2024-09-16 at 5 51 54 PM

I wanted to provide more helpful advice to address this but I think we should consider a meeting between you, me and Lacey 😅

carolyncaron (changes requested):

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: Yes, I am still seeing the lowercase as well and agree with changing it to uppercase in the file name. To get this change to take in git you have to use git mv. For example, `git mv trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php`. Using git to do this ensures that the commit history stays attached to the file. You will likely see something in the git status that says it has been renamed and this should be the only change in this commit.

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: Related to this, I agree we want to load the file based on fid and only based on filename if there is no fid.

However, I don't think this $file_input variable is helpful at all and I think they way we handle having both is to check that they match if the file entity is loaded via fid.

Instead I would do something like this to replace your lines 110-136:
```php
$file_entity = NULL;
if (is_numeric($fid) && $fid > 0) {
  // load the file entity based on the fid.
  // if not successful then failed case based on unloadable fid
  // else (i.e. loaded file entity) 
  //    check that the file entity->filename matches the filename passed into us.
  //    if they don't match then failure.
}
elseif ($filename) {
  // try to load file entity based on the filename.
  // if not successful then failed case based on filename not associated with file object.
}
// at this point $file_entity should be loaded or a failed case returned.
```

Note: you also don't need to know this key for saving to failedItems later. Instead you will have the file entity object and you should save both the filename and the fid to failedItems.

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: I think it would be even more readable to just keep each check in it's own if block and return when there is a failure.

Something like:
```php
// Check that the file is not blank by inspecting the file size
// to see if it is greater than 0.
$file_size = $file_object->getSize();
if (!is_numeric($file_size) OR ($file_size <= 0)) {
  //return the failure.
}

// Check that both the file extension and file MIME type
// are supported by the importer.
$file_mime_type = $file_object->getMimeType();
$file_filename  = $file_object->getFileName();
$file_extension = pathinfo($file_filename, PATHINFO_EXTENSION);
```

Note that there is no nesting of if statements this way because it is not needed. We will only get to the second check if the first one passes.

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: I agree with separating these. Also remember these are developer messages.

Check 1: is the mime type in the supported mime types?
  if not then...
    check file extension is one of supported?
      if not then failed case 'unsupported file mime/extension', failedItems includes both the mime-type and file extension of file entity.
      if yes then failed case 'unsupported file mime and mismatched extension', failedItems includes both the mime-type and file extension of file entity.
  if yes... all is well, we don't care about file extension.

laceysanderson (commented):
I only made it through the ValidDataFile validator before I ran out of time... will come back to this on Monday.

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: This is fixed now.

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: My concerns here have been resolved ✅

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: My concerns here are resolved ✅

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
laceysanderson: My concerns for this file are resolved ✅ but it would be good to see the same approach in  ValidDataFile and your other validators in general to improve readability.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
laceysanderson: Update on this. I think the changes made to the other validator could solve this one too.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,212 @@
...
+      ];
+    }
+
+    // Check if the line has some delimiters used (only if number of expected columns is greater than 1).
+    $expected_columns = $this->getExpectedColumns();
+    if ($expected_columns['number_of_columns'] == 1) {
+      return [
+        'case' => 'Data file raw row is delimited',
+        'valid' => TRUE,
+        'failedItems' => []
```
carolyncaron: I think we might be missing additional checks before returning these values since the only thing we know about $raw_row so far is that its not an empty line. Should we also check for presence of delimiters even when we expect only one column? Ultimately, how will splitRowIntoColumns() behave in this scenario?

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,212 @@
...
+        // Not a strict comparison - at least x number of columns.
+        if (count($columns) < $expected_columns['number_of_columns']) {
+          array_push($delimiters_checked, $delimiter);
+        }
+      }
+    }
+
+    // If all delimiters failed the checks, then the line failed due to none of the
+    // delimiter was able to split the line into the expected number of columns.
+    if ($delimiters_used == $delimiters_checked) {
```
carolyncaron: This isn't obvious to me why this if statement means the file isn't delimited with an expected delimiter 🙈 . Could you expand the comment above this to explain the logic a bit more? Perhaps we could also rename $delimiters_checked since it implies that any delimiters that pass the checks didn't actually get checked? 😅

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,212 @@
...
+      return [  
+        'case' => 'Raw row is not delimited',
+        'valid' => FALSE,
+        'failedItems' => ['raw_row' => $raw_row]
+      ];
+    }
+
+    return [
+      'case' => 'Data file raw row is delimited',
```
carolyncaron: To make this consistent with line 145, this should be
```suggestion
      'case' => 'Raw row is delimited',
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,212 @@
...
+   *   This will indicate whether the value $number_of_columns is the minimum
+   *   number of columns required in an input file's row, or if it is strictly the only
+   *   acceptable number of columns.
+   *   - FALSE (default) = minimum number of columns
+   *   - TRUE = the strict number of required columns
+   * 
+   * @return void
+   * 
+   * @throws \Exception
+   *  - The value 0 as number of columns.
```
carolyncaron:
```suggestion
   *  - If $number_of_columns is not greater than 0.
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+   * @param integer $fid
+   *   The unique identifier (fid) of a file that is managed by Drupal File System.    
+   * 
+   * @return array
+   *   An associative array with the following keys.
+   *    - case: a developer focused string describing the case checked.
+   *    - valid: either TRUE or FALSE depending on if the file is valid or not.
+   *    - failedItems: an array of "items" that failed to be used in the message to the user. This is an empty array if the file input was valid.
+   */
+  public function validateFile($filename, $fid = NULL) {
```
carolyncaron:
```suggestion
  public function validateFile(string $filename, int $fid = NULL) {
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,212 @@
...
+   * @param string $raw_row
+   *   A line in the data file that can be the headers row (line no. 1) or a data row.
+   * 
+   * @return array
+   *   An associative array with the following keys.
+   *     - case: a developer focused string describing the case checked.
+   *     - valid: either TRUE or FALSE depending on if the raw row is delimited or not.
+   *     - failedItems: the failed raw row. This will be an empty array if the line was delimited.
+   */
+  public function validateRawRow($raw_row) {
```
carolyncaron:
```suggestion
  public function validateRawRow(string $raw_row) {
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+   *  - Parameter filename or file id is valid.
+   *  - Has Drupal File Id number assigned and can be loaded.
+   *  - Valid file extension and file mime type configured by the file importer instance.
+   *  - File exists and not empty.
+   *  - File can be opened.
+   * 
+   * @param string $filename
+   *   The full path to a file within the file system (Absolute file path).
+   * @param integer $fid
+   *   The unique identifier (fid) of a file that is managed by Drupal File System.
```
carolyncaron:
```suggestion
   *   [OPTIONAL] The unique identifier (fid) of a file that is managed by the Drupal File System.    
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+        return [
+          'case' => 'Unsupported file MIME type',
+          'valid' => FALSE,
+          'failedItems' => ['mime' => $file_mime_type, 'extension' => $file_extension]
+        ];
+      }
+      else {
+        // Both MIME type and file extension are not supported.
+        return [
+          'case' => 'Unsupported file mime type and mismatched extension',
```
carolyncaron: Its possible the file extension might make sense with the unsupported mime-type, so not technically "mismatched"
```suggestion
          'case' => 'Unsupported file mime type and unsupported extension',
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+      }
+    }
+
+    // Check that the file can be opened.
+    $file_uri  = $file_object->getFileUri();
+    $file_handle = @fopen($file_uri, 'r');
+    
+    if (!$file_handle) {
+      return [
+        'case' => 'The file cannot be opened',
```
carolyncaron: To be consistent with line 193:
```suggestion
        'case' => 'Data file cannot be opened',
```

trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php
carolyncaron: Unfortunately I still have concerns here... But Lacey and I discussed it a bit and think it's worth sitting down to re-think our use of parameters for this method. I created a new issue for this here: #110 so that it doesn't hold back merging in this PR

carolyncaron (changes requested):
Just a few more suggestions! One of them was large enough and requires a lot of thought/discussion, please see the comment I marked as unresolved for context.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+   * @param string $filename
+   *   The full path to a file within the file system (Absolute file path).
+   * @param integer $fid
+   *   [OPTIONAL] The unique identifier (fid) of a file that is managed by Drupal File System.
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *    - case: a developer focused string describing the case checked.
+   *    - valid: either TRUE or FALSE depending on if the file is valid or not.
+   *    - failedItems: an array of "items" that failed to be used in the message to the user. This is an empty array if the file input was valid.
```
laceysanderson:
```suggestion
   *    - failedItems: an associative array indicating the filename and fid for an invalid file. This is an empty array if the file input was valid.
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+      $file_entities = $this->service_EntityTypeManager
+        ->getStorage('file')
+        ->loadByProperties(['uri' => $filename]);
+
+      $file_entity = reset($file_entities);
+
+      if ($file_entity) {
+        $fid = $file_entity->get('fid')->value;
+        $file_object = File::load($fid);
+      }
```
laceysanderson: I don't believe this is necessary as shouldn't the file entity === file object? These are just two different ways to load the same thing, in my understanding.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // Check that the file input provided returned a file object.
+    if (is_null($file_object)) {
+      return [
+        'case' => 'Filename or file id failed to load a file object',
```
laceysanderson: It would be good to know if the filename or the fid were used here :thinking. I think my pseudocode showed this orientation but in practice I now think there should be a check for the file object right after it's loaded above (so two checks: one after loaded via fid and the other after loaded via filename).

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+    $file_filename = $file_object->getFileName();
+    $file_fid = $file_object->id();
+
+    // Check that the file is not blank by inspecting the file size to see if it is greater than 0.
+    $file_size = $file_object->getSize();
+    if (!$file_size) {
+      return [
+        'case' => 'The file has no data and is an empty file',
+        'valid' => FALSE,
+        'failedItems' => $failed_items = ['filename' => $file_filename, 'fid' => $file_fid]
```
laceysanderson:
```suggestion
        'failedItems' => ['filename' => $file_filename, 'fid' => $file_fid]
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,198 @@
...
+
+    // Check that the file input provided returned a file object.
+    if (is_null($file_object)) {
+      return [
+        'case' => 'Filename or file id failed to load a file object',
+        'valid' => FALSE,
+        'failedItems' => ['filename' => $filename, 'fid' => $fid]
+      ];
+    }
+
```
laceysanderson, line 147: We need another case here that checks if the filename provided as a paramter matches the one in the file object loaded via the fid.

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,218 @@
...
+   *    set by the validator setter method.
+   *
+   * @param string $raw_row
+   *   A line in the data file that can be the headers row (line no. 1) or a data row.
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *     - case: a developer focused string describing the case checked.
+   *     - valid: either TRUE or FALSE depending on if the raw row is delimited or not.
+   *     - failedItems: the failed raw row. This will be an empty array if the line was delimited.
```
laceysanderson:
```suggestion
   *     - failedItems: an associative array indicating the the failed raw row under the 'raw_row' key. This will be an empty array if the line was delimited.
```

laceysanderson (changes requested):

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
reynoldtan: Loading a file via fid means that the filename is not provided. Please clarify (include the case title and failed items to return).

laceysanderson: I think that's the way we think of it but technically **you can provide both parameters to this function**. If you do provide both a filename and a fid, the fid is preferentially used since it is checked first in the logic. 

So the scenario is:
if 
 - you are in the first block (i.e loading via fid because it was supplied)
 - you can load the file object via the fid
 - the filename parameter is also non-empty

then
 - check the filename parameter and the file_object->filename match

@carolyncaron mentions this should actually throw an exception since it should only happen if the developer sets the parameters wrong and I think I agree.

reynoldtan: What will be the case title and failed items to return?

laceysanderson: If you throw an exception then you don't need a case/failedItems since you do not return anything. We're in gather if you want to chat about it further 💞

reynoldtan: What will be the exception message so I can test it?

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,218 @@
...
+   *    set by the validator setter method.
+   *
+   * @param string $raw_row
+   *   A line in the data file that can be the headers row (line no. 1) or a data row.
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *     - case: a developer focused string describing the case checked.
+   *     - valid: either TRUE or FALSE depending on if the raw row is delimited or not.
+   *     - failedItems: an associative array indicating the the failed raw row under the 'raw_row' key. This will be an empty array if the line was delimited.
```
carolyncaron:
```suggestion
   *     - failedItems: an associative array indicating the failed raw row under the 'raw_row' key. This will be an empty array if the line was delimited.
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,218 @@
...
+    foreach($input_file_type_delimiters as $delimiter) {
+      if (strpos($raw_row, $delimiter)) {
+        array_push($delimiters_used, $delimiter);
+      }
+    }
+
+    // Not one of the supported delimiters was detected in the raw row.
+    if (empty($delimiters_used)) {
+
+      // Return a valid response if the raw raw is not delimited value
```
carolyncaron:
```suggestion
      // Return a valid response if the raw row is not delimited value
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,218 @@
...
+  }
+
+  /**
+   * Set a number of required columns.
+   *
+   * @param integer $number_of_columns
+   *   An integer value greater than zero.
+   * @param bool $strict
+   *   This will indicate whether the value $number_of_columns is the minimum
+   *   number of columns required in an input file's row, or if it is strictly the only.
```
carolyncaron:
```suggestion
   *   number of columns required in an input file's row, or if it is strictly the only
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -0,0 +1,207 @@
...
+   *   The full path to a file within the file system (Absolute file path).
+   * @param integer $fid
+   *   [OPTIONAL] The unique identifier (fid) of a file that is managed by Drupal File System.
+   *
+   * @return array
+   *   An associative array with the following keys.
+   *    - case: a developer focused string describing the case checked.
+   *    - valid: either TRUE or FALSE depending on if the file is valid or not.
+   *    - failedItems: an associative array indicating the filename and fid for an invalid file. This is an empty array if the file input was valid.
+   */
```
carolyncaron, line 85:
```suggestion
   *
   * @throws \Exception
   *   - If both parameters $filename and $fid are provided, but do not point to
   *     the same file object
   */
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -0,0 +1,218 @@
...
+    // Reference the expected number of columns.
+    $expected_columns = $this->getExpectedColumns();
+
+    // Based on the file mime type of the input file, reference the delimiter
+    // defined specific to the type.
+    $input_file_mime_type = $this->getFileMimeType();
+    $input_file_type_delimiters = $this->getFileDelimiters($input_file_mime_type);
+
+    // Check if the line has some delimiters used, specifically, check the line
+    // includes at least one of the delimiters returned by the get file delimiter method.
```
carolyncaron:
```suggestion
    // includes at least one of the delimiters returned by getFileDelimiters().
```

carolyncaron (dismissed):
These are just some very minor suggestions to comments - no code changes needed! :tada: 

Great work on these pesky validators!!

laceysanderson (approved):
You resolved all my suggestions and a last pass through looks great! 😍 This is ready to merge!!! 🎉

Issue #89, opened 2024-06-26
G4.89 Updates header validator into new style validator


laceysanderson:
### Branch

g4.89-headerValidator

### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Dependencies

- #92 
- #93 

### Describe

This issue is meant to upgrade the existing [Headers](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/src/Plugin/Validators/Headers.php) to the new style described in Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/82.

Does not remove the original validator! Does not update trait importer to use this new validator.

### Design

This will be in inputType header-row and will implement validateRow which will be called only once for an importer with the first row passed to it as an array of column values. It checks that the header line is not empty and that it has the values it expects. The expected values are configured in order to make this validator generic for any importer.

PR #99, opened 2024-07-25
G4.89 header validator

reynoldtan:
**Issue #89** Update header validator

## Motivation

This PR updates header validator as described in the issue.
The update will implement a validator that will check that the header row (row 1) of the data file is not an empty line/string and that values (headers) match the expected headers defined by the importer instance.

## What does this PR do?

1. Create header validator

### Automated Testing

ValidHeadersTest - testHeaders()
1. Test an empty headers array input.
2. Test with missing headers.
3. Test some headers in the wrong order.
4. Test a valid header.

### Manual Testing

```
// Create instance of the validator:
$validator_id = 'valid_headers';
$validator_instance = \Drupal::service('plugin.manager.trpcultivate_validator')
  ->createInstance($validator_id);


// Set the headers expected by the importer:
$validator_instance->setHeaders([
  [
    'name' => 'Header 0',
    'type' => 'required'  // #0
  ],
  [
    'name' => 'Header 1',
    'type' => 'required'  // #1
  ],
  [
    'name' => 'Header 2',
    'type' => 'required'  // #2
  ],
  [
    'name' => 'Header 3',
    'type' => 'optional'  // #3
  ],
  [
    'name' => 'Header 4',
    'type' => 'optional'  // #4
  ],
  [
    'name' => 'Header 5',
    'type' => 'required'  // #5
  ]
]);

// Create an headers input array:
$headers_input = [
  0 => 'Header 0',
  1 => 'Header 1',
  2 => 'Header 2',
  3 => 'Header 3',
  4 => 'Header 4',
  5 => 'Header 5',
];

// Validation will fail if you pass an empty headers array, if any required 
// headers are missing, or if the headers are not in the correct order.

// Validate headers.
$validation_status = $validator_instance->validateRow($headers_input);
dpm($validation_status);
```


carolyncaron (changes requested):
I found a few curious things while testing:
1. When I tested for BOTH missing headers and headers being out of order, it only returns a case string that a header is missing. I think it would be ideal to get told about both. This might also require the `failedItems` element to be a nested array with keys like `missing` and `wrong_order`, just as examples.
Screenshot 2024-10-01 at 12 23 55 PM

>

2. When I play around with providing non-integer keys, I always get a valid of `True` regardless of order (as long as none of the headers are missing). I think we have to decide whether we should enforce an array with only integer keys or if we should explicitly care about the order of the values, regardless of what the keys look like. I think I lean towards the latter option.
Screenshot 2024-10-01 at 12 35 45 PM

reynoldtan:
Headers missing and not in order - could you point me to a validator that returns multiple cases that I can use as a guide?

carolyncaron:
> Headers missing and not in order - could you point me to a validator that returns multiple cases that I can use as a guide?

Yes! In the DuplicateTraits validator, there's a case for if a trait combo is in both the database and the file:  https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/66326bda2979350f4a755c1c5d0de8264160294f/trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php#L198

reynoldtan:
I think this one is just one case, the title indicates two places but the failed items contain just one element. Could you describe the logic and include the case title and failed items? what happens if there's just missing header? what is the failed items in this case?

carolyncaron:
> I think this one is just one case, the title indicates two places but the failed items contain just one element.

In that validator, the failedItems array makes sense because the same thing is triggering 2 different problems (ie. it was the one trait combo from one row in the input file that was triggering a duplicate in two separate places). Line 211 covers when a duplicate is only in the file: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/66326bda2979350f4a755c1c5d0de8264160294f/trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php#L211 
and line 225 covers when it's only duplicated in the database:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/66326bda2979350f4a755c1c5d0de8264160294f/trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php#L225

In your case, 2 separate things in the input header row are causing validation to fail:
1. One or more headers is missing
2. One or more headers is not in the expected order

And then, a 3rd scenario would be if both of the above are happening.

I don't know what the best logic is to cover all 3 scenarios, as I think you will likely come up with something better! But I'd be happy to chat through it with you privately if it helps to clarify what I've written here 😊

reynoldtan:
how's this?

1. parameter check
2. inspect for missing headers and store the result in the missing headers array.
3. inspect for headers in the wrong order and store the results in not in order headers array. 
4. Validate 
Are both missing and not-in-order arrays have elements? 
```
return [
  'case' => 'Missing expected headers and headers not in the correct order',
  'valid' => FALSE,
  'failedItems' => ['missing' => missing array, 'wrong_order' => not in order array]
]
```

Only missing array has elements in it
```
return [
   'case' => 'Missing expected headers',
   'valid' => FALSE,
   'failedItems' => ['missing' => missing array]
]
```

Only not in order array has element in it
```
return [
  'case' => 'Headers not in the correct order',
  'valid' => FALSE,
  'failedItems' => ['wrong_order' => not in order array]
]
```

Please advice if the array should be converted to a string comma-separated values.

carolyncaron:
Looks great! The only thing I'd change is the 'headers' key - for only missing headers you can use just 'missing', and for ones in the wrong order I would use 'wrong_order' only.

I like the idea of keeping the arrays as is instead of converting to a string. Maybe @laceysanderson can confirm but the conversion to a comma separated can happen later when we process the return values for user feedback

laceysanderson:
> > Please advice if the array should be converted to a string comma-separated values.
>
> I like the idea of keeping the arrays as is instead of converting to a string. Maybe @laceysanderson can confirm but the conversion to a comma separated can happen later when we process the return values for user feedback.

We definitely do not want to convert it to a string in failedItems :-) That's for the user feedback to do later if it wants. Keeping everything as arrays where possible is a general rule of thumb for failedItems as it allows for better automated test checks and more flexibility for user friendly messages later.

reynoldtan:
Revised header validator modified the following:
1. Moved the setter and getter for expected number of columns and strict comparison flag from ValidDelimitedFile validator into its own validator trait called ColumnCount
2. Made adjustments to the Headers trait to ensure that the order of headers as they appear is maintained by every getter methods.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -92,132 +92,112 @@ public function setHeaders(array $headers) {
...
-      // With the header type already verified to be one of the valid types, 
-      // push the header into the designated type temporary context array.
-      $context_headers[ $header['type'] ][ $index ] = $header['name'];
-    }
-    
-    // Set each header type context array.
-    foreach($this->types as $type) {  
-      $this->context[ $this->context_key ][ $type ] = $context_headers[ $type ] ?? [];
+      // With the header type already verified to be one of the valid types,
+      // create a header entry where it contains only the name and type keys.
```
laceysanderson, line 96: Expanding the comment to make it clear why there is this extra step b/c I hesitated on it for longer then I should have.

```suggestion
      // create a header entry where it contains only the name and type keys.
      // This is needed because there is nothing stopping the input $headers
      // array from containing additional keys.
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -92,132 +92,112 @@ public function setHeaders(array $headers) {
...
+
+    foreach($this->context[ $this->context_key ] as $index => $header) {
+      if (in_array($header['type'], $types)) {
+        $requested_headers[ $index ] = $header['name'];
       }
     }
-    
-    throw new \Exception('Cannot retrieve requested headers as the ' . $this->context_key . ' key was not set in the context array, or one or more types were not set by setHeaders() method.'); 
+
+    return ($requested_headers) ? $requested_headers : [];
```
laceysanderson: This check is not needed since you set the variable to an empty array on line 193.

```suggestion
    return $requested_headers;
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorValidHeadersTest.php
```diff
@@ -0,0 +1,377 @@
...
+   * An instance of the data file validator.
+   *
+   * @var object
+   */
+  protected $validator_instance;
+
+  /**
+   *
+   */
+
```
laceysanderson: Not sure why this is here so removing it.
```suggestion

```

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorValidHeadersTest.php
```diff
@@ -0,0 +1,377 @@
...
+      ],
+    ];
+  }
+
+  /**
+   * Test headers validator.
+   *
+   * @dataProvider provideHeadersToHeadersValidator
+   */
+  public function testHeaders($scenario, $headers_input, $expected, $expected_columns) {
```
laceysanderson, line 390: It would be good to check that the failedItems in expected match the headers input just as a check that you setup your scenarios properly.

And maybe also check the order matches ;-p

laceysanderson (commented):
- new ColumnCount trait change looks perfect
- regarding the changes in the HeadersTrait: you should update you check that the headers returned matches expected to ensure they are in the same order since that slipped through last time.

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/Headers.php
```diff
@@ -92,132 +92,114 @@ public function setHeaders(array $headers) {
...
-      // requested, keyed by the column index and 'header name' as the value.
-      // Only return if all of the types requested were processed (i.e. were
-      // in the headers context array).
-      if ($processed_types == sizeof($types)) {
-        return $headers;
+
+    // Prepare fullset/subset of the importer headers where each header type matches
+    // the header type requested. The final list of headers generated will maintain
+    // the index and order of the header as they appear in headers
+    // set by in the setHeaders method.
```
carolyncaron:
```suggestion
    // the index and order of the original array provided to setHeaders().
```

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorValidHeadersTest.php
```diff
@@ -0,0 +1,416 @@
...
+            'Header 5',
+          ]
+        ],
+        [
+          'number_of_columns' => 6,
+          'strict' => FALSE
+        ]
+      ],
+
+      // #5: All headers not in order and validator will fail in the first ecounter missing/wrong order.
```
carolyncaron:
```suggestion
      // #5: All headers not in order and validator will fail in the first encounter of missing/wrong order.
```

carolyncaron (dismissed):
Just a couple of re-wording suggestions. Manual testing passed with flying colours!! 🌈 😁

carolyncaron (approved):
Approved! 💯 🎉

PR #101, opened 2024-07-29
G4.100 split row into columns

reynoldtan:
## Motivation

Create a method that will split or explode a line from the data file into columns using a delimiter chosen based on the mime type of the file.

This PR will 
1. update the importer form validate to pass a string value (the raw data row/line) to validateRawRow() in the importer
2. provide a static splitRowIntoColumns() method so the importer can split the row based on the mime type of the file.
3. ensure all automated testing checks this new approach.

### Automated Testing

ValidatorBaseTest - testSplitRowIntoColumns()
1. Test case where no delimiter
2. The method return the same line (delimiter could not split the row)
3. Test that the line was split and that each value was sanitized
4. Test basic row input
5. Do validation 1-4 for all supported mime types.

### Manual Testing

In my site/devel/php of a working T4 clone, paste the code block and execute. The line array and/or the delimiter character can be altered to other value to test other set of values.

```
// LINE:
$line = ['Value A', 'Value "B"', 'Value \'C\'', 'Value D ', ' Value E', ' Value F ', ' Value G           '];

// MIME TYPE:
$mime_type = 'text/tab-separated-values';
$delimiter = "\t";

$str_line = implode($delimiter, $line);
$split_row = \Drupal\trpcultivate_phenotypes\TripalCultivateValidator\TripalCultivatePhenotypesValidatorBase::splitRowIntoColumns($str_line, $mime_type);

print_r($split_row);
```

The method includes sanitizing value from quotes (double and single) as well as leading and trailing spaces.

The code should create the array below where the extracted values are sanitized and in an array.

```

Array
(
    [0] => Value A
    [1] => Value B
    [2] => Value C
    [3] => Value D
    [4] => Value E
    [5] => Value F
    [6] => Value G
)

```

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -508,8 +508,9 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         // ********************************************************************
         // Skip empty lines
         else if (!empty(trim($line))) {
-          // Split line into an array
-          $data_row = str_getcsv($line, "\t");
+          // Split line into an array using the delimter defined by this importer
+          // in the configure values method above.
+          $data_row = $this->splitRowIntoColumns($line);
 
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -508,9 +508,8 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         // ********************************************************************
         // Skip empty lines
         else if (!empty(trim($line))) {
-          // Pass the line (string) and row validator handles separating values
-          // into an array based on the delimiter defined by this importer instance.
-          $data_row = $this->splitRowIntoColumns($line);
+          // Split line into an array
+          $data_row = str_getcsv($line, "\t");
 
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -510,7 +510,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         else if (!empty(trim($line))) {
           // Pass the line (string) and row validator handles separating values
           // into an array based on the delimiter defined by this importer instance.
-          $data_row = $line;
+          $data_row = $this->splitRowIntoColumns($line);
 
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -12,7 +12,7 @@ use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\file\Entity\File;
 use Drupal\Core\Url;
-
+use Drupal\trpcultivate_phenotypes\TripalCultivateValidator\TripalCultivatePhenotypesValidatorBase;
 use Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTraitsService;
 
 /**
@@ -202,6 +202,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
     
     // This importer expects data in tab-separated values.
     // Delimiter: tabulator key - \t
+    // @TODO: use the delimiter setter.
     $context['delimiter'] = "\t";
 
     $instance->context = $context;
@@ -312,7 +313,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#title' => 'Genus',
       '#description' => t('The genus of the germplasm being phenotyped with the supplied traits.
         Traits in this system are specific to the genus in order to ensure they are specific enough to accurately describe the phenotypes.
-        In order for genus to be availabe here is must be first configured in the Analyzed Phenotypes configuration.'),
+        In order for genus to be available here is must be first configured in the Analyzed Phenotypes configuration.'),
       '#empty_option' => '- Select -',
       '#options' => $active_genus,
       '#default_value' => $default_genus,
@@ -348,7 +349,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
 
     $file_id = $form_values['file_upload'];
 
-    // Use a variable to keep track of if one input type had recieved errors
+    // Use a variable to keep track of if one input type had received errors
     // and only continue to the next if no errors.
     $failed_validator = FALSE;
 
@@ -508,9 +509,9 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         // ********************************************************************
         // Skip empty lines
         else if (!empty(trim($line))) {
-          // Split line into an array using the delimter defined by this importer
+          // Split line into an array using the delimiter defined by this importer
           // in the configure values method above.
-          $data_row = str_getcsv($line, "\t");
+          $data_row = TripalCultivatePhenotypesValidatorBase::splitRowIntoColumns($line);
 
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -199,6 +199,10 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       $header_index['Unit'],
       $header_index['Type']
     ];
+
+    // Delimiter for tab-separated values: \t
+    $context['delimiter'] = "\t";
+
     $instance->context = $context;
     $validators['data-row']['empty_cell'] = $instance;
 
@@ -503,8 +507,8 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         // ********************************************************************
         // Skip empty lines
         else if (!empty(trim($line))) {
-          // Split line into an array
-          $data_row = str_getcsv($line, "\t");
+          $data_row = $line;
+
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
             $result = $validator->validateRow($data_row);
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -480,8 +480,11 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         // Header Row Validation
         // ********************************************************************
         if ($line_no == 1) {
+          // Split line into an array using the delimiter defined by this importer
+          // in the configure values method above.
+          $header_row = TripalCultivatePhenotypesValidatorBase::splitRowIntoColumns($line, $file_mime_type);
           foreach ($validators['header-row'] as $key => $validator) {
-            // @TODO: Update to use the validateRow() method
+            // @TODO: Update to use the validateRow() method and use the split $header_row above.
             $result = $validator->validate();
             $validation[$key] = $result;
             // Check for old style...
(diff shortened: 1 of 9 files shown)


trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -201,4 +212,39 @@ public function getConfigAllowNew() {
...
+   * 
+   * @return array
+   *   An array containing the values extracted from the line after splitting it based
+   *   on a delimiter value.
+   */
+  public function splitRowIntoColumns(string $row) {
+    // Delimiter:
+    $delimiter = $this->context['delimiter'] ?? ''; 
+    if (empty($delimiter)) {
+      throw new \Exception(t('No delimiter provided.'));
```
laceysanderson: We don't translate exception messages as they are not meant for the UI but instead for the developer.

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -201,4 +212,39 @@ public function getConfigAllowNew() {
...
+    }
+    
+    // Split the values.
+    $values = explode($delimiter, $row);
+
+    if (count($values) == 1 && $values[0] === $row) {
+      // The delimiter failed to split the row and returned the original row.
+      throw new \Exception(t('The data row or line provided could not be split using the delimiter.'));
```
laceysanderson: Again no translation. 
Also you should include what delimiter this method used that failed as it can be helpful when debugging.

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorBaseTest.php
```diff
@@ -312,4 +325,93 @@ public function testValidatorValidateMethods() {
...
+    ];
+    $instance = new BasicallyBase($configuration, $validator_id, $plugin_definition);
+    $this->assertIsObject(
+      $instance,
+      "Unable to create fake_basically_base validator instance to test the base class."
+    );
+
+    // Create a data row.
+    // This line captures data values with single/double quotes and leading/trailing spaces.
+    $line = $line_copy = ['Value A', 'Value "B"', 'Value \'C\'', 'Value D ', ' Value E', ' Value F ', ' Value G           '];
```
laceysanderson: This test would be easier to follow if you called this `$good_line` and `$messy_line` or something similar; `$line_copy` is not very informative.

laceysanderson (commented):
Did a code review and overall looks good to me :-) I didn't do a functional test though. Some comments below.

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
reynoldtan: Delimiter now part of the exception message 👍

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorBaseTest.php
reynoldtan: Went with good line and raw line 👍

laceysanderson (dismissed):
Code-wise this looks good to me :-)
Now this PR is just waiting on a manual test + input by Carolyn.

carolyncaron (changes requested):
Notes and tasklist from our in person meeting on July 30: 
- I brought up concerns regarding calling the splitRowIntoColumns() within every file-row input type validator for every row of a file
- We collectively decided that the following tasks are still needed to merge this PR to try and prevent this method from causing slowdowns in the future with large files:
- [x] Revert existing validators; validateRow() still takes an array 
- [x] SplitRow() method becomes a static method and it’s tests stay 
- [x] FormValidate now calls splitRow() once per row and provide split row to validateRow() 
- [x] Add validateRawRow() to interface/base which takes a string to be used to validate the unsplit row. This will have a new raw-row input type

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -307,7 +313,7 @@ public function form($form, &$form_state) {
       '#title' => 'Genus',
       '#description' => t('The genus of the germplasm being phenotyped with the supplied traits.
         Traits in this system are specific to the genus in order to ensure they are specific enough to accurately describe the phenotypes.
-        In order for genus to be availabe here is must be first configured in the Analyzed Phenotypes configuration.'),
+        In order for genus to be available here is must be first configured in the Analyzed Phenotypes configuration.'),
```
carolyncaron:
```suggestion
        In order for genus to be available here, it must be first configured in the Analyzed Phenotypes configuration.'),
```

carolyncaron (commented):
Started my code review - still in progress!

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -201,4 +220,41 @@ public function getConfigAllowNew() {
...
+   *   A line in the data file.
+   * 
+   * @return array
+   *   An array containing the values extracted from the line after splitting it based
+   *   on a delimiter value.
+   */
+  public static function splitRowIntoColumns(string $row) {
+    // Delimiter:
+
+    // @TODO: use the delimiter getter.
```
laceysanderson: We cannot use a context variable in a static function. Instead we will have to pass the delimeter in as a variable. I suggest making the delimeter an array to allow us to support multiple types of files :thinking:

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -201,4 +220,41 @@ public function getConfigAllowNew() {
...
+    $delimiter = "\t";
+    if (empty($delimiter)) {
+      throw new \Exception('No delimiter provided.');
+    }
+    
+    // Split the values.
+    $values = explode($delimiter, $row);
```
laceysanderson: I think we're actually better to use str_getcsv() here. The delimiter can be passed into str_getcsv() as the separator. str_getcsv() differs from explode() in that it supports files with quotes around columns and escape charaters to allow quotes/delimiter in the columns.

laceysanderson (dismissed):
Looking good except for the changes suggested in this review. Note: I did not do a manual test nor did I check the tests/test coverage as the changes suggested will change them quite a bit and allow the commented out ones to be completed. Back to you @reynoldtan :-)

laceysanderson:
This is ready to review again @carolyncaron 😍 I also updated the PR description to match the current state including the testing information.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -503,8 +505,10 @@ public function formValidate($form, &$form_state) {
         // ********************************************************************
         // Skip empty lines
         else if (!empty(trim($line))) {
-          // Split line into an array
-          $data_row = str_getcsv($line, "\t");
+          // Split line into an array using the delimiter defined by this importer
+          // in the configure values method above.
+          $data_row = TripalCultivatePhenotypesValidatorBase::splitRowIntoColumns($line, $file_mime_type);
```
carolyncaron, line 513: I think this method is going to need to be called on the header row above as well. Currently only validate() is being called there, so we'll need to remember to include the splitter when it gets updated to validateRow()

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -6,6 +6,18 @@
...
+   *   on the validator. For example, row level validators are passed the raw
+   *   row from the file and thus need the importer to indicate how it should
+   *   be split.
+   *
+   *   Row level validators require the following key(s):
+   *   - delimiter => the delimiter to be passed to explode in order to break
+   *     the raw row string into an array of columns.
+   */
+  public array $context = [];
+
```
carolyncaron: The context array isn't needed for this PR. It will get added in PR #95 instead :)
```suggestion
```

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -201,4 +221,88 @@ public function getConfigAllowNew() {
...
+      }
+
+      // Now lets choose the one with the most columns --shrugs-- not ideal
+      // but I'm not sure there is a better option. asort() is from smallest
+      // to largest preserving the keys so we want to choose the last element.
+      asort($counts);
+      $winning_delimiter = array_key_last($counts);
+      $columns = $results[ $winning_delimiter ];
+      $delimiter = $winning_delimiter;
+    }
```
carolyncaron: @laceysanderson and I had an in person discussion on this already, but this logic is sadly feeling error-prone given how much flexibility we want to allow in the file. For example, we've seen the infamous "comments" column (might have a header or not 🤷 ) multiple times and a sentence could include multiple commas which could overpower the overall number of tabs in a tab-delimited plain text file. 

A few points were also brought up:
- Move this logic to another method and maybe only run it once on the header row?
- We may also need a separate validator to compare to validate that each row contains the minimum number of columns expected
- Handling this case makes more sense to do in a separate PR, and the urgency can be lower than other tasks in this module since we can train our own curator to avoid plain-text formats for now

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -121,6 +121,35 @@ public function validateFile(string $filename, int $fid);
...
+   * Validates rows within the data file submitted to an importer.
+   *
+   * Note: This should only be used when validating the format of the row.
+   * If you are validating the content of the columns then you should use
+   * validateRow() instead.
+   *
+   * @param string $raw_row
+   *  A single line or row extracted from the data file
+   *  containing data entries, values or column headers, with each value
+   *  delimited by a character specified by the importer class.
```
carolyncaron, line 140: Sorry to keep bringing up potential gotchas 🙈 But I wonder if we want to consider allowing empty row and not require it to contain "data entries values or column headers". I'm especially thinking ahead to genotypes where there may be lots of commented rows as well - and we might want to actually do something with them since the format is standardized (in the case of a VCF)! Just a thought.

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
```diff
@@ -121,6 +121,35 @@ public function validateFile(string $filename, int $fid);
...
+   *      (i.e. no record in chado matching project name). If the data is
+   *      is valid then this is not required but could be 'data verified'.
+   *  - 'valid': a boolean indicating the data is valid (TRUE) or not (FALSE)
+   *  - 'failedIems': an array of the items that failed validation. For example,
+   *      if this validator validates that a number of indicies are not emptya then
+   *      this will be an array of indices that were empty. Another example is
+   *      that if this validator checks that a number of indices have values in
+   *      a specific list, then this array would use the index as the key and
+   *      the value the column actually had that was not in the list for each
+   *      failed column.
```
carolyncaron, line 155: I recall fixing the 'failedIems' typo in a previous PR but I guess it didn't stick? 😢 

```suggestion
   *  - 'failedItems': an array of the items that failed validation. For example,
   *      if this validator validates that a number of indices are not empty, then
   *      this will be an array of indices that were empty. Another example is
   *      if this validator checks that a number of indices have values in a 
   *      specific list, then this array would use the index as the key and
   *      the value the column actually had, for each failed column.
```

carolyncaron (commented):
This is my code review - manual review to follow! 😄

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
laceysanderson: We discussed fixing this in the setters PR to stop it getting lost in another merge.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
laceysanderson: Yup, good catch. I added the call to this where it will eventually be used to make sure we remember: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/101/files#diff-67c6ba893a67e6b0852eebeafa9c18ea19db7d7c7fa68a22530f746f526e89b1R485

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
laceysanderson: Good catch -I've removed this now!

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
laceysanderson: I've commented out this code now and thrown an exception if we get here.
I also documented these points in the code:

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/101/files#diff-7a382d0ff713c22d6d4e698b3803bd293a38705c195cf953cbee1c4ca0059c20R254-R255

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/101/files#diff-7a382d0ff713c22d6d4e698b3803bd293a38705c195cf953cbee1c4ca0059c20R264-R277

We'll also want to add an issue to create the suggested validator here.

trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorInterface.php
laceysanderson: Good call. I've added a comment regarding this to the header of this method including your example.

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/101/files#diff-42407f9985be3441b31c0f61bede838234ef2f4758fbd47f57120033a8b5081bR131-R135

carolyncaron (approved):
My manual review shows that it's working great!
I tried an additional case with each csv + tsv option where additional commas are inside double quotes (which follows the csv spec and should be ignored by str_getcsv) and also got the expected results 🎉 

Screenshot 2024-08-22 at 3 25 08 PM

laceysanderson:
Thank you @carolyncaron! 😍

PR #102, opened 2024-07-29
Switch tests to use chado_connection consistently.

laceysanderson:
## Motivation

We need to use a consistent variable in tests for the chado connection in order for test traits to work well. I had started to make this change in my setters PR but then realized it would be better on it's own and make my setters PR less confusing.

So here it is.

## What does this PR do?

1. Ensure that all tests connecting to chado, have a $chado_connection variable defined consistently with full type hinting.
2. Ensure that these tests all now correctly use the variable above instead of the original `chado` or `connection` variables each one had before.
3. Add the use statement for ChadoConnection to ensure the type-hinting works properly.
4. Removes any `@file` docblocks from testing files I changed since Drupal code standards says not to include them if there is only a single class in the file.

## Testing

This PR only changes automated tests so a simple code review and confirming that the tests still pass should be sufficient.

### DIFF
--- a/trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
+++ b/trpcultivate_phenotypes/tests/src/Functional/ConfigOntologyTermsTest.php
@@ -1,13 +1,8 @@
 <?php
-
-/**
- * @file
- * Functional test of Ontology and Terms configuration page.
- */
-
 namespace Drupal\Tests\trpcultivate_phenotypes\Functional;
 
 use Drupal\Component\Utility\SafeMarkup;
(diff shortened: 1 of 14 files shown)


reynoldtan:
Confirmed automated tests still pass

image

image

image

image

image

image

image

image

image

image

image

image

image

reynoldtan (approved):
Automated tests are still passing 👍

carolyncaron (approved):
Looks like all tests have been updated to use $chado_connection. I searched the code for both `chado` and `connection` separately and confirmed no remnants were left behind 👍

PR #104, opened 2024-08-19
Updates to docker/workflows based on PR tripal/tripal#1951 and base#23.

### LINKED ISSUE #103: G0.103 Update docker to meet the changes in Tripal PR #1951
### Branch

g0.103-updateDocker

### Groups

Group 0 - General, Group 5 | Automated Testing | Performance

### Describe

After changes to Tripaldocker in PR https://github.com/tripal/tripal/pull/1951we no longer have to restart postgresql. Also 10.4.x-dev is now supported. This also follows up on https://github.com/TripalCultivate/TripalCultivate/pull/23

### Design

Not applicable.

laceysanderson:
**Issue #103**

This PR adds the following:

1. Supports the new TripalDocker version as of PR https://github.com/tripal/tripal/pull/1951 which merges all PHP versions into a single Dockerfile and ensures the Docker handles keeping up PostgreSQL via Supervisord this means we no longer need to run `service postgresql restart` on a fresh TripalDocker container and also meant the test-tripal-action needs to confirm postgresql is done starting before running tests.
2. Supports reporting Test Code Coverage to CodeClimate Quality using the test-tripal-action to make these workflows easier to maintain.
3. Updates the versions of Drupal/PHP/PostgreSQL being tested, the version grid in the readme and makes the workflows easier to keep track of with better job names and comments.
4. Adds support for devcontainers in this repo including extensions to manage automated tests and workflow results in VSCode.

Testing:

This PR really just makes changes to automated testing workflows so as long as they pass here on github it should be fine. That said, you can also try out the devcontainer locally using the instructions here: https://github.com/tripal/tripal_doc/pull/92.

### DIFF
--- a/.github/workflows/ALL-PHPUnit.yml
+++ b/.github/workflows/ALL-PHPUnit.yml
@@ -42,7 +42,7 @@ jobs:
         uses: actions/checkout@v4
       # Here we pull the development tripaldocker image for this combo in our matrix
       - name: Run Automated testing
-        uses: tripal/test-tripal-action@v1.5
+        uses: tripal/test-tripal-action@v1.6
         with:
           directory-name: $PKG_NAME
           modules: $MODULES
--- a/.github/workflows/ALL-PHPUnit.yml
+++ b/.github/workflows/ALL-PHPUnit.yml
@@ -19,6 +19,7 @@ env:
 
 jobs:
   run-tests:
+    name: "Drupal ${{ matrix.drupal-version }} - PHP ${{ matrix.php-version }} - PostgreSQL ${{ matrix.pgsql-version }}"
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/ALL-PHPUnit.yml
+++ b/.github/workflows/ALL-PHPUnit.yml
@@ -1,6 +1,17 @@
 
 name: PHPUnit Full Test Suite
-on: [push]
+on:
+  push:
+    branches-ignore:
+      - 4.x
+  # Allows us to manually trigger this workflow.
+  # This is great if there is a change made to Tripal core that we want to test our modules on ASAP.
+  workflow_dispatch:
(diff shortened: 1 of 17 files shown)


.github/workflows/ALL-PHPUnit.yml
```diff
@@ -19,25 +31,19 @@ jobs:
           - "8.3"
         pgsql-version:
           - "13"
+          - "16"
```
carolyncaron, line 34: I'm more curious than concerned about this, but why the jump from Postgresql 13 to 16? What happens if I wanted to test with 14 or 15?

carolyncaron (dismissed):
This looks great to me! I tested it locally by building the container within VScode and my site has the versions I expected to see:
Screenshot 2024-08-20 at 3 39 26 PM

Thanks Lacey! 🎉

.github/workflows/ALL-PHPUnit.yml
laceysanderson: For the automated testing we check with the highest and the lowest supported versions. However, in repos where we automate building the docker images, we build images for all versions in between. 

You can see that in the build docker workflows:
- [TripalCultivate Base Images](https://github.com/TripalCultivate/TripalCultivate/blob/4.x/.github/workflows/MAIN-buildDocker.yml#L21-L103)
- [TripalDocker images](https://github.com/tripal/tripal/blob/4.x/.github/workflows/MAIN-buildDocker.yml#L21-L103)
- [TripalDocker Drupal images](https://github.com/tripal/tripal/blob/4.x/.github/workflows/SCHEDULE-buildDrupalDocker.yml#L20-L102)

We don't need to automate building of the phenotype/genotype/germplasm builds because they are not used by others really 🤷‍♀️  We could I guess but currently don't.

Because all the base images have versions with 14 or 15, you would just pass 14 as the postgresql version when building your phenotypes docker and it would happily do so :-)

Issue #105, opened 2024-08-26
G5.105 Implement Tripal Logger to log messages from validators/traits


carolyncaron:
### Branch

g5.105-loggerInValidators

### Groups

Group 5 | Automated Testing | Performance

### Describe

We recently had a discussion regarding validator and validator traits throwing exceptions. We decided we only want to throw exceptions if this issues found are developer issues. However, if the input that is invalid is provided by the user then we want to tell them nicely not crash things with an exception. With validators this is done by returning a failure of the validator with details so that the importer can portray this to the user. However, in the validator trait getter/setters we can't do this. As such, we want to use the job logger to report back to the site maintainer that this has occured and then the validator traits just leave the variable unset or null. There are then validators that check these things and provide the message to the user.

Anyway, this issue is to add a setter/getter in the validator base that allows the importer to set the TripalImporter instance it uses for the validator to use.

### Design

The Tripal Logger gives access to job logging, which is useful in many situations. For example, the setFileMimeType() method should log when the mime-type given is unsupported, since we don't want to throw an exception for the user to see (that's the DataFile validator's job).

Specifically, this issue will 

1. Add a method to allow the TripalLogger to be shared with validator and validator traits.
```php
/**
 * Sets the TripalLogger instance for the importer using this validator.
 * 
 * @param TripalLogger $logger
 *   The TripalLogger instance. In the case of validation done on the form 
 *   the job will not be set but in the case of any validation done in the 
 *   import run job, the job will be set.
 */
protected function setLogger(TripalLogger $logger);
```
2. Add a getter so that this logger can be easily used.
```php
/**
 * Provides a configured TripalLogger instance for reporting things to 
 * site maintainers.
 * 
 * @return TripalLogger
 *   An instance of the Tripal logger.
 */
protected function getLogger();
```

This can then be used as follows:
```php
$this->getLogger()->error(
  'When validating a user submitted file, the FileTypes Trait setFileMimeType() was passed a mime type (@mimetype) that we do not support. This can happen if the user submitted a file whose extension had been changed.', 
  [ '@mimetype' => $mime_type ]
);
```
You can see more docs on how to use the logger here: https://tripaldoc.readthedocs.io/en/latest/dev_guide/module_dev/logging.html#tripal-logger

carolyncaron:
Ready to close this as it was addressed in PR #107!

Issue #85, opened 2024-06-26
G4.85  Update Trait Data Validators


laceysanderson:
### Branch

g4.85-updateTraitDataValidators

### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Dependencies

- #77 
- #92
- #93

### Describe

This issue is meant to carry on from PR #77 after the discussion documented in issue #82. More specifically, it will update the validators to accept configuration using setters and to return the new return values. This will also involve updating the tests and configuring the validators in the importer formValidate.

**This is meant to update the data values validators - ie. ValueInList, EmptyCell & DuplicateTraits**

#### Task 1a
(use the getters)
Currently the validators receive their config via a temporary class property set by load values (see comment https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/79#issuecomment-2192382968) and then that property is used wholesale in the `validateRow` method. This task is to update the validators to use the setter methods developed in #93 and stop using the context property.

#### Task 1b
(use the setters)
Currently the config is set in the importer configureValidators using the loadAssets. This task is to switch to calling the individual setters developed in #95 . 

#### Task 2

Update the validators to indicate their inputType and remove their scope in the annotation block.

#### Task 3

Currently the validators return the old style of return value which includes the keys `title`, `status` and `details`. This task is to add the new design keys. We will likely keep the old keys being returned unchanged? but add a comment saying they are deprecated. Essentially, 
 - the `title` becomes more developer friendly and becomes the `case`
 - the `status` becomes the `valid` key with `'fail'` becoming `FALSE` and `'pass'` becoming `TRUE`.
 - the items appending into the `details` string (i.e. failed indices or original values) are instead individual elements in a `failedIems` array.

Here is the design information for the new return value style:
 - 'case': a developer focused string describing the case checked
 - 'valid': either TRUE or FALSE depending on if the data is valid or not
 - 'failedItems': an array of "items" to be used in the message to the user. This is empty if the data was valid

PR #106, opened 2024-09-04
g4.85 Update Validators: EmptyCell, ValueInList and DuplicateTraits

carolyncaron:
**Issue #85**

## Dependencies
#105 - complete!

## Motivation
The 3 validators I have been primarily working on - EmptyCell, ValueInList, and Duplicate Traits - needed to be updated to accept configuration using setters and to return the new return values. This involved updating the tests and configuring the validators in the importer formValidate().
 
## What does this PR do?

1. Updates each validator to use their respective traits to set and get configuration values

   **EmptyCell**: 
   - [x] Use getIndices() in validateRow()
   - [x] Use setIndices() in configureValidators()

   **ValueInList**:
   - [x] Use getIndices() in validateRow()
   - [x] Use setIndices() in configureValidators()
   - [x] Use getValidValues() in validateRow()
   - [x] Use setValidValues() in configureValidators()

   **DuplicateTraits**:
   - [x] Use getConfiguredGenus() in validateRow()
   - [x] Use setConfiguredGenus() in configureValidators()
   - [x] Use getIndices() in validateRow()
   - [x] Use setIndices() in configureValidators()

2. Updates the validators to indicate their inputType and removes their scope in the annotation block.
   - [x] EmptyCell
   - [x] ValueInList
   - [x] DuplicateTraits
   - [x] Update ValidatorFileRowScopeTest and ValidatorTraitImporterTest to remove mention of scope

3. Updates the validators to return the new style of return information:
   > 'case': a developer focused string describing the case checked
   > 'valid': either TRUE or FALSE depending on if the data is valid or not
   > 'failedItems': an array of "items" to be used in the message to the user. This is empty if the data was valid

   - [x] EmptyCell
   - [x] ValueInList
   - [x] DuplicateTraits

4. Updates were made to the formValidate() method in the Traits Importer class:
   - [x] Removed all reference to the old return style specifically for data-row validators
   - [x] Removed the declaration of the $validation array and its use- validation results are no longer stored as they happen
   - [x] Instead, validation results are now being stored in the $failures array only if validation fails. The $failures[$validation_name] array only gets declared if validation occurs at all for that validator instance, which will be used later to determine which validation statuses need to be 'pass' vs 'todo'.
   - [x] Added a new method processValidationMessages() to handle the $failures array and to define the default messages which will later be passed to the user through the UI

### Automated Testing
**Some tests were re-organized as follows:**
- [x] `ValidatorFileRowScopeTest.php` was removed
   - [x] `ValidatorFileRowScopeTest::testValidatorValueInList()` => `ValidatorValueInListTest::testValidatorValueInList()`
   - [x] `ValidatorFileRowScopeTest::testValidatorEmptyCell()` => `ValidatorEmptyCellTest::testValidatorEmptyCell()`
- [x] `ValidatorTraitImporterTest` was renamed and all of its methods moved to `ValidatorDuplicateTraitsTest`

**NEW Test added**
- [x] `testIndexKeyExceptions()` was added to `ValidatorDuplicateTraitsTest`
   - checks for each of the index keys that were configured with setIndices() using data provider `provideWrongIndexKeys()`
      - [x] Trait Name
      - [x] Method Short Name
      - [x] Unit

**The following tests were updated by adding additional asserts to check for all of the expected values in the return array**

- [x] ValidatorValueInListTest::testValidatorValueInList()
- [x] ValidatorEmptyCellTest::testValidatorEmptyCell()
- [x] ValidatorDuplicateTraitsTest::testValidatorDuplicateTraitsFile()
- [x] ValidatorDuplicateTraitsTest::testValidatorDuplicateTraitsDatabase()

**Miscellaneous updates**
- [x] TraitImporterFormValidateTest::testTraitFormValidation() has a new assert for ensure the 'details' value in the data provider is not an empty string
   - [x] also added documentation to both the data provider and test method
- [x] ValidatorBaseTest was modified to use the base fake validator instead of being dependent on an instance of the ValueInList validator

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -523,7 +523,8 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
             // Then new style.
             elseif (array_key_exists('valid', $result) && $result['valid'] === FALSE) {
               $failed_validator = TRUE;
-              //array_push($failures['data-row'][$validator_name]['failed_rows'], $line_no);
+              //$failures['data-row'][$validator_name][$line_no] = $result;
+              array_push($failures['data-row'][$validator_name]['failed_rows'], $line_no);
             }
           }
         }
@@ -534,9 +535,14 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         foreach($failures['data-row'] as $validator_name => $validator_messages) {
           if (!empty($validator_messages['failed_rows'])) {
             $validation[$validator_name] = [
+              // Old return values
               'title' => $validator_messages['fail_title'],
               'status' => 'fail',
-              'details' => $validator_messages['fail_details'] . implode(', ', $validator_messages['failed_rows'])
+              'details' => $validator_messages['fail_details'] . implode(', ', $validator_messages['failed_rows']),
+              // New return values
+              'case' => $validator_messages['fail_details'],
+              'valid' => FALSE,
+              'failedItems' => $validator_messages['failed_rows']
             ];
           }
         }
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -352,6 +352,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
     // of one validator. Then within that we expect line number
     $failures = [];
     // Preset the fail messages for the data-row validators
+    /*
     $failures['data-row'] = [
       'empty_cell' => [
         'fail_title' => 'Required columns were found to be empty',
@@ -369,6 +370,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         'failed_rows' => []
       ]
     ];
+    */
 
     // Array to hold all validation result for each level.
     // Each result is keyed by the validator scope or id.
@@ -391,16 +393,25 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
         'details' => ''
       ],
       'empty_cell' => [
+        'case' => 'No empty values found in required column(s)',
+        'valid' => TRUE,
+        'failedItems' => [],
         'title' => 'Genus exists and/or matches the project/experiment',
         'status' => 'todo',
         'details' => ''
       ],
       'valid_data_type' => [
+        'case' => 'Values in required column(s) are valid',
+        'valid' => TRUE,
+        'failedItems' => [],
         'title' => 'Genus exists and/or matches the project/experiment',
         'status' => 'todo',
         'details' => ''
       ],
       'duplicate_traits' => [
+        'case' => 'All trait-method-unit combinations are unique',
+        'valid' => TRUE,
+        'failedItems' => [],
         'title' => 'Genus exists and/or matches the project/experiment',
         'status' => 'todo',
         'details' => ''
@@ -514,35 +525,27 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
           // Call each validator on this row of the file
           foreach($validators['data-row'] as $validator_name => $validator) {
             $result = $validator->validateRow($data_row);
-            $validation[$validator_name] = $result;
-            // Check for old style...
-            if (array_key_exists('status', $result) && ($result['status'] == 'fail')) {
+            // Check if validation failed
+            if (array_key_exists('valid', $result) && $result['valid'] === FALSE) {
               $failed_validator = TRUE;
-              array_push($failures['data-row'][$validator_name]['failed_rows'], $line_no);
-            }
-            // Then new style.
-            elseif (array_key_exists('valid', $result) && $result['valid'] === FALSE) {
-              $failed_validator = TRUE;
-              //$failures['data-row'][$validator_name][$line_no] = $result;
-              array_push($failures['data-row'][$validator_name]['failed_rows'], $line_no);
+              $failures['data-row'][$validator_name][$line_no] = $result;
             }
           }
         }
       }
+
+      // @todo: Convert the following into a method
       // For each data-row validator, check if the validation status failed.
       // If so, format the validation message and set the status to 'fail'.
-      if ($failed_validator === TRUE) {
-        foreach($failures['data-row'] as $validator_name => $validator_messages) {
-          if (!empty($validator_messages['failed_rows'])) {
+      if (($failed_validator === TRUE) && (array_key_exists('data-row', $failures))) {
+        foreach($validators['data-row'] as $validator_name => $validator) {
+          if (array_key_exists($validator_name, $failures['data-row'])) {
+            $first_failed_row = array_key_first($failures['data-row'][$validator_name]);
+            $message = $failures['data-row'][$validator_name][$first_failed_row]['case'] . 'at row #: ' . $first_failed_row;
             $validation[$validator_name] = [
-              // Old return values
-              'title' => $validator_messages['fail_title'],
-              'status' => 'fail',
-              'details' => $validator_messages['fail_details'] . implode(', ', $validator_messages['failed_rows']),
-              // New return values
-              'case' => $validator_messages['fail_details'],
+              'case' => $message,
               'valid' => FALSE,
-              'failedItems' => $validator_messages['failed_rows']
+              'failedItems' => $failures['data-row'][$validator_name][$first_failed_row]['failedItems']
             ];
           }
         }
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -358,20 +358,21 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
     // ************************************************************************
     // Metadata Validation
     // ************************************************************************
-    foreach ($validators['metadata'] as $key => $validator) {
+    foreach ($validators['metadata'] as $validator_name => $validator) {
+      // Set failures for this validator name to an empty array to signal that
+      // this validator has been run
+      $failures[$validator_name] = [];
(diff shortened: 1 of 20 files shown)


trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -135,10 +171,12 @@ public function validateRow($row_values) {
     }
 
     // Check if our trait combo exists at the database level
-    // First make sure to set our genus before getting our trait combo
-    $this->service_traits->setTraitGenus($context['genus']);
+    // First make sure to get our genus and then set it in our Traits Service
+    // before we can grab our trait combo
+    $genus = $this->getConfiguredGenus();
+    $this->service_PhenoTraits->setTraitGenus($genus);
```
laceysanderson: Is this actually needed now that we use the trait? Or is this done by the trait when you configure the genus?

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
```diff
@@ -86,40 +110,52 @@ public static function create(ContainerInterface $container, array $configuratio
    *   stored as an array element.
    *
    * @return array
-   *   An associative array with the following keys:
-   *   - title: string, section or title of the validation as it appears in the result window.
-   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
-   *   - details: details about the offending field/value.
+   *   An associative array with the following keys.
+   *   - case: a developer focused string describing the case checked.
+   *   - valid: either TRUE or FALSE depending on if the genus value is valid or not.
```
laceysanderson:
```suggestion
   *   - valid: TRUE if the trait is unique and FALSE if it already exists.
```

trpcultivate_phenotypes/src/Plugin/Validators/EmptyCell.php
```diff
@@ -60,26 +57,28 @@ public static function create(ContainerInterface $container, array $configuratio
    *
    * @return array
    *   An associative array with the following keys.
-   *   - title: string, section or title of the validation as it appears in the result window.
-   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
-   *   - details: details about the offending field/value.
+   *   - case: a developer focused string describing the case checked.
+   *   - valid: either TRUE or FALSE depending on if the genus value is valid or not.
```
laceysanderson:
```suggestion
   *   - valid: FALSE if any of the cells being checked are empty and TRUE otherwise.
```

trpcultivate_phenotypes/src/Plugin/Validators/ValueInList.php
```diff
@@ -63,63 +61,66 @@ public static function create(ContainerInterface $container, array $configuratio
    *
    * @return array
    *   An associative array with the following keys.
-   *   - title: string, section or title of the validation as it appears in the result window.
-   *   - status: string, pass if it passed the validation check/test, fail string otherwise and todo string if validation was not applied.
-   *   - details: details about the offending field/value.
+   *   - case: a developer focused string describing the case checked.
+   *   - valid: either TRUE or FALSE depending on if the genus value is valid or not.
```
laceysanderson:
```suggestion
   *   - valid: FALSE if any of the cells being checked contain a value not in the 
   *     configured list of approved values and TRUE otherwise.
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -168,7 +168,7 @@ public function configureValidators($form_values) {
     // - All header row cells are not empty.
     // @TODO: Uncomment the following code when the Headers validator has been
     //        updated and no longer uses the validate() method.
-    // $instance = $manager->createInstance('empty_cell');
+    // $instance = $manager->createInstance('empty_header_cell');
```
laceysanderson: This will still need to be `empty_cell` being passed to create instance but then it will be saved as `$validators['header-row']['empty_header_cell']` to indicate it's uniqueness.

```suggestion
    // $instance = $manager->createInstance('empty_cell');
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -483,84 +401,208 @@ public function formValidate($form, &$form_state) {
...
+   * user of the importer
+   *
+   * @param array $failures
+   *   An array containing the return values from any failed validators, keyed
+   *   by the unique name assigned to each validator-input type combination
+   *
+   * @return array
+   *   An array of all the final return values for all of the validators used by
+   *   this importer, keyed by a string that is associated with a line in the
+   *   validate UI. Specifically:
```
```suggestion
   *   An array of feedback to provide to the user which summarizes the validation results 
   *   reported by the validators in the formValidate (i.e. $failures). This array is keyed 
   *   by a string that is associated with a line in the validate UI. Specifically,
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -483,84 +401,208 @@ public function formValidate($form, &$form_state) {
...
+   *     - 'status': One of: 'todo', 'pass', 'fail'.
+   *     - 'raw_results': A nested array keyed by validator name, which contains
+   *       the raw return values when validation failed. Essentially, the
+   *       contents of $failures['validator_name'].
+   */
+  public function processValidationMessages($failures) {
+    // Array to hold all validation results for each validator. Everything is
+    // set to status of 'todo' to start and will only change to one of 'pass' or
+    // 'fail' if the $failures[] array is defined for that validator, indicating
+    // that validation did take place.
```
```suggestion
    // Array to hold all the user feedback. Currently this includes an entry for each 
    // validator. However, in future designs we may combine more then one validator into a
    // single line in the validate UI and, thus, a single entry in this array. Everything is
    // set to status of 'todo' to start and will only change to one of 'pass' or
    // 'fail' if the $failures[] array is defined for that validator, indicating
    // that validation did take place.
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -483,84 +401,208 @@ public function formValidate($form, &$form_state) {
...
+        // doesn't keep track of line numbers.
+        else if (array_key_exists('case', $failures[$validator_name])) {
+          // @todo: Update the message to not use the 'case' string by default
+          // and to incorporate the 'failed_details'.
+          $message = $failures[$validator_name]['case'];
+          $messages[$validator_name] = [
+            'status' => 'fail',
+            'details' => $message
+          ];
+          $messages['raw_results'] = $failures[$validator_name];
```
laceysanderson: For these we want to ensure the raw results are at the same level of the array as the status/details. So it should be this instead:
```suggestion
            'status' => 'fail',
            'details' => $message,
            'raw_results' => $failures[$validator_name],
          ];
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
```diff
@@ -88,12 +88,25 @@ protected function setUp(): void {
 
   /**
    * Data Provider: provides files with expected validation result.
+   *
+   * For each scenario we expect the following:
+   * -- the filename of the test file used for this scenario (test files are
+   *    located in: tests/src/Fixtures/TraitImporterFiles/)
+   * -- an array indicating the expected validation results
+   *    - Each key is the unique name of a validator instance that was instant-
+   *      iated by the configureValidators() method in the Traits Importer class
```
laceysanderson:
```suggestion
   *    - Each key is the unique name of a feedback line provided to the validator UI 
   *      by the processValidationMessages(). Currently there is a feedback line for 
   *      each unique validator instance that was instantiated by the 
   *      configureValidators() method in the Traits Importer class
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
```diff
@@ -166,20 +182,36 @@ public function provideFilesForValidation() {
...
    * Tests the validation aspect of the trait importer form.
    *
+   * @param string $filename
+   *   The name of the file being tested. This file is located in: tests/src/Fixtures/TraitImporterFiles/
+   * @param array $expectations
+   *   An array that is keyed by the unique name of each validator instance
+   *   (these keys are declared in the configureValidators() method in the Traits
+   *   Importer class). Each validator instance in the array is further keyed by
+   *   the following. Some are required but others are optional, dependent upon
+   *   the expected validation results.
```
laceysanderson, line 207: If we change the data provider tobe clear that this is keyed by the unique identifier of the feedback line which just so happens to map to the validator instance right now then we should use similar wording here.

laceysanderson (changes requested):
Overall this looks great! Just a few more concrete comments:
- Do we need to set the genus in the validator or is it done by the trait?
- ensure raw_results at the same level as the title/details
- make clear in comments that the key from processValidationMessages() is the unique identifier of the validator UI line that just happens to currently be the validator instance key but may not continue to be so.

trpcultivate_phenotypes/src/Plugin/Validators/DuplicateTraits.php
carolyncaron: It is still needed for tests to pass. 🤷  TraitImporterFormValidateTest() throws this exception if I comment it out: 
```
Exception : No genus has been set. See setting a genus in the Traits Service and make sure to
        use a configured genus. To configure a genus or see all configured genus, go to /admin/tripal/extension/tripal-cultivate/phenotypes/ontology.
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
laceysanderson: We decided in discussion that we would not make these changes to the tests until we have a solid design in place :-)

laceysanderson (commented):
All the changes look great.

Only thing I noticed was that there is a gap in the test coverage that must have snuck in the previous PR 🙈 
Screenshot 2024-09-26 at 2 52 07 PM

carolyncaron:
Thanks @laceysanderson! I created a new test `testIndexKeyExceptions` and data provider `provideWrongIndexKeys` to cover all 3 of those exception messages  in commit [b65a321](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/106/commits/b65a321f0efa68d38092dafe999a9b5961649d73)

laceysanderson (approved):
Changes look good! Only remaining gaps in test coverage are related to new vs. old return style which will soon be resolved!

Ready to merge! 🎉

PR #107, opened 2024-09-06
g5.105 Implement TripalLogger to allow validators/traits to log their own messages

carolyncaron:
**Issue #105**

## Motivation

We want the ability to allow the validators to utilize the TripalLogger instance from the Tripal Importer. This will allow us to log messages in the service logs in place of exception messages where it is more appropriate to do so.

## What does this PR do?
1. Adds a setter/getter in the validator base that allows the importer to set the TripalImporter instance it uses for the validator to use.
2. Update the Trait Importer to set the TripalLogger instance for each validator
   - In this PR, this only applies to the DuplicateTraits validator, so I've added a single call to setLogger() by this instance within the configureValidators() method. 
   - **For future discussion**: We know that once the Traits Importer is updated to use all of the new validators, there will be more calls to setLogger(). @laceysanderson has suggested we can loop through all of the validators at the end of the configureValidators() method and set it for every validator regardless if they need it or not, to cut down on maintenance every time we want to add a new validator.
4. Update exception messages currently marked as `@todo` to log those messages instead
   - **GenusConfigured Trait**:
   - [x] Check that genus exists in chado
   - [x] Check that genus is configured for the module
   - **Project Trait**:
   - [x] Check if project id <= 0
   - [x] Check if project name is an empty string
   - [x] Check if project exists in the database
   - **FileTypes Trait**:
   - [x] Check if the input file's MIME type is in the mime_to_delimiter_mapping array
5. Create and update tests accordingly

### Automated Testing
**ValidatorBaseTest::testTripalLoggerGetterSetter**
- [x] Tests an exception is thrown when trying to get a logger that hasn't been set yet
- [x] Tests setLogger() with a valid TripalLogger object and ensures it doesn't throw an exception
- [x] Tests getLogger() after setting with a valid TripalLogger object to ensure we get the object back

_The following tests were **updated**_:

**ValidatorTraitProjectTest::testProjectSetterGetter**
- [x] Updated all cases where setProject() is called to check for logged messages instead of catching exceptions
- [x] Include checks that getProject() throws exceptions directly after setProject() was called and triggered a logged error message

**ValidatorTraitGenusConfiguredTest::testConfiguredGenusSetterGetter**
- [x] Converted the 2 cases checking if 1) a genus exists and 2) is configured from checking for exceptions thrown to checking for a logged error message

**ValidatorTraitFileTypesTest:testFileMimeType**
- [x] Added a check for logged error messages inside the try-catch whenever setFileMimeType() is tested

### DIFF
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
@@ -3,6 +3,7 @@
 namespace Drupal\trpcultivate_phenotypes\TripalCultivateValidator;
 
 use Drupal\Component\Plugin\PluginBase;
+use Drupal\tripal\Services\TripalLogger;
 
 abstract class TripalCultivatePhenotypesValidatorBase extends PluginBase implements TripalCultivatePhenotypesValidatorInterface {
 
@@ -265,7 +266,7 @@ abstract class TripalCultivatePhenotypesValidatorBase extends PluginBase impleme
    *   on a delimiter value.
    */
   public static function splitRowIntoColumns(string $row, string $mime_type) {
-    
+
     $mime_to_delimiter_mapping = self::$mime_to_delimiter_mapping;
 
     // Ensure that the mime type is in our delimiter mapping...
@@ -366,5 +367,45 @@ abstract class TripalCultivatePhenotypesValidatorBase extends PluginBase impleme
     else {
       throw new \Exception('Cannot retrieve file delimiters for the mime-type provided: ' . $mime_type);
     }
+  }
+
+  /**
+   * The TripalLogger service used to report status and errors to both site users
+   * and administrators through the server log.
+   *
+   * @var TripalLogger
+   */
+  public TripalLogger $logger;
+
+  /**
+   * Sets the TripalLogger instance for the importer using this validator.
+   *
+   * @param TripalLogger $logger
+   *   The TripalLogger instance. In the case of validation done on the form
+   *   the job will not be set but in the case of any validation done in the
+   *   import run job, the job will be set.
+   */
+  public function setLogger(TripalLogger $logger) {
+    $this->logger = $logger;
+  }
+
+  /**
+   * Provides a configured TripalLogger instance for reporting things to
+   * site maintainers.
+   *
+   * @return TripalLogger
+   *   An instance of the Tripal logger.
+   *
+   * @throws \Exception
+   *   If the $logger property has not been set by the setLogger() method.
+   */
+  public function getLogger() {
+    if(!empty($this->logger)) {
+      return $this->logger;
+    }
+    else {
+      throw new \Exception('Cannot retrieve the Tripal Logger property as one has not been set for this validator using the setLogger() method.');
+    }
+
   }
 }
--- a/trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
+++ b/trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
@@ -370,7 +370,7 @@ abstract class TripalCultivatePhenotypesValidatorBase extends PluginBase impleme
   }
 
   /**
-   * The TripalLogger service used to report status and errors to both site users
+   * The TripalLogger service is used to report status and errors to both site users
    * and administrators through the server log.
    *
    * @var TripalLogger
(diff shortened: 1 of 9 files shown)


trpcultivate_phenotypes/src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
```diff
@@ -367,4 +368,44 @@ public static function getFileDelimiters(string $mime_type) {
       throw new \Exception('Cannot retrieve file delimiters for the mime-type provided: ' . $mime_type);
     }
   }
+
+  /**
+   * The TripalLogger service used to report status and errors to both site users
```
reynoldtan:
```suggestion
   * The TripalLogger service is used to report status and errors to both site users
```

trpcultivate_phenotypes/src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```diff
@@ -47,15 +47,17 @@ public function setFileMimeType(string $mime_type) {
...
 
-    /** @TODO: Log the below message instead of throwing an exception
-     *         This check should occur in the DataFile validator
+    // Check if mime-type is in our mapping array
     if (!isset(self::$mime_to_delimiter_mapping[ $mime_type ])) {
-      throw new \Exception('The FileTypes Trait requires a supported mime-type but ' . $mime_type . ' is unsupported.');
+      // Since this is checking a user-provided value, the error is going to be
+      // logged and then checked by a validator so that the error can be passed
+      // to the user in a friendly way.
+      $this->logger->error("The setFileMimeType() setter requires a supported mime-type but '" . $mime_type . "' is unsupported.");
```
reynoldtan:
```suggestion
      $this->logger->error("The setFileMimeType() setter requires a supported mime-type but '$mime_type' is unsupported.");
```
The variable in this message can be translated without the dot operator (concatenate).

trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitFileTypesTest.php
```diff
@@ -307,13 +324,20 @@ public function provideMimeTypeForSetter() {
...
     // For each senario we expect the following:
     // -- scenario label to provide helpful feedback if a test fails.
-    // -- a string that is the mime-type to pass to setFileMimeType() and that
-    //    we also expect to have returned by getFileMimeType()
-    // -- an array indicating whether to expect an exception with the keys
-    //    being the method and the value being TRUE if we expect an exception
-    //    when calling it for this senario.
-    // -- an array of expected exception messages with the key being the method
-    //    and the value being the message we expect (NULL if no exception expected)
+    // -- a string that is the mime-type to pass to setFileMimeType()
```
reynoldtan, line 327: This is my first time to see this notation (double dash and single dash lines). Is this used for describing elements in data providers?

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorBaseTest.php
```diff
@@ -587,4 +588,80 @@ public function testSplitRowIntoColumnsMultiDelimiter() {
...
+      $expected_message,
+      $exception_message,
+      "The exception thrown does not have the message we expected when trying to get the Tripal Logger property but it hasn't been set yet."
+    );
+
+    // Create a TripalLogger object and set it using setLogger()
+    $my_logger = \Drupal::service('tripal.logger');
+
+    $exception_caught = FALSE;
+    $exception_message = 'NONE';
```
reynoldtan: None of the cases from this line to the closing bracket checks $exception_message, either drop this or implement an assert equals/containsString.

carolyncaron: Thanks for catching this! I opted to drop the variable (+ the one on line 646) since these cases don't expect an exception to get thrown (and the actual exception message does get printed if triggered). Addressed in commit [5699f5d](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/107/commits/5699f5d1e5527b03f63d184d3a1c8c1546e5f25e)

trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitFileTypesTest.php
carolyncaron: I just had a quick chat with Lacey about this and it isn't a standard for describing data providers, it's just what was intuitive at the time. Your way of using just single dashes in the Headers trait tests is also perfectly fine! 😄

laceysanderson (approved):
✅ Code looks good 👍

PR #108, opened 2024-09-16
G4.90 Utilize all remaining new validators

### LINKED ISSUE #90: G4.90 Update trait importer to use only new style validators
### Branch

g4.90-updateImporterWithNewValidators

### Groups

### Dependencies

- #92 
- #93 
- #87
- #88 
- #89 
- #85 

### Describe

Update trait importer to use the new genus, file and header validators by adding them to the configure and removing the old versions. These validators were created in issues #87, #88, and #89. Issue #79 has already laid the groundwork for the importer using the new validators which allows this PR to focus on the config only.

### Tasks

- [ ] Replace the old `genus`, `project`, `datafile` and `headers` validators in the Trait Importer with their new versions. 
- [ ] Update the Trait importer config to use the new setters
- [ ] Update the trait importer formValidate to use the new return value style and ignore the old one. Update any import tests with this in mind. (Only this task needs #85)
- [ ] Update the associated tests related to return value.

reynoldtan:
**Issue #90**

## Motivation

Update trait importer to use the new genus, file and header validators by adding them to the configure and removing the old versions. These validators were created in issues https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/87, https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/88, and https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/89. Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/79 has already laid the groundwork for the importer using the new validators which allows this PR to focus on the config only.

## What does this PR do?

1. Update validation of 'Genus' to `GenusExists`
   - [x] Configure and use the new `GenusExists` validator in the Traits Importer
   - [x] Remove old references to `Genus` validator in the Traits Importer
   - [x] Add a scenario to `testTraitFormValidation` for when a submitted genus is invalid
   - [x] Remove the `Genus` validator and its automated test method

2. Update validation of 'DataFile' to `ValidDataFile`
   - [x] Configure and use the new `ValidDataFile` validator in the Traits Importer
   - [x] Add a new scenario to `testTraitFormValidation` to fail this validator and update existing scenarios to check that it passes in them
   - [x] Remove old references to `DataFile` validator in the Traits Importer
   - [x] Remove the `DataFile` validator and its automated test method

3. Incorporate the new validator `ValidDelimitedFile` of input type `raw-row`
   - [x] Configure and use the new `ValidDelimitedFile` validator in the Traits Importer
      - [x] Includes adding a new section for `raw-row` validation
   - [x] Add a new scenario to `testTraitFormValidation` to fail this validator and update existing scenarios to check that it passes in them

4. Update validation of 'Headers' to `ValidHeaders`
   - **NOTE**: The `$headers` property was changed to the new nested array format expected by the `setHeaders()` trait. The old property has been renamed to `$old_headers` and changed throughout the TraitImporter class where it was used, since updating those to use the new format is outside the scope of this PR. The `formValidate()` and its helper methods now use the new `$headers` property exclusively.
   - [x] Configure and use the new `ValidHeaders` validator in the Traits Importer
   - [x] Update the existing scenario in `testTraitFormValidation` that fails this validator and ensure all other scenarios remain pass/todo as expected
   - [x] Remove old references to `Headers` validator in the Traits Importer
   - [x] Remove the `Headers` validator and its automated test method

5. Remove the old `Project` validator (NOT used by the Traits Importer)
   - [x] Remove automated tests specific to this validator
   - [x] Update `projectExists` and `projectGenusMatch` validators to match the new naming standard

### Automated Testing
3 new scenarios were added to the data provider `TraitImporterFormValidateTest::provideFilesForValidation()`:
- [x] An invalid genus is submitted through the form (File input would otherwise pass all validation)
   - triggers `GenusExists`
- [x] A valid genus and an empty file
   - triggers `ValidDataFile`
- [x] A valid genus and a file that is improperly delimited on the **second** row
   - triggers `ValidDelimitedFile`
   - **NOTE**: `ValidHeaders` will pass since the header row was properly delimited and valid. The `data-row` validators remain as `todo` since the second row (first data row) fails `ValidDelimitedFile` validation


trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -102,92 +137,80 @@ public static function create(ContainerInterface $container, array $configuratio
...
-
-    // Make the header columns into a simplified array where the header names
-    // are the values
-    $headers = array_keys($this->headers);
-
-    // Take our simplified headers array and flip the array keys and values
-    // This is the format that the validators will expect to know which indices
-    // in the row of data to act on
-    // For example: ['Trait Name'] => 0
-    $header_index = array_flip($headers);
```
reynoldtan, line 160: // Load the file object.
```suggestion
    // Open the file
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -102,92 +137,80 @@ public static function create(ContainerInterface $container, array $configuratio
...
-    $instance = $manager->createInstance($validator);
-    // @deprecated loadAssets in issue #93
-    $instance->loadAssets($project, $genus, $file_id, $headers, $skip);
-    // @TODO: Rename according to the new validator_id for scope 'FILE'
-    $validators['file']['FILE'] = $instance;
+    // - File exists and is the expected type
+    $instance = $manager->createInstance('valid_data_file');
+    // Set supported mime-types using the valid file extensions supported by this
+    // importer.
+    $supported_file_extensions = [
```
reynoldtan: You can pull the plugin annotaiton definition for file_types in line 27 and this way the two will always match. ie. $this->getDefinition()['file_types'];
```suggestion
    $supported_file_extensions = [
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -102,92 +137,80 @@ public static function create(ContainerInterface $container, array $configuratio
...
-    $instance->loadAssets($project, $genus, $file_id, $headers, $skip);
-    // @TODO: Rename according to the new validator_id for scope 'HEADERS'
-    $validators['header-row']['HEADERS'] = $instance;
+    $instance = $manager->createInstance('valid_headers');
+    // Use our $headers property to configure what we expect for a header in the
+    // input file
+    $instance->setHeaders($this->headers);
+    // Configure the expected number of columns and set it to be strict
+    $instance->setExpectedColumns($num_columns, TRUE);
+    $validators['header-row']['valid_header'] = $instance;
```
reynoldtan, line 211: In other validator instances, any call to getter/setters fall in between the instantiation and $validators assignement statement.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: This is a great suggestion! getDefinition() wasn't recognized but after a bit of digging I found that info is stored in $this->plugin_definition

This is implemented now in commit 4d5dde7

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: After discussing in person, this suggestion was meant for a previous commit and has since been resolved

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: After discussing in person, this suggestion was meant for a previous commit and has since been resolved

carolyncaron:
Commit [a48ba3d](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/108/commits/a48ba3d54a761a5f2abca6a3e295f6258e223eae) is my attempt at lowering the cognitive complexity score (previous code is kept in a comment block).

Notice the order of the validators in the $messages array - `valid_delimited_file` and `valid_headers` have been switched, which is key here. If the header row fails raw row validation, `valid_headers` will not appear in the failures array. If it passes raw row validation, the messages for `valid_headers` are processed first.

If the order bothers anyone (it kinda does for me 😛), another approach that I'm curious of trying is to separate the initial $messages array into 2 separate ones:
1. One for the validators that are run once: `genus_exists', 'valid_data_file', 'valid_header'
2. The remaining validators which are run on a per-row basis

Then we have separate foreach loops that deal with each array. At the end, combine both arrays before returning. It's very possible that we'll run into the cognitive complexity issue again once we try to formulate messages later on for the validators run per row, and splitting up into separate foreach loops might help with this.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -536,46 +650,49 @@ public function processValidationMessages($failures) {
...
+    // @TODO: an alternative way to identify the validators input type.
+    $raw_row_validators = [
+      'valid_delimited_file'
+    ];
+
+    $raw_row_failed = FALSE;
+
+    foreach (array_keys($messages) as $validator_name) {
       // Check if this validator exists in the failures array, which indicates
       // that it was run.
```
```suggestion
      // that it was run. If it was not run then continue as it is already marked 
      // todo in $messages above.
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -536,46 +650,49 @@ public function processValidationMessages($failures) {
...
+      if (empty($failures[$validator_name])) {
         // Check if $failures[$validator_name] is empty, which indicates there
         // are no errors to report for this validator.
-        if (count($failures[$validator_name]) === 0 ) {
-          $messages[$validator_name] = [
-            'status' => 'pass',
-          ];
+        // If raw row validation at any point, make sure the data row validators
+        // are not set to 'pass' and remain as 'todo' since they haven't been
+        // run on every line.
```
laceysanderson: Maybe just a little bit more here to explain how this logic will only affect data row validators and not the metadata/file/header validators. What do you think of this?

```suggestion
        // If raw row validation fails at any point, make sure the data row 
        // validators are not set to 'pass' and remain as 'todo' since they 
        // haven't been run on every line. This approach works because the 
        // order of the validators in the default messages array ensures
        // that the raw row validators are checked for failures directly
        // before the data row validators. It also assumes that only data row
        // validators are after raw row validators in the default messages array.
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
```diff
@@ -84,122 +84,228 @@ protected function setUp(): void {
...
+        'valid_delimited_file' => ['status' => 'todo'],
+        'valid_header' => ['status' => 'todo'],
+        'empty_cell' => ['status' => 'todo'],
+        'valid_data_type' => ['status' => 'todo'],
+        'duplicate_traits' => ['status' => 'todo']
+      ],
+      $num_form_validation_messages
+    ];
+
+    // #2: 2nd row of file is improperly delimited
```
laceysanderson: It would be really good to have a test scenario where the header row is improperly delimited. Just to confirm raw row is also functioning on the header and that all data row validation is skipped if the header is bad. I would use a file with two correct data rows and a badly delimited but otherwise perfect header.

laceysanderson (commented):
Good job everyone! This is almost perfect. I do suggest one more scenario in the test but it will be a simple one and a few additional comments but otherwise approve!

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php
carolyncaron: Great idea - scenario added in commit [56cae6a](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/108/commits/56cae6ae718a2e54e38e1110c92912911d3707d6) 😊

laceysanderson (approved):
Great! Thanks!

PR #113, opened 2024-10-09
Updates file validator test

### LINKED ISSUE #112: G5.112 Update automated tests in ValidDataFile to create files using the Test Trait
### Branch

g5.112-updateValidDataFileTests

### Groups

Group 5 | Automated Testing | Performance

### Describe

The tests for the validator `ValidDataFile` ([`trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorValidDataFileTest.php`](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/da06095cc44fa62e4e7698bb2e8ca94c94bd94cc/trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php)) need to be updated to use the [`createTestFile()`](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/da06095cc44fa62e4e7698bb2e8ca94c94bd94cc/trpcultivate_phenotypes/tests/src/Traits/PhenotypeImporterTestTrait.php#L181) method in the `PhenotypeImporterTestTrait` class to create test files. 

### Design

See below for an example where this is used within this module:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/da06095cc44fa62e4e7698bb2e8ca94c94bd94cc/trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterFormValidateTest.php#L238-L242

reynoldtan:
**Issue #112 Update automated test in ValidDataFile**

## Motivation
This PR will update automated test in ValidDataFile.

## What does this PR do?

1. Implement Phenotype Trait method that will create a test file.

### Automated Testing

- 
- 
- 

### Manual Testing


carolyncaron:
**Status**: This PR is currently waiting on PR #108 to be merged before it can be reviewed.

trpcultivate_phenotypes/tests/src/Traits/PhenotypeImporterTestTrait.php
```diff
@@ -183,69 +183,77 @@ protected function createTestFile($details) {
...
-    }
-    else if (array_key_exists('file', $details['content'])) {
-      $size = @filesize($full_path);
-      // Set size unless there was an error.
-      $this->assertNotFalse($size, 'Unable to determine size of test file: ' . $full_path);
-      $file->setSize($size);
-    }
+    // Reference file attributes:
+    $file_id = $file->id();
+    $file_uri = $file->getFileUri();
```
carolyncaron, line 205: Just out of curiosity, why does $file_uri need to be redefined here after it was defined on line 193? Shouldn't it still be the same or is it possible that it changed?

carolyncaron (approved):
This makes sense to me although I think it would be ultimately helpful to have @laceysanderson take a look at it, who is more familiar with the Test Trait.

trpcultivate_phenotypes/tests/src/Traits/PhenotypeImporterTestTrait.php
reynoldtan: They could be the same, I am not sure of how it's stored, but the 205 clearly implies that all subsequent modifications to the file will refer to the file uri, id etc. of the file object that it just created.

laceysanderson (commented):
This does look all good to me as well :-)

Issue #114, opened 2024-10-17
G4.114 Update validation feedback shown to the user


carolyncaron:
### Branch

g4.114-updateValidationFeedback
g4.114-updateProcessMessages

### Groups

Group 4 - API | Services | Plugins

### Describe

This issue is meant to address the overall design of how validation statuses and messages get relayed back to the user.

This is where the current design is at previous to our current milestone for this module:
![image](https://github.com/user-attachments/assets/75683aa8-6a45-4e14-af30-b7400e202d51)

### Design

We (@laceysanderson, @reynoldtan, @ruobinLiu and I) held a meeting to discuss what we'd like to see, both in terms of how it should look for all importers, and specifically the needs of the Traits Importer. These are the things we decided that can begin to be implemented and we can reconvene as needed:

- [x] Update the feedback "engine" to expect a render array from each feedback item (which may include feedback from 1+ validators)
- [x] To update the processValidationMessages() method in `TraitsImporter` to format the messages we want the user to see. We may need to entertain additional methods either in the `TraitsImporter` class or to the individual validator classes to prevent the code from being too overwhelming
- [x] We think that continuing to use status icons (representing pass/fail/todo) are helpful and even colour-coding if applicable

laceysanderson:
## Feedback Render Engine

This will render the display to the user providing feedback.

Currently it takes an array structure where each item in the array is a **Feedback Item**:
 - title: this is always displayed to the user regardless of status and is a short title describing what validation was done.
 - status: one of 'pass', 'fail', 'todo' and indicates which icon to be displayed in front of the title.
 - details: currently this is a simple string that describes what failed and is only present if the status is 'fail'. **NEW: we want this to be a render array**

Changes are that the Feedback Render Engine will simply pass the render array it is given to the Drupal Render API. The result from the render API will then be embedded in the markup required to indent it within that feedback item. The Feedback Render Engine only cares about the markup wrapping whatever is returned from the Drupal render engine.

![IMG_1431](https://github.com/user-attachments/assets/be5a36f2-8991-42a4-8004-ee5fd33b9f00)

### Tasks

 - [x] Decide on the design of the validation window
 - [x] determine markup needed to wrap details render array to achieve that
 - [x] Research drupal render engine to get it to render the render array

## Process Messages

This is the task of going from developer validation results returned from the validators to a user friendly feedback of failures which is described by a render array (i.e. this is `developer case` + `failedItems` => `Render Array`). 

[...]

Issue #91, opened 2024-06-26
G4.91 Final cleanup of Validator upgrade


laceysanderson:
### Branches

g4.91-removeOldDesignRemnants
g4.91-improveCodeStandards

remnants 
### Groups

Group 2 - Data Importing, Group 4 - API | Services | Plugins

### Dependencies

- #98 
- #99
- #106 
- #108

### Describe

Go back and remove all mention of scope and the old return value keys (i.e. title, details and status). These will be removed from all validators, importers, the annotation and plugin interface. Also remove the `getValidatorIdWithScope` method.

Also ensure that the changes work with the existing infrastructure of the module. The trait importer should be able to be used through the UI at the end of this PR. More specifically,

**PART 1 - Remove deprecations**
 - Check that the validation UI renders the messages returned by processValidationMessages(). (Do not change the messages returned by processValidationMessages())
 - Remove the `$old_headers` property and resolve any of its dependencies in the TraitsImporter class to use the new `$headers` property
 - Remove deprecated properties from the `ValidatorBase` class: `$project, $genus, $file_id, $headers, $skip`
 - Remove method `loadAssets()` from `ValidatorInterface` and `ValidatorBase`
 - Remove method `getValidatorScope()` from `ValidatorInterface` and `ValidatorBase`
 - Remove method `validate()` from `ValidatorInterface` and `ValidatorBase`
 - Remove method `getValidatorIdWithScope()` from `ValidatorManager`
 - Remove the `$validator_scope` property from the `Validator` class

**PART 2 - Improve to better meet Drupal/PHP Standards**
- Use PHP Sniffer & Beautify extension for VS code as a guide to bring `trpcultivate_phenotypes` to better standards according to Drupal and PHP
- This includes remove unnecessary use statements, using proper if/elseif statements, proper dependency injection, etc.

### Design

Pretty much exactly what the description says.

PR #116, opened 2024-10-30
G4.91 Remove all deprecations

### LINKED ISSUE #82: G4.82 Revisit the use of scope by the Validator API

carolyncaron:
**Issue TripalCultivate/TripalCultivate-Phenotypes#91**
This is also the last task needed to complete Issue TripalCultivate/TripalCultivate-Phenotypes#82!

CLOSES TripalCultivate/TripalCultivate-Phenotypes#82 
But is just progress on TripalCultivate/TripalCultivate-Phenotypes#91, the next PR will address coding standards cleanup.

## Motivation
To remove all mention of scope and outdated methods, such as the `getValidatorIdWithScope` method. These will be removed from all validators, importers, the annotation and plugin interface.

Any changes made in this PR should work with the existing infrastructure of the module. The trait importer should be able to be used through the UI.

## What does this PR do?
 - [x] Ensures the validation UI renders the messages returned by processValidationMessages().
     - [x] formValidate now skips empty lines before validating the raw row, since all of our example files were failing due to VS code automatically adding empty rows at the end of each file
     - [x] Added an additional scenario to formValidateTest to ensure line numbers are reported correctly when empty rows are involved
     - [x] Modified "file_types" in the TraitImporter annotation to be only `tsv` instead of both `tsv` and `txt`, as `splitRowIntoColumns()` doesn't accomodate mime-types with more than one possible file extension. See Issue TripalCultivate/TripalCultivate#46 
 - [x] Removes the `$old_headers` property and resolve any of its dependencies in the TraitsImporter class to use the new `$headers` property
 - [x] Removes deprecated properties from the `ValidatorBase` class: `$project, $genus, $file_id, $headers, $skip`
 - [x] Removes method `loadAssets()` from `ValidatorInterface` and `ValidatorBase`
 - [x] Removes method `getValidatorScope()` from `ValidatorInterface` and `ValidatorBase`
 - [x] Removes method `validate()` from `ValidatorInterface` and `ValidatorBase`
 - [x] Removes method `getValidatorIdWithScope()` from `ValidatorManager`
 - [x] Removes the `$validator_scope` property from the `Validator` class
 - [x] Removes an array of counts of validation status for each row of a data file that were added in PR TripalCultivate/TripalCultivate-Phenotypes#90. In case we later find we need this code, a patch has been made and attached to Issue TripalCultivate/TripalCultivate-Phenotypes#114 since this is the most likely place where we'll find out.

### Automated Testing
No tests were added in this PR, but some asserts have been removed where deprecated methods were used:
- `getValidatorScope` in `testBasicValidatorGetters()`
- `validate` in `testValidatorValidateMethods()`

I lied, one scenario was added to the data provider in `TraitImporterFormValidateTest`:
- Tests file `correct_header_emptycell_after_empty_line.tsv` which is the same scenario as `correct_header_emptycell_method.tsv` but with an empty line added to line 3 (previous line 3 is bumped to line 4).

### Manual Testing

1. You can search the codebase for this PR for the following keywords:
- `deprecate` should bring up no results
- `scope` will bring up a bunch of hits in `TripalCultivatePhenoshareImporter.php` but these can be ignored as that importer will be dealt with in a separate PR. The remaining hits should all be in `*.md` files where scope has a different meaning than its use in our original design for this module.
- Try out other keywords you can think of that might reference the previous design, like old methods or property values.
- Note that `todo` is still referenced A LOT due to being in the actual code as a validation status and there are more `@todo`s that are outside the scope of this PR since they need a lot of work.

2. To test that the module is working appropriately through the interface:
   1. Clone this branch and build a docker off of it
   2. Login to your site and create a new organism at `Tripal > Content > Add Tripal Content > Organism`
   3. Once your organism was successfully created, navigate to `Tripal > Extensions > Tripal Cultivate Phenotypes Configuration ` Go to Ontology Terms and select vocabularies for every term, just ensure they aren't all the same vocab :) Click on `Save Configuration`
   4. Navigate to `Tripal > Data Loaders > Tripal Cultivate: Phenotypic Trait Importer`
   5. Make sure your genus from Step ii) is selected, and then upload a file to test. You can find examples in `trpcultivate_phenotypes/tests/src/Fixtures/TraitImporterFiles`, ***however*** you must select:
      - a TSV file, since TXT is not yet supported
      - NOT `correct_header_no_data.tsv` since there is a known bug with this case 🐞 See Issue TripalCultivate/TripalCultivate-Phenotypes#117 
   6. Check that the validation feedback makes sense for the file uploaded. If you need help figuring out what should pass and what shouldn't, search for the file you used in `TraitImporterFormValidateTest.php` to see the expected statuses of that scenario.


trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -158,10 +147,10 @@ public function configureValidators(array $form_values, string $file_mime_type)
...
-    // Make the header columns into a simplified array
+    // Make the header columns into a simplified array for easy reference
     //  - Keyed by the column header name
-    //  - Values are the column header's poition in the $headers property (ie. its
-    //    index if we assume no keys were assigned)
+    //  - Values are the column header's position in the $headers property (ie.
+    //    its index if we assume no keys were assigned)
     $header_index = [];
     $headers = $this->headers;
     $i = 0;
```
reynoldtan: Foreach allows you to create a key that will increment each time by adding it in the foreach contstruct ie. foreach(array as $key => $value) this way you don't need to initialize a variable and keep track each time.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -24,7 +24,7 @@
  *   id = "trpcultivate-phenotypes-traits-importer",
  *   label = @Translation("Tripal Cultivate: Phenotypic Trait Importer"),
  *   description = @Translation("Loads Traits for phenotypic data into the system. This is useful for large phenotypic datasets to ease the upload process."),
- *   file_types = {"txt","tsv"},
```
reynoldtan, line 27: That is sad 😿

reynoldtan:
Performed manual test and all works as described. 

image

1. a TSV file, since TXT is not yet supported
With the modified file_types definition, the form will not let you select type other than a tsv 👍 

image

2. I tested the known bug and it generated the import job. What is happening here?

Other comments:

1. in TripalCultivatePhenotypesTraitsImporter.php the line below was called twice with the same parameters. I think the Line# 198 along with its comment can go.

```
  // Configure the expected number of columns and set it to be strict
  $instance->setExpectedColumns($num_columns, TRUE);
```

2. To match the naming standard we set for services, I suggest the following:

in line 87   protected $service_genusontology; to
```
 /**
   * Genus Ontology service.
   *
   * @var TripalCultivatePhenotypesGenusOntologyService
   */
protected TripalCultivatePhenotypesGenusOntologyService $service_PhenoGenusOntology;
```

and line 90 protected $service_traits; to
```
/**
  * Traits service.
  *
 * @var TripalCultivatePhenotypesTraitsService
 */
protected TripalCultivatePhenotypesTraitsService $service_PhenoTraits;
```

A basic search and replace will update all instances of the old service names.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: Can you expand on this a bit? I'm not sure I follow. In this particular case, I am incrementing the value and not the key, so that I can refer to the $headers_index array using the header names, and not have to guess which position they're in.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: I agree 😢 I'll create an issue to remind us to update the splitRowIntoColumns() method to accommodate more file formats!

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
reynoldtan: yes forgive me.. if you set your foreach so that you reference the key then $i in your setup is no longer needed. For example foreach($headers as $i => $column_details) ... then you can $header_index[$column_details['name']] = $i;

reynoldtan:
```
$i = 0;
    foreach ($headers as $column_details) {
      $header_index[$column_details['name']] = $i;
      $i++;
    }

to 

    foreach ($headers as $i => $column_details) {
      $header_index[$column_details['name']] = $i;
    }
```

carolyncaron: Thank you for the code example! Now I see how it works, and it is more compact this way. This change is made in commit [1461f72](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/116/commits/1461f7296d8aa9f5081033b796d35ece2713c4bb) 👍

carolyncaron:
### Replying to @reynoldtan's comment above:


> 2. I tested the known bug and it generated the import job. What is happening here?

It's definitely unexpected 😅 I think since no validation errors occur, the form is set to rebuild and thus, the job gets submitted. I created issue TripalCultivate/TripalCultivate-Phenotypes#117 so that we can start brainstorming how to prevent this for empty data.


> 1. in TripalCultivatePhenotypesTraitsImporter.php the line below was called twice with the same parameters. I think the Line# 198 along with its comment can go.
> 
> ```
>   // Configure the expected number of columns and set it to be strict
>   $instance->setExpectedColumns($num_columns, TRUE);
> ```

In this case, both are needed because `setExpectedColumns()` is being called on 2 different validator instances.


> 2. To match the naming standard we set for services, I suggest the following:
> 
> in line 87 protected $service_genusontology; to
> 
> ```
>  /**
>    * Genus Ontology service.
>    *
>    * @var TripalCultivatePhenotypesGenusOntologyService
>    */
> protected TripalCultivatePhenotypesGenusOntologyService $service_PhenoGenusOntology;
> ```
> 
> and line 90 protected $service_traits; to
> 
> ```
> /**
>   * Traits service.
>   *
>  * @var TripalCultivatePhenotypesTraitsService
>  */
> protected TripalCultivatePhenotypesTraitsService $service_PhenoTraits;
> ```
> 
> A basic search and replace will update all instances of the old service names.

Good catch! Also implemented in commit [1461f72](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/116/commits/1461f7296d8aa9f5081033b796d35ece2713c4bb) 😁

laceysanderson (approved):
Tested this through the UI and it works beautifully:
Screenshot 2024-11-01 at 4 35 05 PM

Code review is awesome and super satisfying! 😍

reynoldtan (approved):

PR #123, opened 2024-11-18
Update feedback engine template

reynoldtan:
**Issue #114 **

## Motivation
This PR will enable the validation result window template to render failed items structured as a Drupal render array. 

## What does this PR do?

1. Modified the template to render failed item details render array on the fly.
2. Implemented a script that will inspect the DOM to apply specific class names to lengthy details. 
3. Applied visual cue element that partially reveals bottom item to indicate more items for scroll.
4. Implemented a kernel automated test.

### Automated Testing

ClassName->testResultWindowDisplay:
- Test various details render array types: item, markup, list, list+table, raw row, image, link, concurrent items, and malformed render array.
- Test that validation item has the correct status-based class name.
- Test that validation item title matches the rendered markup output.
- Test the details rendered array rendered output can be found in the overall validation window markup output.

### Manual Testing

1. Checkout the branch g4.114-updateValidationFeedback
2. Setup Tripal Cultivate Phenotypes so that Traits Importer Page is accessible.
3. By pass the validation result window requiring a data file by adding the code snippet below in line number 352 of TripalCultivatePhenotypesTraitsImporter.php

Refer to this issue for details render array structures:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/114#issuecomment-2486787283

```
// Full validation result summary.
if (isset($storage[$this->validation_result])) {
....  
}
// Line 352 here
 
// Begin line to add
$validation_result = [
  [
     'title' => 'Validation Title',
     'status' => 'pass',
     'details' => []
  ],

  // Another validation result item here...
];

$form['validation_result'] = [
  '#type' => 'inline_template',
  '#theme' => 'result_window',
  '#data' => [
    'validation_result' => $validation_result
  ],
  '#weight' => -100
];

```

4. Save, clear cache and reload page.

**NOTES:** 
- do not commit changes.
- test partial reveal and scrollbars with a detailed and lengthy failed item.
- test raw row by applying #attributes and set class to tcp-raw-row.


reynoldtan:
This is the output of the feedback engine:
image

- A partial reveal (visual cue) to suggest more items using scrollbar since scrollbar is not applied by default.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+  /**
+   * {@inheritDoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Set test environment.
+    \Drupal::state()->set('is_a_test_environment', TRUE);
+
+    $this->renderer = \Drupal::service('renderer');
```
carolyncaron: Since this is a service call, you should be able to just do the following:
```suggestion
    $this->renderer = $this->container->get('renderer');
```

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+   *
+   * @return array
+   *   Each scenario/element is an array with the following values.
+   *   - A string, human-readable short description of the test scenario.
+   *   - An array that contains a set of values with each set containing the
+   *     keys title, status and details as returned by a validator. The details
+   *     key contains values of the failed items structured as
+   *     a Drupal renderarray.
+   *   - The expected class name an item is tagged with. Each validation item
+   *     corresponds to one class name based on the the value of the status key.
```
carolyncaron: I borrowed your description of the validation status array from your twig file for consistency :)

```suggestion
   *   - One or more validation status arrays. Each array contains the following
   *     keys:
   *     - 'title': Validation title text that is paired with appropriate status
   *        icon in the markup output.
   *     - 'status': The status of the validation that indicates if it passed 
   *        (pass), failed (fail) or upcoming task (todo).
   *     - 'details': A Drupal render array that will generate the required markup
   *        to output any failed items, only if 'status' is set to 'fail'.
   *   - The expected class name an item is tagged with. Each validation item
   *     corresponds to one class name based on the value of the status key.
```

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+            'status' => 'todo',
+            'details' => [
+              '#type' => 'link',
+              '#title' => 'KnowPulse',
+              '#url' => 'knowpulse.usask.ca',
+            ],
+          ],
+        ],
+        ['tcp-validate-todo'],
+      ],
```
carolyncaron: I expected for these 2 scenarios to have a status of 'fail' so that the render array will get rendered on the page. The arrays will need updating - see my manual test for the current display output.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+      // Pass, fail and todo concurrent validation items.
```
carolyncaron:
```suggestion
      // #9: Pass, fail and todo concurrent validation items.
```

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+        ['tcp-validate-fail'],
+      ],
+    ];
+  }
+
+  /**
+   * Test result window display.
+   *
+   * @dataProvider provideValidationresultRenderArray
+   */
```
carolyncaron: This docblock should include @params. Feel free to reuse the same text in the @return for the data provider!

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+        ],
+        ['tcp-validate-fail'],
+      ],
+    ];
+  }
+
+  /**
+   * Test result window display.
+   *
+   * @dataProvider provideValidationresultRenderArray
```
carolyncaron:
```suggestion
   * @dataProvider provideValidationResultRenderArray
```

carolyncaron (changes requested):
### Review from manual testing:

Testing went very well, I was able to simulate the screenshot and also test the raw row scenario. All the scrollbars worked flawlessly 💯 

I also went further and tested some of the test scenarios in `trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php`. I was surprised that the scenarios testing an image and a link provided in details was set to 'todo'. I think we would want to test that those things actually get rendered on the page, so in my testing, I set them to 'fail'. However, I didn't get an image or a link. Perhaps the render array is not properly formatted and/or is missing the correct type?

Here, you can see the output from all of my tested validation status arrays:

Screenshot 2024-11-21 at 3 50 30 PM

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+      [
+        'an image',
+        [
+          [
+            'title' => 'Validation Title Text - Todo',
+            'status' => 'todo',
+            'details' => [
+              '#theme' => 'image_style',
+              '#style_name' => 'thumbnail',
+              '#uri' => __DIR__ . '/../Fixtures/png.png',
```
laceysanderson: The URI needs to be relative to the DrupalRoot in order to render properly but here you are providing the full directory path. Try using the DRUPAL_ROOT global variable instead of __DIR__.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
laceysanderson: Yes, I agree :-)

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+              '#markup' => 'Lorem ipsum dolor sit amet',
+              'table' => [
+                '#type' => 'table',
+                '#captionzzzz' => 'Validator Case Message',
+                '#header-array' => ['Header 1', 'Header 2'],
+                '#row' => [
+                  ['Row #1 - Value #1', 'Row #1 - Value #2', 'Row #1 - Value #3', 'Row #1 - Value #4'],
+                ],
+              ],
+            ],
```
laceysanderson: This is not really a malformed render array. I would instead do something like this 😈 
```suggestion
            'details' => new \stdClass(),
```

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
```diff
@@ -0,0 +1,406 @@
...
+      );
+
+      $details_markup = $this->renderer->renderRoot($validation_item['details']);
+      $this->assertStringContainsString(
+        $details_markup,
+        $validation_window_markup,
+        'The details markup of validation input #' . $i . ' in scenario ' . $scenario . ', was not found in the validation window markup'
+      );
+    }
+  }
```
laceysanderson: This section of asserts is only testing that the parts of a single validation item are SOMEWHERE in the validation window, not that they are where they are supposed to be. For example, if you have two validations items where the first one passes and the second fails, this only checks that there is a pass class somewhere and a fail class somewhere. This means the test would still pass if they were switched or if the surrounding markup was mangled.

Instead we would want to ensure that the first item contains the elements the first item should with the wrapper markup that we expect.

laceysanderson (changes requested):

reynoldtan:
For the image and link render arrays:
I used incorrect type value for the image, replaced it with image from image_style

image

and the link, it seems that it would not render if the uri is not sanitize, validated etc by Drupal and so passing the link to fromUri() works.

image

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
reynoldtan: With the 2nd parameter type of the test method set to array this case triggers a type error.
image

laceysanderson: That error is being thrown by your test though, not Drupal or our module. As such it indicates an assumption in the testing environment 🤔

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
laceysanderson: You've now changed it to `public://logo.svg` but since the test doesn't add anything here it makes this test brittle as it will fail if that file is not in a site running the tests.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
reynoldtan: It is now set to fail for both. The render arrays have revisions too, one for incorrect #theme property value for image and the other is to sanitize the #url for link.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
reynoldtan: Super! wonder why it worked though

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
reynoldtan: I see, I had similar concern for the file in fixtures too since this directory may be subject to modification and alterations as more test are prepared. How about the logo in core/themes/olivero/logo.svg? then I can point to DRUPAL_ROOT . '/core/themes/olivero/logo.svg'

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
reynoldtan: The 😈 in the details... I tried this case and nothing is rendered 
image

carolyncaron: When I tried Lacey's suggestion through the UI I got this:
Screenshot 2024-11-22 at 4 01 54 PM

So I think the easier way to test this scenario is to test it in a separate test method, and there be a try-catch to catch the Exception.

carolyncaron:
The new render array formatting for both image and link work for me now through the UI 🎉  

The `public://logo.svg` example in the test doesn't render anything in the UI, but it worked when I substituted a random URL for an image from the web.

trpcultivate_phenotypes/tests/src/Kernel/Display/DisplayValidationResultWindowTest.php
laceysanderson: Let's just remove this scenario entirely. Instead we can add a separate test to check exceptions.

laceysanderson (commented):
Current state:

1. As discussed previously, the tests were only checking that each string occurred somewhere in the output and not confirming that it was in the correct validation item. Commit ed363a1d8086db4c107d1812bf02356e48443998 fixes this by using the Symphony DomCrawler to pull out each validation item based on a shared class.
2. The change in 1 is causing 3 scenarios to fail due to whitespace differences. More specifically, the Symphony DomCrawler returns a DomElement and when DomElement::saveHTML() is called it adds a lot of spaces and newlines in order to format the output. The call to render on the details does not produce the same whitespace formatting and therefore the strings are not equal.
3. While troubleshooting 2, I realized that the test environment is rendering things differently then when you use the same render arrays through the importer or through php-eval. This means we are not actually testing the contents of the details section at all. More specifically, the test environment does not seem to render the details render array. As such, you end up with all the markup of the validation window and even the validation items but within the details wrapper is empty in all cases.

carolyncaron (approved):

laceysanderson (approved):
The tests are still not truely representing the results seen in the UI as described in comment https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/123#pullrequestreview-2459568074.

However, we've decided that it no longer makes sense to keep this PR separate from #127 due to the following:
 - the render engine code seems solid based on all the testing we can do right now
 - the reason to keep these separate was to ensure that the render engine did not make any assumptions based on the trait importer and that goal has been realized to the best of our ability all ready
 - we don't want to merge into 4.x yet since the automated testing is still not solid
 - merging in to the other PR allows better validation and testing of processMessages
 - we don't need processMessages to be independent of the render engine.

As such, I'm marking this approved and will continue to work on the tests in Carolyns PR (linked above). Thanks for all the hard work @reynoldtan! 🎉

PR #127, opened 2024-11-22
G4.114 Finalize processValidationMessages() to generate user-friendly messages

carolyncaron:
**Issue #114**

**PR #123 has been merged into this PR, please see that PR for context.**
This PR _solely_ covers changes done to: 
  - [TripalCultivatePhenotypesTraitsImporter.php](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/127/files#diff-67c6ba893a67e6b0852eebeafa9c18ea19db7d7c7fa68a22530f746f526e89b1)
  - [TraitImporterFormValidateTest.php](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/127/files#diff-48c14597475a5b9b2e678d0744a866aa9a8fd8f8697843d7996dc93704e349d2)

## Motivation

This PR is meant to tackle changes needed to processValidationMessages() in the Traits Importer. Once the feedback validation engine has been upgraded to expect a render array to display validation errors to the user, then processValidationMessages() is updated to create those render arrays for each validator that fails.

## What does this PR do?
1. For each validator that is instantiated by the Traits Importer, create a new method that chooses the appropriate render array type/markup, takes the contents of `$failures[$validator_name]` and plugs them into the appropriate places of the render array, and finally returns the finished render array.
   - [x] `processGenusExistsFailures()`
   - [x] `processValidDataFileFailures()`
   - [x] `processValidHeadersFailures()`
   - [x] `processValidDelimitedFileFailures()`
   - [x] `processEmptyCellFailures()`
   - [x] `processValueInListFailures()`
   - [x] `processDuplicateTraitsFailures()`

2. Update `processValidationMessages()` to no longer loop through all of the failed validators, and instead checks each validator if a failure occurred and calls the correct method to build the render array.

- [x] Update `TraitImporterFormValidateTest` to check for expected elements in the render arrays (ie. contents of 'details') for each scenario.

- [x] Creates new tests in `TraitImporterProcessValidationTest` to trigger every case from the following methods and render the output to compare with expectations. Each test will have its own data provider.
   - [x] `testProcessGenusExistsFailures()`
   - [x] `testProcessValidHeadersFailures()`
   - [x] `testProcessDuplicateTraitsFailures()`

### Automated Testing
No new automated tests have been added yet, though some will need to be added to maintain 100% coverage of the new process methods.

### Manual Testing
Thanks to PR #123, we can now test this through the UI!
1. Create a new docker container from this branch.
2. Open the site in a browser and log in.
3. Configure the site by:
   - Creating a new organism (Tripal > Content > Add Tripal Content)
   - Configuring the module (Tripal > Extensions > Tripal Cultivate Phenotypes Configuration > Ontology Terms)
4. Go to Tripal > Data Loaders > Tripal Cultivate: Phenotypic Trait Importer
5. Upload test files (see: `trpcultivate_phenotypes/tests/src/Fixtures/TraitImporterFiles` for the test files used by `traitImporterFormValidateTest`) and see if the validation output is intuitive for each test case. Note that validators run once (ie. GenusExists, ValidDataFile, ValidHeaders) will show an itemized list for failed validation, whereas row-level validators (ie. ValidDelimitedFile, EmptyCell, ValueInList, DuplicateTraits) will display a table with a row per failed validation of a line in the input file. Feel free to get creative and modify the test files to try different scenarios.


trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -677,86 +679,622 @@ public function processValidationMessages($failures) {
...
+          $validation_result['failedItems']['combo_provided'][$trait],
+          $validation_result['failedItems']['combo_provided'][$method],
+          $validation_result['failedItems']['combo_provided'][$unit],
+        ]);
+      }
+    }
+    // Check which tables were created, and assign the correct message.
+    // Note that both tables can exist at the same time, hence not an 'elseif'.
+    if (array_key_exists('file', $table)) {
+      $table['file']['message'] = 'These traits were found to be duplicated within your input file.';
```
```suggestion
      $table['file']['message'] = 'These trait-method-unit combinations occurred multiple times within your input file. The line number indicates the duplicated occurrence(s).';
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -677,86 +679,622 @@ public function processValidationMessages($failures) {
...
+    }
+    if (array_key_exists('database', $table)) {
+      $table['database']['message'] = 'These traits were found to be duplicated within the database.';
```
```suggestion
      $table['database']['message'] = 'These trait-method-unit combinations have already been imported into this site. Please confirm the existing version fully represents your data. If it does, then you can remove it from your input file. If not then you need to make the names more specific.';
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -677,86 +679,622 @@ public function processValidationMessages($failures) {
...
+   *
+   * @return array
+   *   A render array of type unordered list which is used to display feedback
+   *   to the user about the case that failed and the failed items from the
+   *   input file. Each item in the list contains the genus that was selected
+   *   in the form which failed validation.
+   */
+  public function processGenusExistsFailures(array $validation_result) {
+    if ($validation_result['case'] == 'Genus does not exist') {
+      $title = 'The selected genus does not exist in the database.';
```
```suggestion
      $title = 'The selected genus does not exist in this site. Please contact your administrator to have this added.';
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -677,86 +679,622 @@ public function processValidationMessages($failures) {
...
+      $items = $validation_result['failedItems']['genus_provided'];
+    }
+    elseif ($validation_result['case'] == 'Genus exists but is not configured') {
+      $title = 'The selected genus exists in the database but is not configured for this module.';
```
```suggestion
      $title = 'The selected genus has not yet been configured for use with phenotypic data. Please contact your administrator to have this set up.';
```

laceysanderson (commented):
Code Review:

- [x] `processValidationMessages()` looks good! I like the simpler flow and the updated documentation is ❤️ 
- [x] `processGenusExistsFailures()` looks good :-) I think I would use a simple paragraph with more details to fix instead of a list that will only ever include one item BUT realistically this should not be triggerable (I tried just in case ;-p) so it's not worth changing. I did expand the message a small amount though just in case.
- [ ] `processValidDataFileFailures()` this one is really rough. I don't think we ever want to expose the FID to end users as it feels wrong and is likely not helpful to them. Instead I think it would be better if we processed these a bit more but knowing that we have the issue open requesting some changes to this validator I'm hesitant to suggest too much right now.
- [x] `processValidHeadersFailures()` I'm liking this one :-) I think the only improvement I would suggest is including above the linked list a tabular template of the column headers we expect.
- [x] `processValidDelimitedFileFailures()` looks good :-) We should add the class that Reynold made to format the raw row properly.
- [ ] `processEmptyCellFailures()` for this one do we want to combine the missing column headers for a single line in the file together into one row in the table?
- [ ] `processValueInListFailures()` might this one be better as a table where the header contains the line number and then the column headers. Then there can only be one row per line number and the invalid value detected for a given column goes in that column. I would also like this to mention what is a valid value as currently it would be hard to fix.
- [x] `processDuplicateTraitsFailures()` I made some suggestions to the messages but otherwise like this one :-)

reynoldtan:
Performed manual test as describe.

1. Test duplicate lines: 👍 
image

2. Empty cell after empty line: 👍 
image

3. Empty method: 👍 
image

4. Line improperly delimited: 👍  
image

5. Not quantitative or qualitative value in Type: 👍  
image

6. Line not properly delimited with data: 👍  
image
 
7. Incorrect header: 👍  
image

Other test scenario:

1. Testing non-existent genus:
To setup, I added an element in the genus array passed to the #options property of the $form['genus'] form render array.

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/3b884bdcb48d526f2c0ea518d3b8ffcb10a71556/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L374
```
    $active_genus['not registered'] = 'Not Registered';

    // Field genus.
    $form['genus'] = [
      ...
      '#options' => $active_genus,
      '#default_value' => $default_genus,
      '#weight' => -99,
      '#required' => TRUE,
    ];
```

image

The error can be fixed by putting the genus failed item value (string) in an array ie:
in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/2a4e5923eb883e3625d4ce0ecbf9ea3f23671a3b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L817
```
$items[] = $validation_result['failedItems']['genus_provided'];
or
$items = [ $validation_result['failedItems']['genus_provided'] ]; // as in the process for file

```
as required when setting up an item type render array. See #items property in this comment
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/114#issuecomment-2486787283

2. Invalid file: 👍 
image

Setup: I created a png file on the fly in formValidate() method and set the file id to the file id of the created file. This will bypass the valid file in upload.

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/2a4e5923eb883e3625d4ce0ecbf9ea3f23671a3b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L410

```
$file_test = File::create([
  'filename' => 'aa.png',
  'uri' => 'public://aa.png',
  'status' => 0,
]);

$file_test->save();
$file_id = $file_test->id();

```

2. Combination of errors no. 1: 👍 
Duplicate lines and empty cells
image

4. Combination of errors no. 2: 👍 
Not delimited line, Unexpected value in Type, and Duplicate lines
image


Overall, great job! It's really impressive!

carolyncaron:
Thank you @reynoldtan for the very thorough manual review! 🙏 Very clever how you forced a non-existent genus. 😅  Your suggested fixed in processGenusExistsFailures() is now implemented in commit [9b399e7](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/127/commits/9b399e7ab99bc899af29789c3491d54792eb7e58)!

reynoldtan:
Here is my code review of the process failures:

**processGenusExistsFailures:**
The failed item is always the genus as described in the docblock that means that the $items is the same for both Genus does not exits and Genus not configured.

The $items can be safely removed from if-elseif and just have one line or directly pass it to the #items key of the render array.

**processValidDataFileFailures:**
The expression of this if statement can be simplified by making it the else statement of the if-elseif-else, this way you don't have to type and maintain all these case titles.
```
if (($validation_result['case'] == 'Filename is empty string') ||
        ($validation_result['case'] == 'Invalid file id number') ||
        ($validation_result['case'] == 'Filename failed to load a file object') ||
        ($validation_result['case'] == 'File id failed to load a file object')) {
```

The log message that contains the username is missing a use statement to use Drupal\user\Entity\User;

I tried executing the line to see the username output and this is what I got:
image

Did a bit of research and here's one possible fix
image

```
use Drupal\Core\Session\AccountInterface;

$current_user = \Drupal::currentUser();
$username = $current_user->getAccountName();

dpm($username);

```
more here https://gist.github.com/radheymkumar/4093d0158ba3642a08e662965555866a

This block is right before the render array being returned and the $filename being establish is not referenced in any key of the final render array.
```
// Majority of cases return the filename, so check that it exists before
// assigning a value to help the user with debugging.
$filename = '';
if (array_key_exists('filename', $validation_result['failedItems'])) {
   $filename = 'Filename: ' . $validation_result['failedItems']['filename'];
 }
```

**processValidHeadersFailures:**
Looking good 👍 

**processValidDelimitedFileFailures:**
Looking good 👍 

**processEmptyCellFailures:**
Looking good 👍 

**processValueInListFailures:**
The $table['message'] is long, I wonder if PHP beautifier will complain if we add couple of new lines.

**processDuplicateTraitsFail:**
Looking good 👍

carolyncaron:
Thank you for your suggestions, Reynold! I addressed most of them in commit [8866975](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/127/commits/8866975515aa9ec74a8a1f150063840e1c7189b9)

> **processValidDataFileFailures:**
> The expression of this if statement can be simplified by making it the else statement of the if-elseif-else, this way you don't have to type and maintain all these case titles.
> ```
> if (($validation_result['case'] == 'Filename is empty string') ||
>         ($validation_result['case'] == 'Invalid file id number') ||
>         ($validation_result['case'] == 'Filename failed to load a file object') ||
>         ($validation_result['case'] == 'File id failed to load a file object')) {
> ```

For this one, I have plans to add exceptions to these methods in another PR for when the case does not match with anything, so I don't want to put these as the else case. Additionally, your PR #130 should help in lowering the number of cases 😄

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -884,57 +890,67 @@ public function processGenusExistsFailures(array $validation_result) {
...
+          'Filename: ' . $validation_result['failedItems']['filename'],
+        ];
+      }
+      // Log a message for the administrator to help with debugging the issue.
+      // Get the current user.
+      $current_user = \Drupal::currentUser();
+      $username = $current_user->getAccountName();
+      // Get the fid of the uploaded file.
+      $fid = $validation_result['failedItems']['fid'];
+      $this->logger->info("The user $username uploaded a file with FID $fid using the Traits Importer, but could not import it as something is wrong with the filename/FID.");
```
```suggestion
      $this->logger->info("The user $username uploaded a file with FID $fid using the Traits Importer, but could not import it as something is wrong with the filename/FID. More specifically, the case message '" . $validation_result['case'] . "' was reported.");
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -884,57 +890,67 @@ public function processGenusExistsFailures(array $validation_result) {
...
+      $message = 'The file provided could not be opened. Please contact your administrator for help.';
+      $filename = $validation_result['failedItems']['filename'];
+      $fid = $validation_result['failedItems']['fid'];
       $items = [
-        'Filename: ' . $validation_result['failedItems']['filename'],
-        'File ID: ' . $validation_result['failedItems']['fid'],
+        'Filename: ' . $filename,
       ];
+      // Log more info for the administrator.
+      $this->logger->info("The user $username uploaded a file with FID $fid using the Traits Importer, but the file could not be opened using \'@fopen\'");
```
```suggestion
      $this->logger->info("The user $username uploaded a file with FID $fid using the Traits Importer, but the file could not be opened using \'@fopen\'. Filename was '$filename'.");
```

reynoldtan (dismissed):
Looking good 👍

laceysanderson (commented):
✅ Code review of processMessages*() looks good now :-)

Code review of the automated tests:

- provideGenusExistsFailedCases
  - Data provider covers all the failed cases for this validator.
  - Test logic looks good 👍 
  - 100% test coverage
- testProcessValidHeadersFailures
  - Data provider covers all the failed cases for this validator.
  - Test logic looks good 👍 
  - 100% test coverage
- testProcessDuplicateTraitsFailures
  - Data provider covers all the failed cases for this validator.
  - Data provider also covers various combinations of failed cases.
  - Test logic looks ok :-) (See Note below) 
  - 100% test coverage

Note regarding testProcessDuplicateTraitsFailures logic:
The current logic is good and fine to stay as is. That said, I wanted to respond to the following comment:
```
// Pull out the table rows for this table case.
$selected_table_rows = $this->cssSelect("table.table-case-$table_case tbody tr td");
// Keep track of the index in $selected_table_rows. This is necessary
// since there's no way to distinguish different rows in a table.
// ie:
// Row 1: $selected_table_rows[0], ..[1], ..[2], ..[3]
// Row 2: $selected_table_rows[4], ..[5], ..[6], ..[7]
// etc...
```
There is actually a way. What you would do is use the cssSelect to select the row instead of the cell. Then within the loop you already have you can do the following:
```
$selected_rows = $this->cssSelect("table.table-case-$table_case tbody tr");
$current_row_index = 0;
foreach ($table as $expected_line_no => $validation_status) {
  $select_row_cells = (array) $selected_rows[$current_row_index]->td;
}
```
which gives you a simple array of the cells for each row.

laceysanderson (approved):

Issue #120, opened 2024-11-07
G5.120 Test for Validator Trait ColumnCount is missing checks for exception status


carolyncaron:
### Branch

g5.120-addColumnCountAsserts

### Groups

Group 5 | Automated Testing | Performance

### Describe

In PR #119, I was notified by PHP Sniffer that `ValidatorTraitColumnCountTest.php` has unused variable `$exception_caught` on lines 151, 158, 169, and 177. See code block below:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/17ba2dea7b554b588a556c09756a83ecdc2f439c/trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitColumnCountTest.php#L151-L177

### Design

Instead of removing the variables, there should be asserts added to check for it. This involves adding expected outcomes to the data provider `provideExpectedColumnsForSetter()`.

PR #128, opened 2024-12-09
Fixes unused variable problem

reynoldtan:
**Issue #[#120 ]** Test for Validator Trait ColumnCount is missing checks for exception status

## Motivation
PHP Sniffer shows test method in class ValidatorTraitColumnCountTest.php has unused variable $exception_caught on lines 151, 158, 169, and 177.

Include an expected outcome to the data provider and implement an assert test instead of removing the variable as the proposed fix.

## What does this PR do?

1. It updates the data provider to include expected outcome pertaining to whether a scenario will throw an exception or not.
2. Implement test to compare the variable to the expected value.

### DIFF
--- a/trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitColumnCountTest.php
+++ b/trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitColumnCountTest.php
@@ -82,12 +82,12 @@ class ValidatorTraitColumnCountTest extends ChadoTestKernelBase {
    *     - number_of_columns: the number of columns to expect.
    *     - strict: strict comparison flag.
    *   - Expected exception catch status and message for setter and getter:
-   *     - setter:
-   *       - catch_status: True to indicate an exception is expected.
-   *       - message: The setter exception message if catch_status is true.
-   *     - getter: getter exception message.
-   *       - catch_status: True to indicate an exception is expected.
-   *       - message: The getter exception message if catch_status is true.
+   *     - 'setter':
+   *       - 'catch_status': True to indicate an exception is expected.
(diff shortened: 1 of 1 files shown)


trpcultivate_phenotypes/tests/src/Kernel/Validators/Traits/ValidatorTraitColumnCountTest.php
```diff
@@ -81,9 +81,13 @@ protected function setUp(): void {
...
    *     - strict: strict comparison flag.
-   *   - Expected exception message for both setter and getter:
-   *     - setter: setter exception message.
+   *   - Expected exception catch status and message for setter and getter:
+   *     - setter:
+   *       - catch_status: True to indicate an exception is expected.
+   *       - message: The setter exception message if catch_status is true.
    *     - getter: getter exception message.
+   *       - catch_status: True to indicate an exception is expected.
+   *       - message: The getter exception message if catch_status is true.
```
carolyncaron:
```suggestion
   *     - 'setter':
   *       - 'catch_status': True to indicate an exception is expected.
   *       - 'message': The setter exception message if catch_status is true.
   *     - 'getter': getter exception message.
   *       - 'catch_status': True to indicate an exception is expected.
   *       - 'message': The getter exception message if catch_status is true.
```

carolyncaron (dismissed):
This looks great Reynold! I confirmed that the initial unused variable message is no longer there and there are no problems whatsoever in this test file. My only suggestion is to add single quotes around the strings used as keys in the doc block for the data provider scenario arrays. This helps to let the developer know those are keys in the array. 

Otherwise, this looks ready to merge!

laceysanderson (approved):
Looks good to me too!

PR #129, opened 2024-12-13
Fixes empty row bug

### LINKED ISSUE #117: G2.117 BUG: Files with only a header and empty data rows pass validation and submit the job
### Branch

_No response_

### Groups

### Describe

It was discovered while working on PR #116 that the example file `trpcultivate_phenotypes/tests/src/Fixtures/TraitImporterFiles/correct_header_no_data.tsv` does not fail any validators, therefore the job gets submitted despite the fact that there is only a header in the file, and no actual data. 

You can test this through the UI by following Step 2) in PR #116 using the above mentioned file.

### Design

It would be expected that the user should be informed if data is missing. This could be handled in different ways:
- The job should only be submitted if ALL validators 'pass' and none of them are set to 'todo'
- We create a separate validator that checks there is data in the file beyond the header
- Add an additional check to `ValidDataFile` somehow?

This also had me and @laceysanderson question whether validation status should ALWAYS appear when submitting the importer form, even when it passes and the job is submitted. Because it's still worth sharing feedback when everything is good 👍

reynoldtan:
**Issue #117 - File with only a header row and empty row pass validation and submit a Tripal Job.

## Motivation

A bug has been confirmed that when importing a data file with just a column header, the form will get submitted. In addition to form being submitted, A Tripal Jobs is created. 

## What does this PR do?

This PR:
1. solves both issues of form being submitted and Tripal job being created.
2. displays an error message when import has validation error and if the import was successful.
3. adds assert cases to check that form was not submitted and job was not created.

image

image

### Manual Testing

1. Fully setup a clone site with the phenotypes module installed and activated.
4. Create an organism and configure the genus accordingly.
5. Prepare 2 data files where one has error and the other is valid data file.
6. Upload each file and compare the status messages as shown in the screenshots above.
7. Upload a file with just the header row, alternatively under test/fixtures a test file with the same setup is also available.
8. Test the the form did not submit and no job was initiated.

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -590,10 +590,35 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
     $storage[$this->validation_result] = $validation_feedback;
     $form_state->setStorage($storage);
 
-    if ($failed_validator === TRUE) {
+    // Check if the $validation_feedback contains 'fail' or 'todo' status.
+    // If either is found, prevent form submission.
+    $submit_form = TRUE;
+
+    foreach ($validation_feedback as $feedback_item) {
+      if ($feedback_item['status'] == 'todo' || $feedback_item['status'] == 'fail') {
+        $submit_form = FALSE;
+
+        // No need to inspect other validators, a single instance of fail/todo
+        // is sufficient to prevent form submission.
+        break;
+      }
+    }
+
+    if ($submit_form === FALSE) {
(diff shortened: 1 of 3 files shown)


laceysanderson (approved):
✅ Code-wise everything looks solid + intuitive

Manual Testing
✅ Uploaded a file with only a header and no data rows and job submission was blocked, the error message did show and everything was as expected.
Screenshot 2024-12-20 at 4 24 53 PM

✅ Uploaded another test file with row level errors only (no todo) and confirmed that a job was not submitted, the error message did show and everything was as expected.
Screenshot 2024-12-20 at 4 26 32 PM

✅ Uploaded a file that does pass all validation with no todos and confirmed that the job WAS submitted, all validation did pass, the success message was shown and everything was as expected.
Firefox_Screenshot_2024-12-20T22-29-42 471Z

PR #130, opened 2024-12-18
Revises valid data file validator

### LINKED ISSUE #110: G4.110 Re-evaluate the parameters given to validateFile()
### Branch

g4.110-repairValidDataFile

### Groups

### Describe

I brought up during review of #98 that it was unclear which parameters given to validateFile() by the ValidDataFile validator were required or not, and whether the $fid parameter should be listed first as it is more important. Lacey clarified with me that we should only expect either $filename on its own or both parameters, since acquiring the fid of the file also means the developer has access to the filename before calling this function. See this comment thread on the PR for context:
Screenshot 2024-09-27 at 2 19 54 PM

However, I still have a few concerns that Lacey and I realized deserves its own issue, since it will involve changes in multiple places (for eg. `testValidDataFile`, `ValidatorInterface` and the validator itself). These concerns are:
- What should we do if the $fid and $filename are both provided, but they aren't of the same file? Shouldn't the developer be warned about this?
- Ideally, we would prevent a case where the developer might provide the $filename as an empty string and then the $fid, since they know that the method will happily accept it. This is how it would look when written: `validateFile('', $fid);`
- The current logic of the code assumes a file object managed by the Drupal File System, which begs the question why both parameters are even necessary. If we want to cater to situations where we aren't relying on the Drupal File System (for example: utilize a new drag-and-drop module, or use drush to call the method directly), then we'd have to change our current approach

### Design

See `validateFile()` in `trpcultivate_phenotypes/src/Plugin/Validators/validDataFile.php` for the current design.

reynoldtan:
**Issue #110** Re-evaluate the parameters given to validateFile()

## Motivation
Improve the parameter requirements of validateFile().

## What does this PR do?

1. Modified the parameters to the method by removing the filename option.
2. Removed validation cases and parameter checks relating to filename parameter.
3. Update the logic pertaining to loading of file.
4. Update validator base and validator interface to match the new parameter requirements.
5. Update calls to validateFile() with the new parameter setup.
6. Update automated test for the validator and validator base.

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
@@ -97,7 +97,8 @@ class ValidDataFile extends TripalCultivatePhenotypesValidatorBase implements Co
    */
   public function validateFile(int|null $fid) {
 
-    // Parameter check, verify the file id number is not 0 or a negative value.
+    // Parameter check, verify the file id number is not null, 0 or
+    // a negative value.
     if (is_null($fid) || $fid <= 0) {
       return [
         'case' => 'Invalid file id number',
@@ -115,24 +116,13 @@ class ValidDataFile extends TripalCultivatePhenotypesValidatorBase implements Co
 
     // Check that the file input provided returned a file object.
     if (!$file_object) {
-      if (is_null($fid)) {
-        return [
-          'case' => 'Filename failed to load a file object',
-          'valid' => FALSE,
-          'failedItems' => [
-            'filename' => $filename,
-          ],
-        ];
-      }
-      else {
-        return [
-          'case' => 'File id failed to load a file object',
-          'valid' => FALSE,
-          'failedItems' => [
-            'fid' => $fid,
-          ],
-        ];
-      }
+      return [
+        'case' => 'File id failed to load a file object',
+        'valid' => FALSE,
+        'failedItems' => [
+          'fid' => $fid,
+        ],
+      ];
     }
 
     // File object loaded successfully. Any subsequent failed checks will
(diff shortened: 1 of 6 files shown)


trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php
```diff
@@ -74,16 +74,14 @@ public static function create(ContainerInterface $container, array $configuratio
    * Validate that the input file is a valid file.
    *
    * Checks include:
-   * - Parameter filename or file id is valid.
+   * - $fid is positive number, strictly greater than 0, and cannot be null.
```
laceysanderson:
```suggestion
   * - File ID parameter is positive non-zero integer, and cannot be null.
```

laceysanderson (commented):
Code review-wise:
- I think the test case where the filename doesn't exist should be kept but just altered. Specifically, I think it should now be a managed file but after creating it, delete the file. It's expected this should fall into the 'file is not readable' case.
- Why did you remove the two valid file cases? Specifically 'Test a valid file - primary type (tsv)' and 'Test a valid file - alternative type (txt)'?
- I also noticed you removed the base validator test that validateFile() throws an exception because you said it didn't. This was a mistake so I reverted that change and it did pass (i.e. throw an exception) as expected locally.

Note: The tests are failing right now due to the experiment PR have unintended consequences on downstream modules 🤦‍♀️ I'll go look at that now.

laceysanderson (approved):
✅ Code review looks good now!

Manual review

✅  uploaded an empty file and the correct case was still triggered
Screenshot 2024-12-20 at 3 40 57 PM

✅ altered line 462 in Traitimporter to pass 0 instead of the file id and dpm'd $result from the importer. Got invalid file id as expected. Also tried with a negative number with the same results as expected.
Screenshot 2024-12-20 at 3 47 28 PM

✅ altered line 462 in Traitimporter to pass non-existing file id and dpm'd $result from the importer. Got unable to load the file case as expected.
Screenshot 2024-12-20 at 3 50 30 PM

The other cases are really hard to trigger through the UI but the automated tests show they work as well.

Good job @reynoldtan -this looks perfect! 🎉

PR #133, opened 2024-12-20
Updates file template generator service

### LINKED ISSUE #124: G2.124 Automated test needed for describeUploadFileFormat()
### Branch

g2.124-describeUploadFileFormat

### Groups

Group 2 - Data Importing

### Describe

We don't yet have an automated test that specifically checks the functionality of the `describeUploadFileFormat()` method in the Traits Importer (`TripalCultivatePhenotypesTraitsImporter.php`). While there is code coverage from other tests, we would want to ensure that the excel-compatible template is in fact being generated in the way we expect. Here is the method below (note: this may no longer be the most up-to-date version):

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/d774a374f426186fa779e3395c312243e1139404/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L747-L773

### Design

In addition to designing an automated test, also ensure that it is working as expected through the UI. 😅

reynoldtan:
**Issue #124** Automated test needed for describeUploadFileFormat().

## Motivation
This PR updates the template generator service and automated test to accommodate customization of file extension of the generated template file. The previous version supports a single file extension and is hard-coded into the logic of the class. In addition, an unused variable $notes in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/3d934ca9215e046bd890e00cce6f2388cd19445f/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L1488 was not implemented but is now added into the header information section (see image below).

image

## What does this PR do?

1. Display additional notes in the headers information section of the Importer form.
2. Replace the static file extension by improving the logic of determining the supported file types.
3. Reference static variables in validator base and validator file types to establish MIME type and delimiter.
4. Upgrade automated test.

### Automated Testing

ClassName->testTemplateGeneratorService()
- file_types annotation definition with single item - tsv.
- file_types annotation definition with multiple items but only the first item is preferred.
- A file type (txt - text/plain) where the mapping of delimiter contains multiple characters but only the first item is selected.

For every test case, the template generator is check to:
1. It creates the file uri to the template file.
2. The template file is stored in the configured directory for template files.
3. Filename is a combination of the plugin id, a string, current user username.
4. Filename contains the plugin id.
5. Filename contains the username.
6. File extension is the expected file extension.
7. The header row matches the headers array if it was encoded as string using the delimiter.
8. Separating the header row string should match the raw headers array.

### Manual Testing

1. Switch to g2.124-describeUploadFileFormat
2. In a fully setup site, paste the code below in your site/devel/php
```
$template_generator = \Drupal::service('trpcultivate_phenotypes.template_generator');

$column_headers = [
  'Header A',
  'Header B',
  'Header C'
];

$importer_id = 'my-importer';

$file_extensions = [
  'tsv',
  'csv'
];

$l = $template_generator->generateFile($importer_id, $column_headers, $file_extensions);

dpm($l);

```
3. Copy the link and paste in the location bar in another browser window.
image
4. Append the uri with the host.
localhost/site/default/file....
5. Click enter and save file when prompted. Adjust parameters to test.

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1483,6 +1483,10 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
    * notes are laid out using the theme 'importer_header'. This is rendered
    * using the referenced TWIG file below.
    *
+   * @return string
+   *   The fully rendered HTML string produced by the 'importer_header' theme
+   *   with the pertinent variables supplied by this method.
+   *
    * A template geneartor service is utilized to provide a downloadable file
    * template, pre-configured to contain all headers required. The link to
    * this template file is also formatted using the theme 'importer_header'.
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1484,7 +1484,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
    * service is utilized to provide a downloadable file template, pre-configured
    * to contain all headers required.
    *
-   * @see Tripal\src\TripalImporter\TripalImporterBase.php
+   * @see Drupal\tripal\TripalImporter\TripalImporterBase::describeUploadFileFormat()
    * @see templates\trpcultivate-phenotypes-template-importer-header.html.twig
    */
   public function describeUploadFileFormat() {
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -831,12 +831,24 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
   public function describeUploadFileFormat() {
     // A template file has been generated and is ready for download.
     $importer_id = $this->pluginDefinition['id'];
+    // The file extension of the template file.
+    // Define file properties of the template file.
+    $file_properties = [];
+    // Select the first item in the file type plugin definition list and use it
+    // as the primary file extension of the template file.
+    $file_extension = $this->plugin_definition['file_types'][0];
+    $file_properties['extension'] = $file_extension;
+    // Mime type to extension mapping.
+    $file_properties['mime'] = file_mime_type($file_extension);
+    // Delimiter.
+    $file_properties['delimiter'] = self::$mime_to_delimiter_mapping[$file_properties['mime']];
+
     // Only the header names are needed for making the template file, so pull
     // them out into a new array.
     $column_headers = array_column($this->headers, 'name');
 
     $file_link = $this->service_FileTemplate
-      ->generateFile($importer_id, $column_headers);
+      ->generateFile($importer_id, $column_headers, $file_properties);
 
     // Additional notes to the headers.
     $notes = $this->t('The order of the above columns is important and your file must include a header!
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1480,9 +1480,12 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
    * NOTE: This method supports full HTML markup output.
    *
    * All relevant information relating to expected column headers and usage
-   * notes are laid out using the theme 'importer_header'. A template geneartor
-   * service is utilized to provide a downloadable file template, pre-configured
-   * to contain all headers required.
+   * notes are laid out using the theme 'importer_header'. This is rendered
+   * using the referenced TWIG file below.
+   *
+   * A template geneartor service is utilized to provide a downloadable file
+   * template, pre-configured to contain all headers required. The link to
+   * this template file is also formatted using the theme 'importer_header'.
    *
    * @see Drupal\tripal\TripalImporter\TripalImporterBase::describeUploadFileFormat()
    * @see templates\trpcultivate-phenotypes-template-importer-header.html.twig
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1472,10 +1472,20 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
   }
 
   /**
-   * {@inheritdoc}
+   * Provide a more informative description about the template file.
    *
-   * @see src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
-   * @see src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
+   * Class TripalImporterBase is the parent class of this method and additional
+   * documentation is available in reference link below.
+   *
+   * NOTE: This method supports full HTML markup output.
+   *
+   * All relevant information relating to expected column headers and usage
+   * notes are laid out using the theme 'importer_header'. A template geneartor
+   * service is utilized to provide a downloadable file template, pre-configured
+   * to contain all headers required.
+   *
+   * @see Tripal\src\TripalImporter\TripalImporterBase.php
+   * @see templates\trpcultivate-phenotypes-template-importer-header.html.twig
    */
   public function describeUploadFileFormat() {
     // A template file has been generated and is ready for download.
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1483,13 +1483,13 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
    * notes are laid out using the theme 'importer_header'. This is rendered
    * using the referenced TWIG file below.
    *
-   * @return string
-   *   The fully rendered HTML string produced by the 'importer_header' theme
-   *   with the pertinent variables supplied by this method.
-   *
    * A template geneartor service is utilized to provide a downloadable file
    * template, pre-configured to contain all headers required. The link to
    * this template file is also formatted using the theme 'importer_header'.
+   *
+   * @return string
+   *   The fully rendered HTML string produced by the 'importer_header' theme
+   *   with the pertinent variables supplied by this method.
    *
    * @see Drupal\tripal\TripalImporter\TripalImporterBase::describeUploadFileFormat()
    * @see templates\trpcultivate-phenotypes-template-importer-header.html.twig
@@ -1519,7 +1519,6 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
 
     // Render the header and notes/lists in a template and use the file link as
     // the value to href attribute of the link to download a template file.
-
     $supported_file_extensions = implode(', ', $file_extensions);
 
     $build = [
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -26,7 +26,7 @@ use Symfony\Component\DependencyInjection\ContainerInterface;
  *   label = @Translation("Tripal Cultivate: Phenotypic Trait Importer"),
  *   description = @Translation("Loads Traits for phenotypic data into the system. This is useful for large phenotypic datasets to ease the upload process."),
  *   file_types = {"tsv"},
- *   upload_description = @Translation("Please provide a txt or tsv data file."),
+ *   upload_description = @Translation("Please provide a data file."),
  *   upload_title = @Translation("Phenotypic Trait Data File*"),
  *   use_analysis = FALSE,
  *   require_analysis = FALSE,
@@ -1512,10 +1512,14 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
 
     // Render the header and notes/lists in a template and use the file link as
     // the value to href attribute of the link to download a template file.
+
+    $supported_file_extensions = implode(', ', $file_extensions);
+
     $build = [
       '#theme' => 'importer_header',
       '#data' => [
         'headers' => $this->headers,
+        'file_extensions' => $supported_file_extensions,
         'notes' => $notes,
         'template_file' => $file_link,
       ],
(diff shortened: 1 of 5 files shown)


trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1473,23 +1473,34 @@ public function postRun() {
...
+    $file_extensions = $this->plugin_definition['file_types'];
+
     $file_link = $this->service_FileTemplate
-      ->generateFile($importer_id, $column_headers);
+      ->generateFile($importer_id, $column_headers, $file_extensions);
 
     // Additional notes to the headers.
     $notes = $this->t('The order of the above columns is important and your file must include a header!
     If you have a single trait measured in more than one way (i.e. with multiple collection
     methods), then you should have one line per collection method with the trait repeated.');
```
carolyncaron:
```suggestion
    methods), then you should have one line per collection method with the trait name/description repeated.');
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
```diff
@@ -1,94 +1,127 @@
...
+   *   An array of column headers to be written into the template file as the
+   *   column header row.
+   * @param array $file_extensions
+   *   The file extension of the template file. This is taken from the the
+   *   'file_type' plugin annotation definition of the Importer.
    *
-   * @return path
-   *   Path to the template file.
+   *   Only the first item is used as the primary file extension, in case of
+   *   multiple file type values were provided.
```
carolyncaron:
```suggestion
   *   NOTE: Only the first item is used as the primary file extension, in case
   *   of multiple file type values were provided.
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
```diff
@@ -1,94 +1,127 @@
...
+   *
+   * @return string
+   *   Abosolute path to the template file.
```
carolyncaron:
```suggestion
   *   The absolute path to the generated template file.
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
```diff
@@ -1,94 +1,127 @@
...
    *
    * @param string $importer_id
-   *   String, The plugin ID annotation definition.
+   *   String, The plugin ID annotation definition used to prefix the filename.
    * @param array $column_headers
-   *   Array keys (column headers) as defined by the header property in the importer.
+   *   An array of column headers to be written into the template file as the
+   *   column header row.
+   * @param array $file_extensions
+   *   The file extension of the template file. This is taken from the the
```
carolyncaron:
```suggestion
   *   The file extension of the template file. This is taken from the
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1473,23 +1473,34 @@ public function postRun() {
 
   /**
    * {@inheritdoc}
```
carolyncaron: The inherited doc is located in TripalImporterBase but is quite minimal and I think this function has added a lot. Might it be better to add our own documentation rather than inheriting?

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1473,23 +1473,34 @@ public function postRun() {
...
+   *
+   * @see src/TripalCultivateValidator/TripalCultivatePhenotypesValidatorBase.php
+   * @see src/TripalCultivateValidator/ValidatorTraits/FileTypes.php
```
carolyncaron: I don't know what the standards are for @see in doc blocks, but it would be helpful if it was explained briefly why I should see those 2 classes. :)

trpcultivate_phenotypes/tests/src/Kernel/ServiceTemplateGeneratorTest.php
```diff
@@ -37,26 +67,176 @@ protected function setUp() :void {
...
+   * @param string $importer_id
+   *   The importer plugin id.
+   * @param array $column_headers
+   *   The list of headers that will become the header row in file.
+   * @param array $file_extensions
+   *   The 'file_types' plugin annotation definition of the importer.
+   * @param array $expected
+   *   An array of expected values, with the following keys:
+   *     - 'file_filename': the expected filename of the template file.
+   *     - 'file_content': the expected content (header row) of the file.
```
carolyncaron:
```suggestion
   *     - 'filename': the expected filename of the template file.
   *     - 'extension': the expected file extension of the template file.
   *     - 'delimiter': the expected delimiter used to encode the header row.
   *     - 'header_row': the expected content (header row) of the file.
```

carolyncaron: I copy and pasted from the data provider doc since this section appeared to have gone out of sync.

trpcultivate_phenotypes/tests/src/Kernel/ServiceTemplateGeneratorTest.php
```diff
@@ -37,26 +67,176 @@ protected function setUp() :void {
...
+    // Generate the template file.
+    $link = $this->service_FileTemplate->generateFile($importer_id, $column_headers, $file_extensions);
 
     // Assert a link has been created.
-    $this->assertNotNull($link, 'Failed to generate template file link.');
+    $this->assertNotNull($link, 'Failed to generate template file link in scenario ' . $scenario);
 
-    // Works locally but fails in actions. Importer share functional test implements the same file size check.
-    // $this->assertGreaterThanOrEqual(1, filesize($link), 'The template file generated is empty.');
+    // Assert that a file has been created in the the configured directory
```
carolyncaron:
```suggestion
    // Assert that a file has been created in the configured directory
```

carolyncaron (changes requested):
Overall, this is looking great! I made some minor suggestions in comments/documentation but the code overall looks good to me 👍 

For manual testing, the thing that stood out right away was the following deprecation notice:
Screenshot 2025-01-14 at 4 02 50 PM
We've discussed this a bit with Lacey and realized that fixing this may be out of the scope of this PR. But I'm documenting it here as something that popped up while testing. The error also shows up on the Traits Importer page as well 🙈 

When I switch the order of tsv and csv in the example PHP Execute code, it generates a template with comma separated headers as expected. But, the help text on line 2 tells the user to use the TAB KEY for delimiting their file, which isn't ideal 😢 
Screenshot 2025-01-14 at 4 07 11 PM

I did confirm that downloading a template file from the Traits Importer page still works as expected 👍

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
reynoldtan: This appears to be incorrect, from the manual test, the output is a relative path.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
reynoldtan: According to PROBLEMS tab, @see is a path to file, any @see with a text (to describe or other purpose) triggers a problem.

image

Now I realized that it should be in this method
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/56dfa0c771723cebe9fc6bd3184e1ee525c1f168/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php#L72

as reference regarding the mapping variables used then there is no need to explain. Good catch!

reynoldtan:
I found the deprecation notice in the official PHP website

https://www.php.net/manual/en/language.oop5.traits.php

image

It looks like we need to access the mapping property using a getter method ie.

```
trait FileTypes {
  ...
  public static array $extension_to_mime_mapping = [
    'tsv' => ['text/tab-separated-values'],
    'csv' => ['text/csv'],
    'txt' => ['text/plain'],
  ];


  public static function getExtensionToMIME() {
    return self::$extension_to_mime_mapping;
  }

 ...

FileTypes::getExtensionToMIME();

instead of
FileTypes::$extension_to_mime_mapping; // Deprecated PHP 8.1

```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
carolyncaron: Oops, at least now we've been made aware of it!

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: Ahh, makes sense! For the future though, you could just add a sentence or two above the @see tags describing why they're needed, not necessarily on the @see line itself

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
reynoldtan: I updated to indicate a relative path.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
reynoldtan: I expanded the inheritdoc with references

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
reynoldtan: it is serous about leaving the @see alone 😸 
image

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1472,24 +1472,45 @@ public function postRun() {
...
+   * documentation is available in reference link below.
+   *
+   * NOTE: This method supports full HTML markup output.
+   *
+   * All relevant information relating to expected column headers and usage
+   * notes are laid out using the theme 'importer_header'. A template geneartor
+   * service is utilized to provide a downloadable file template, pre-configured
+   * to contain all headers required.
+   *
+   * @see Tripal\src\TripalImporter\TripalImporterBase.php
```
laceysanderson: This should be the namespaced class and should point to the particular method.

```suggestion
   * @see Drupal\tripal\TripalImporter\TripalImporterBase::describeUploadFileFormat()
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1472,24 +1472,45 @@ public function postRun() {
   }
 
   /**
-   * {@inheritdoc}
+   * Provide a more informative description about the template file.
```
laceysanderson: This feels more narrow then this method actually is. What about

```suggestion
   * Describe the upload format including column descriptions + template file.
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1472,24 +1472,45 @@ public function postRun() {
...
+   *
+   * Class TripalImporterBase is the parent class of this method and additional
+   * documentation is available in reference link below.
+   *
+   * NOTE: This method supports full HTML markup output.
+   *
+   * All relevant information relating to expected column headers and usage
+   * notes are laid out using the theme 'importer_header'. A template geneartor
+   * service is utilized to provide a downloadable file template, pre-configured
+   * to contain all headers required.
```
laceysanderson: This could be reworded to be more clear and to make the connection between the twig file and the mentioned theme.

```suggestion
   * All relevant information relating to expected column headers and usage
   * notes are laid out using the theme 'importer_header'. This is rendered
   * using the referenced TWIG file below.
   *
   * A template geneartor service is utilized to provide a downloadable file
   * template, pre-configured to contain all headers required. The link to
   * this template file is also formatted using the theme 'importer_header'.
```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -1472,24 +1472,45 @@ public function postRun() {
...
+   *
```
laceysanderson, line 1554: Since you didn't use the inheritdoc, we need to redocument the return value.

```suggestion
   *
   * @return string
   *   The fully rendered HTML string produced by the 'importer_header' theme
   *   with the pertinent variables supplied by this method.
   *
```

laceysanderson (changes requested):
This PR looks mostly good :-) I just have a number of suggestions regarding the docblock for the describeUploadFormat.

laceysanderson (commented):
Code coverage of the service is 100% however, there is no coverage for the describeUploadFormat() method in the importer...

laceysanderson (dismissed):
This all looks good to me now + thanks to the additional test we now have 100% coverage on both the template generator and the describeUploadFileFormat() method in the Trait Importer.

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: I think that expanding the docblock above describeUploadFileFormat() definitely helps in giving context as to why the current 2 files are listed, so I'll mark this as resolved 👍

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesFileTemplateService.php
```diff
@@ -1,94 +1,137 @@
...
     if (!file_exists($dir_template_file)) {
-      mkdir($dir_template_file, 0777, true);
+      mkdir($dir_template_file, 0777, TRUE);
     }
 
-    // Convert the headers array into a tsv string value and post into the first line of the file.
-    $fileheaders = implode("\t", $column_headers) . "\n# DELETE THIS LINE --- START DATA HERE AND USE TAB KEY #";
+    // Convert the headers array into a delimited string value and post into the
+    // first line of the file.
+    $fileheaders = implode($file_delimiter[0], $column_headers) . "\n# DELETE THIS LINE --- START DATA HERE AND USE APPROPRIATE DELIMITER/VALUE SEPARATOR KEY #";
```
carolyncaron: This looks much better :+1: But I would remove the word "KEY" since I think it might be confusing. I think "Use appropriate delimiter/value separator" is clear enough!

```suggestion
    $fileheaders = implode($file_delimiter[0], $column_headers) . "\n# DELETE THIS LINE --- START DATA HERE AND USE APPROPRIATE DELIMITER/VALUE SEPARATOR #";
```

carolyncaron (dismissed):
For the deprecation notice, it ended up being a really easy fix in commit [f9f45bd](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/133/commits/f9f45bd8db01ee1992eb36aecabc2d562100e161)

After looking into it further, the deprecation was not fixed by the addition of `use FileTypes;` alone, because `FileTypes::$extension_to_mime_mapping` still uses the trait version of the static variable. So changing it to `self` removed the notice because it now uses the inherited version of the static variable.

That said, we should still look into migrating static properties and methods to their own generic class in the future to make them more accessible to all plugins and services. This will make more sense to tackle after we migrate a lot of our validator code to TripalCultivate Base (see issue https://github.com/TripalCultivate/TripalCultivate/issues/36).

carolyncaron (approved):
Looks great! 👍

PR #137, opened 2025-01-02
g2.134 Add tests for remaining validation processing methods

### LINKED ISSUE #134: G2.134 Outstanding tasks to complete processing of messages to the user
### Branch

G2.134-finishProcessTests
g2.134-addExceptionsToProcessors

### Groups

Group 2 - Data Importing, Group 5 | Automated Testing | Performance

### Describe

This issue outlines development tasks needed to follow PR #127 after its merged. Note that separate PRs can be done to address each task:

- [x] Finish creating new tests in `TraitImporterProcessValidationTest` to trigger every case from each processing method implemented in PR #127 and render the output to compare with expectations. Each test will have its own data provider.
   - [x] `testProcessValidDataFileFailures()`
   - [x] `testProcessValidDelimitedFileFailures()`
   - [x] `testProcessEmptyCellFailures()`
   - [x] `testProcessValueInListFailures()`
   
- [x] Add checks to each new processing method that:
   - [x] Validation statuses are indeed set to 'fail' and that 'failedItems' contains content. Otherwise, throw an exception and update the tests from task 1) accordingly.
   - [x] That the case message is one of the anticipated cases, otherwise throw an exception.

### Design

The design for task 1) should follow the standard set by these test methods from PR #127:
- `testProcessGenusExistsFailures()`
- `testProcessValidHeadersFailures()`
- `testProcessDuplicateTraitsFailures()`

carolyncaron:
**Issue #134**

## Motivation

Finish creating new tests in TraitImporterProcessValidationTest to trigger every case from each processing method implemented in PR #127 and render the output to compare with expectations. Each test will have its own data provider.

## What does this PR do?
Implements the following test methods and their associated data provider:
   - [x] `testProcessValidDataFileFailures()`
   - [x] `testProcessValidDelimitedFileFailures()`
   - [x] `testProcessEmptyCellFailures()`
   - [x] `testProcessValueInListFailures()`

Updates were also made to the process methods themselves (in `TripalCultivatePhenotypesTraitsImporter.php`) to further simplify, make consistent with eachother, and add html classes to the render arrays.

One update to the `ValidDelimitedFile` validator was made to return the `$raw_row` parameter as it was provided in the case of it being empty. This was made to simplify testing (no separate check for a "raw row is empty" returned value) but also may help troubleshoot future edge cases where perhaps the line contains "gremlins" that show up as empty.

### Automated Testing
Adds 4 tests and data providers to: `/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php` (See checklist above)

These tests bring coverage of the Traits Importer back up to 100% 🎉 

### Manual Testing
To verify the changes that have been made to the process methods, you can add a `print_r($rendered_markup);` under the line:

> `$this->setRawContent($rendered_markup);`

in each of the test methods. Run the test, then copy the printed output from any of the data provider scenarios and paste using PHP Execute in a TripalCultivate Phenotypes docker container, or more simply, in an online HTML renderer such as https://htmledit.squarefree.com/

You can also check the class of each render array to make sure each process method produces a uniquely-classed rendered object and, when multiple tables appear for a single scenario (should only happen for `ValidDelimitedFile` and `DuplicateTraits`), that each table is distinguishable.

### DIFF
--- a/trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
@@ -45,6 +45,11 @@ class ValidDelimitedFile extends TripalCultivatePhenotypesValidatorBase {
    *   - 'failedItems': an array of items that failed with any of the following
    *      keys. This is an empty array if row is properly delimited.
    *      - 'raw_row': The raw row or a string indicating the row is empty.
+   *      - 'expected_columns': The number of columns expected in the input file
+   *        as determined by calling getExpectedColumns().
+   *      - 'strict': A boolean indicating whether the number of expected
+   *        columns by the validator is strict (TRUE) or is the minimum number
+   *        required (FALSE).
    */
   public function validateRawRow(string $raw_row) {
 
--- a/trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
+++ b/trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
@@ -44,7 +44,7 @@ class ValidDelimitedFile extends TripalCultivatePhenotypesValidatorBase {
    *   - 'valid': TRUE if the raw row is properly delimited, FALSE otherwise.
    *   - 'failedItems': an array of items that failed with any of the following
    *      keys. This is an empty array if row is properly delimited.
-   *      - 'raw_row': The raw row or a string indicating the row is empty.
+   *      - 'raw_row': The raw row as it was provided.
    *      - 'expected_columns': The number of columns expected in the input file
    *        as determined by calling getExpectedColumns().
    *      - 'strict': A boolean indicating whether the number of expected
@@ -59,7 +59,7 @@ class ValidDelimitedFile extends TripalCultivatePhenotypesValidatorBase {
         'case' => 'Raw row is empty',
         'valid' => FALSE,
         'failedItems' => [
-          'raw_row' => 'is an empty string value',
+          'raw_row' => $raw_row,
         ],
       ];
     }
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -1275,7 +1275,9 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#items' => [
         [
           [
+            '#prefix' => '<div class="case-message">',
             '#markup' => $table['message'],
+            '#suffix' => '</div>',
           ],
           [
             '#type' => 'table',
--- a/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
+++ b/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
@@ -860,7 +860,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#title' => $message,
       '#wrapper_attributes' => [
         'class' => [
-          'genus-exists-failures',
+          'tcp-genus-exists-failures',
         ],
       ],
       'items' => [
@@ -952,7 +952,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#title' => $message,
       '#wrapper_attributes' => [
         'class' => [
-          'valid-data-file-failures',
+          'tcp-valid-data-file-failures',
         ],
       ],
       'items' => [
@@ -1007,7 +1007,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#type' => 'ul',
       '#attributes' => [
         'class' => [
-          'valid-headers-failures',
+          'tcp-valid-headers-failures',
         ],
       ],
       '#items' => [
@@ -1162,7 +1162,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#type' => 'ul',
       '#attributes' => [
         'class' => [
-          'valid-delimited-file-failures',
+          'tcp-valid-delimited-file-failures',
         ],
       ],
       '#items' => $tables,
@@ -1226,7 +1226,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#type' => 'ul',
       '#attributes' => [
         'class' => [
-          'empty-cell-failures',
+          'tcp-empty-cell-failures',
         ],
       ],
       '#items' => [
@@ -1332,7 +1332,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#type' => 'ul',
       '#attributes' => [
         'class' => [
-          'value-in-list-failures',
+          'tcp-value-in-list-failures',
         ],
       ],
       '#items' => [
@@ -1463,7 +1463,7 @@ class TripalCultivatePhenotypesTraitsImporter extends ChadoImporterBase implemen
       '#type' => 'ul',
       '#attributes' => [
         'class' => [
-          'duplicate-traits-failures',
+          'tcp-duplicate-traits-failures',
         ],
       ],
       '#items' => $tables,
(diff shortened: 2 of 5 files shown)


reynoldtan:
I have reviewed the code for the methods and performed the manual test proceedures:

✅ Rendered markup is the expected HTML tag.
✅ Rendered markup has the expected CSS class selector name.
✅ Failed Item name and failed item are rendered correctly and in order.
✅ Error messages are instructive and detailed.
✅ Source code for test and process message methods use a similar code structure and are easy to follow.
✅ PROBLEMS tab has reported no major code error.

Suggestions and comments:

1. The method processValidDataFileFailures() has new lines inconsistent with the rest of the process methods.
- A new line after the opening curly brace of the method.
- A new line between each elseif.

2. testProcessValueInListFailures() has couple of print statements that are causing 6 problems.
image

3. Help differentiate system or other module CSS class selector names from our  own module class names by prefixing names with a unique string. ie. tcp-raw-row (tcp as tripal cultivate phenotypes).

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/6a650916547cef473a2534dd670f3968d227fdbd/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L959

4. The render array has a css class selector (ie. valid-data-file-failures or case-message) in the wrapper element and all tests use this class to reference the element but in testProcessValidDataFileFailures(), it uses form-item instead of div.valid-data-file-failures. 

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/6a650916547cef473a2534dd670f3968d227fdbd/trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php#L367

5. The second part of the message (in bold font) of ValueInList needs a minor rewording to imply this is applicable if there was an empty cell.

Applies to this example
image

not this
image

carolyncaron:
Thank you @reynoldtan for your review! I addressed points 1) and 5) in commit [fe8dd4a](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/137/commits/fe8dd4a4173d93218e39ab12a5cb4169337caa87)

For 5), please let me know if the bolded message is more clear for both situations now! 
> **If any cell in the table below is empty, then the value given in the file for that cell was one of the allowed values.**

For 2) I couldn't find any commented print statements in the file, perhaps they were added while you were testing? Please let me know if there's anything I missed.

Both 3) and 4) are good points. I think I'll wait for @laceysanderson's input before addressing them. Note that `testProcessValidDataFileFailures()` was made consistent with `testProcessGenusExistsFailures()`, since those process methods are the only 2 that don't render any tables.

reynoldtan:
image

image

That's better 👍

laceysanderson:
> Both 3) and 4) are good points. I think I'll wait for @laceysanderson's input before addressing them. Note that testProcessValidDataFileFailures() was made consistent with testProcessGenusExistsFailures(), since those process methods are the only 2 that don't render any tables.
> 
> > 3. Help differentiate system or other module CSS class selector names from our  own module class names by prefixing names with a unique string. ie. tcp-raw-row (tcp as tripal cultivate phenotypes).
> > 
> > https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/6a650916547cef473a2534dd670f3968d227fdbd/trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php#L959
> > 
> > 4. The render array has a css class selector (ie. valid-data-file-failures or case-message) in the wrapper element and all tests use this class to reference the element but in testProcessValidDataFileFailures(), it uses form-item instead of div.valid-data-file-failures. 
> > 
> > https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/6a650916547cef473a2534dd670f3968d227fdbd/trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php#L367

3) I'm not sure I feel this is necessary as the classes are already very unique to our modules 🤔 That said, it doesn't hurt either 🤷‍♀️ 

4) I think that Reynolds suggestion is to update the test cssSelect() to use the class `div.valid-data-file-failures` instead of `div.form-item`. If that's the case, I agree as it makes the test a little less brittle 🤷‍♀️ @reynoldtan you weren't suggesting that processValidDataFileFailures() should not use a form item to report the failures, right?

laceysanderson (commented):
I think we should be consistent with all places that we have added a `case-message` class and also include a class specific to that case (even if there is only one case). The same goes for the case-specific class on the table.

For example, in processValidHeaderFailures() there is only a single case and a list with table is created. You have already added the validator-specific class to the item_list but I think there should also be a case specific class added to the message and to the table (see following code example).

Note: The variable `$case-key` currently doesn't exist but rather is something I would set in the if-elseif based on the case. In this example, it might be `empty-header`, `unexpected-header-value`, and `invalid-header-colnum` respectively.

```diff
    $render_array = [
      '#theme' => 'item_list',
      '#type' => 'ul',
      '#attributes' => [
        'class' => [
          'valid-headers-failures',
        ],
      ],
      '#items' => [
        [
          [
-             '#prefix' => '<div class="case-message">',
+             '#prefix' => '<div class="case-message case-$case_key">',
            '#markup' => $message,
            '#suffix' => '</div>',
          ],
          [
            '#type' => 'table',
-             '#attributes' => [],
+             '#attributes' => [
+               'class' => [
+                 'table-case-' . $case_key,
+               ],
+             ],
            '#rows' => [
              [
                'data' => [
                  'header' => [
                    'data' => 'Expected Headers',
                    'header' => TRUE,
                  ],
                ] + $expected_headers,
                'class' => ['expected-headers'],
              ],
              [
                'data' => [
                  'header' => [
```

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
```diff
@@ -54,7 +59,7 @@ public function validateRawRow(string $raw_row) {
         'case' => 'Raw row is empty',
         'valid' => FALSE,
         'failedItems' => [
-          'raw_row' => 'is an empty string value',
+          'raw_row' => $raw_row,
```
laceysanderson, line 62: I think all failing cases should consistently have the same keys which means there are a few here (including this one and 'None of the delimiters supported by the file type was used') now missing 'expected_columns' and 'strict'.

carolyncaron: Unfortunately this inconsistency impacts more validators than just ValidDelimitedFile and I'm not sure it makes sense to try and force it to be consistent 😞 
See [ValidDataFile](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4.x/trpcultivate_phenotypes/src/Plugin/Validators/ValidDataFile.php) for our most extreme example of this. There are a lot of different failedItems keys but many are not possible to access in certain cases.

carolyncaron:
I addressed points 3) and 4) by Reynold in commit [1036fac](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/137/commits/1036fac1d3ed53d42932607c4dd06040e62219ac) after Lacey gave input. Thank you both 🤗  @reynoldtan please let me know if my changes for 4) was not what you had in mind!

trpcultivate_phenotypes/src/Plugin/Validators/ValidDelimitedFile.php
laceysanderson: Ahhh, I see 🙈 I retract my suggestion in that case :-)

carolyncaron:
> I think we should be consistent with all places that we have added a `case-message` class and also include a class specific to that case (even if there is only one case). The same goes for the case-specific class on the table.
> 
> For example, in processValidHeaderFailures() there is only a single case and a list with table is created. You have already added the validator-specific class to the item_list but I think there should also be a case specific class added to the message and to the table (see following code example).
> 
> Note: The variable `$case-key` currently doesn't exist but rather is something I would set in the if-elseif based on the case. In this example, it might be `empty-header`, `unexpected-header-value`, and `invalid-header-colnum` respectively.

Thank you for this detailed explanation! After more in person discussion, we decided to maybe leave this suggestion for if/when the requirement for case-specific classes comes up. The reason is that it could potentially take too much time to come up with the case keys and verify they all appear in the rendered output 🙈

laceysanderson (approved):
As mentioned in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/137#issuecomment-2599316367, I agree that the classes, although nice, are maybe not worth the effort right now.

With that all my suggestions are handled to my satisfaction and I think this code is ready to be merged.

reynoldtan (approved):
The code looks great, and I approve of it. Well done!

PR #139, opened 2025-01-21
g2.134 Add sanity checks to ensure validation result arrays are formatted properly

carolyncaron:
**Issue #134**

## Motivation
Right now, we are assuming that the validation result array is consistently and appropriately formatted by each validator. We would like to make some checks to ensure this is the case and throw an exception to warn the developer when we encounter something unexpected.

## What does this PR do?
- [x] Creates a method for checking the validation status array is properly formatted, ie:
   - [x] Contains the expected keys 'case', 'valid' and 'failedItems'.
   - [x] The key 'valid' has been set to the value 'FALSE' to indicate failed validation. 
   - [x] The key 'failedItems' contains values in the form of a non-empty array.
- [x] Add checks to each process method for each validation result array (note: row-level validator return one array per row that's been validated)
   - [x] Call the previously mentioned method to check the format of the array.
   - [x] The case message is one of the anticipated cases for that validator.
   - [x] The case message is not set to the message provided by the validator when validation passes.
- [x] Update the tests to check for all exception messages
   - [x] Method and data provider for the generic method that checks the validation status array.
   - [x] Method and data provider that passes in invalid case messages to each of the process methods. 

### Automated Testing
TraitImporterProcessValidationTest->**testCheckValidationStatusArray**
Tests the following scenarios:
- 'case' and 'failedItems' keys are missing
- 'valid' has not been set to FALSE
- 'failedItems' contains a value of empty array
- 'failedItems' contains a value that is not an array
- the validation status array is formatted correctly and shouldn't trigger an exception

TraitImporterProcessValidationTest->**testProcessFailuresExceptions**
Tests the following scenarios for EACH process validator method:
- The validation result array contains the passed validation 'case' message, despite 'valid' being set to FALSE and 'failedItems' having content.
- The validation result array contains an unrecognizable case message, with 'valid' set to FALSE and 'failedItems' containing content.

### Manual Testing
This is not possible to manually test, since the exceptions added in this PR should only be thrown if the developer makes a mistake in formatting the validation result returned by a validator.


reynoldtan:
This is a good check to make sure that validation result array keys exist before further processing it. 
My only suggestion is:

```
    // Check the format of the validation_result parameter.
    try {
      $this->checkValidationStatusArray($validation_result);
    }
    catch (\Exception $e) {
      throw new \Exception("The validation result array returned by the GenusExists validator was not formatted correctly. Details: " . $e->getMessage());
    }
```

Is there a way to include this block into the checkValidationStatusArray() that way it will save us from implementing this block in every validator failures processor. For example:

```
public function checkValidationStatusArray(array $validation_result, $validator, $extra_info = NULL) {
  ....
  $failed_message = 'Expected to find the key \'$key\' in the validation result array';
  $has_failed = TRUE; 
  ....

  if ($has_failed) {
    $failed_message .= ($extra_info) ? '@ Line#: ' . $extra_info : '';
    throw new \Exception('The validation result array returned by the ' . $validator . ' validator was not formatted correctly. Details: ' .  $failed_message . ' ' . $extra_info);
  }
}
```

With this, all we have to call in every failures processor is:

```
public function processGenusExistsFailures(array $validation_result) {
  checkValidationStatusArray($validation_result, 'processGenusExistsFailures');
  ...
}

```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
```diff
@@ -1477,4 +1477,202 @@ public function testProcessDuplicateTraitsFailures(array $failures, array $expec
...
+   *           column header for the cell containing the failed value.
+   *           - 'Trait Name': The trait name provided in the file.
+   *           - 'Method Short Name': The method name provided in the file.
+   *           - 'Unit': The unit provided in the file.
+   *   - An array of expectations that we want to find in the resulting rendered
+   *     output. Each array has the following keys:
+   *     - 'expected_exception_message': The message expected by the exception
+   *       being triggered.
+   */
+  public function provideProcessFailuresExceptions() {
```
laceysanderson: For this data provider I think I would provide a scenario for each validator to keep it from getting super large. Then I would test both cases you mention below within the same test method.

So something like the following for the data provider:
```
  /**
   * Data Provider for triggering exceptions in all process failures methods.
   *
   * @return array
   *   Each scenario is an array with the following:
   *   - the name of the validator process function (i.e. processGenusExistsFailures).
   *   - the case returned when the validator passes.
   *   - The failures array that gets passed to the process method. It contains
   *     the following keys:
   *     - [ROW-LEVEL ONLY] The line number that triggered this failed
   *       validation status. This key is not set for non row-level validators.
   *       - 'case': a developer-focused string describing the case checked.
   *       - 'valid': FALSE to indicate that validation failed.
   *       - 'failedItems': array of items that failed consistent with the validator
   *         indicated in this scenario.
   *   - An array of each test case indicating the expected exception information.
   *     Specifically, this array has the keys: 'passed_case' and 'unrecognized_case'
   *     and the value is the message expected by the exception being triggered.
   */
```

Then the test using this can be setup to

1. Call the indicated validator process function with the passed_case indicated by the data provider and check that the exception indicated with 'passed_case' is thrown.
2. Call the indicated validator process function with a random made up case (maybe just use uniqid()) and check that the exception indicated with 'unrecognized_case' is thrown.

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
```diff
@@ -1477,4 +1477,202 @@ public function testProcessDuplicateTraitsFailures(array $failures, array $expec
...
+   *     being triggered.
+   *
+   * @dataProvider provideFaultyValidationStatusArray
+   */
+  public function testCheckValidationStatusArray(array $validation_result, array $expectations) {
+
+    $exception_caught = FALSE;
+    $exception_message = 'NONE';
+    try {
+      $this->importer->checkValidationStatusArray($validation_result);
```
laceysanderson: I also agree with [Reynolds comment above](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139#issuecomment-2622499007). In order to use this approach in your current test, I would simply pass in a made-up validator name here and then below when you check the exception message you add another assertStringContainsString() that confirms this made-up validator name is in the exception message.

This approach also gives you maximum coverage without having to trigger an exception for checkValidationStatusArray for each validator.

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
carolyncaron: This is a great idea, thank you!

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
carolyncaron: My reason for my initial approach is to also provide access to the row numbers once I start working on the row-level process methods. If I change it, I'll have to make row number an optional parameter in checkValidationStatusArray, and I seem to have an aversion to making methods with optional parameters 🙈 I hope there are not additional details we might want to pass in...

But this IS really tempting to keep the process methods less cluttered AND avoid testing an extra exception in each process method. I will try it out with a row-level one next and let you both know how it goes 😃

carolyncaron:
@reynoldtan & @laceysanderson; I made the suggested changes to checkValidationStatusArray() in commit [eba51aa](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/eba51aab240210ad18930113ecfab2163690f119) and updated the test for it in commit [e4145e4](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/e4145e4e79ee6a5910228cfcc69b4b2694dd8905). Whenever you have time, please let me know what you think 😄

reynoldtan:
Here are my comments/suggestions:

1. The loop will repeat the sentence 'Expected to find the key \'$key\' in the validation result array. '

how about:
```
    $missing_keys = array_diff($keys, array_keys($validation_result));
    if ($missing_keys) {
      $errors_found++;
      $error_message = 'Expected to find key(s) ' . implode(', ', $missing_keys) . ' in the validation result array.';
    }

```
2.  Slight difference may be consolidated?

```
from:
 if (array_key_exists('failedItems', $validation_result)) {
      // Check that 'failedItems' contains a value of type array.
      if (!is_array($validation_result['failedItems'])) {
        $errors_found++;
        $error_message .= "Expected the validation result to contain an array for the key 'failedItems', but it did not. ";
      }
      // Check that 'failedItems' is not an empty array.
      elseif ($validation_result['failedItems'] === []) {
        $errors_found++;
        $error_message .= "Expected the validation result to have content for the key 'failedItems', but it was set to an empty array. ";
      }
    }

to:

if (array_key_exists('failedItems', $validation_result)) {
 if (!is_array($validation_result['failedItems']) || $validation_result['failedItems'] === []) {
  $errors_found++;
  $error_message .= 'Expected the validation result to contain an array for key 'failedItems' and must not be empty array.
}




```

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -925,12 +925,7 @@ public function checkValidationStatusArray(array $validation_result, string $val
    */
   public function processGenusExistsFailures(array $validation_result) {
     // Check the format of the validation_result parameter.
```
reynoldtan, line 926: Beautiful 😍

carolyncaron:
In response to @reynoldtan's comments:
> 1. The loop will repeat the sentence 'Expected to find the key '$key' in the validation result array. '
> 
> how about:
> 
> ```
>     $missing_keys = array_diff($keys, array_keys($validation_result));
>     if ($missing_keys) {
>       $errors_found++;
>       $error_message = 'Expected to find key(s) ' . implode(', ', $missing_keys) . ' in the validation result array.';
>     }
> ```
Good catch - I like this suggestion a lot! Implemented in commit [c9e3065](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/c9e30652d412ebf107de46cb1a6ec4056c0e3172)


> 2. Slight difference may be consolidated?
> 
> ```
> from:
>  if (array_key_exists('failedItems', $validation_result)) {
>       // Check that 'failedItems' contains a value of type array.
>       if (!is_array($validation_result['failedItems'])) {
>         $errors_found++;
>         $error_message .= "Expected the validation result to contain an array for the key 'failedItems', but it did not. ";
>       }
>       // Check that 'failedItems' is not an empty array.
>       elseif ($validation_result['failedItems'] === []) {
>         $errors_found++;
>         $error_message .= "Expected the validation result to have content for the key 'failedItems', but it was set to an empty array. ";
>       }
>     }
> 
> to:
> 
> if (array_key_exists('failedItems', $validation_result)) {
>  if (!is_array($validation_result['failedItems']) || $validation_result['failedItems'] === []) {
>   $errors_found++;
>   $error_message .= 'Expected the validation result to contain an array for key 'failedItems' and must not be empty array.
> }
> ```
I think I'd like to keep these 2 separated, since the expected content of 'failedItems' for a passed validation status is an empty array. I'm hoping that by reporting the empty array separate from a different type of value altogether, it will help the developer with debugging. For example, someone might copy and paste a passing validation status when adding a new failed validation case to an existing validator. They might change 'valid' to FALSE but forget to change anything about the value for 'failedItems'.

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
laceysanderson: Instead of making a number of different optional values you could make a single one: array $context that could have both the validator name, row number and any other context for messages we want in the future 🤔 Alternatively, you could make a single required parameter which is string $message_prefix. This way row level validators could include both the validator name and line number but other validators could provide different information.

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
```diff
@@ -1477,4 +1477,575 @@ public function testProcessDuplicateTraitsFailures(array $failures, array $expec
     }
   }
 
+  /**
+   * Data Provider for triggering exceptions in checkValidationStatusArray().
+   *
+   * @return array
+   *   Each scenario is an array with the following:
+   *   - The validation status array returned by a single validator and where
+   *     applicable, a single row of the input file (ie. a row-level validator).
```
laceysanderson: I think this has become outdated. Perhaps it should be

```suggestion
   *   - The validation status array returned by a single validator.
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
```diff
@@ -1477,4 +1477,575 @@ public function testProcessDuplicateTraitsFailures(array $failures, array $expec
...
+
+    return $scenarios;
+  }
+
+  /**
+   * Tests the method that checks the integrity of the validation status array.
+   *
+   * @param array $validation_result
+   *   The validation status array returned by a single validator and where
+   *   applicable, a single row of the input file (ie. a row-level validator).
```
laceysanderson:
```suggestion
   *   The validation status array returned by a single validator.
```

laceysanderson (commented):
Only one small documentation comment --otherwise this looks great :-)

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
```diff
@@ -829,6 +829,77 @@ public function processValidationMessages($failures) {
...
+   */
+  public function checkValidationStatusArray(array $validation_result, string $validator_name, int|null $line_no = NULL) {
+    $error_message = '';
+    $errors_found = 0;
+    // Check for validation status keys: 'case', 'valid', 'failedItems'.
+    $keys = ['case', 'valid', 'failedItems'];
+    $missing_keys = array_diff($keys, array_keys($validation_result));
+    if ($missing_keys) {
+      $errors_found++;
+      $error_message = 'Expected to find key(s) \'' . implode('\', \'', $missing_keys) . '\' in the validation result array. ';
```
reynoldtan: To make it consistent with the rest of the $error_message use of double quotes.
```
$error_message = "Expected to find key(s) '" . implode("', '", $missing_keys) . "' in the validation result array. ";
```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
```diff
@@ -1477,4 +1477,575 @@ public function testProcessDuplicateTraitsFailures(array $failures, array $expec
...
+   *     - 'failedItems': An array of items that failed.
+   *   - An array of expectations that we want to find in the resulting rendered
+   *     output. Each array has the following keys:
+   *     - 'expected_exception': TRUE or FALSE if an exception is expected to
+   *       occur for this scenario.
+   *     - 'expected_errors': The number of expected problems with the array.
+   *     - 'expected_details': The details in the message expected to be in the
+   *        exception being triggered.
+   */
+  public function provideFaultyValidationStatusArray() {
```
reynoldtan, line 1498: Can it also test if a key is missing from the array? ie:

```
// case key is missing.
[
  'valid' => FALSE,
  'failed_Items' => [
    'item' => 'failed'
  ],
]

```

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
carolyncaron: Good catch! Addressed in commit [41d238b](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/41d238bfd13b84e9ee9aa2586074356d3e886f2b)

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
carolyncaron: Addressed in commit [41d238b](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/41d238bfd13b84e9ee9aa2586074356d3e886f2b)

trpcultivate_phenotypes/src/Plugin/TripalImporter/TripalCultivatePhenotypesTraitsImporter.php
carolyncaron: Thanks! Addressed in commit [cbdff65](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/cbdff6575e7edddcf99cf25608687c28458801d3)

trpcultivate_phenotypes/tests/src/Kernel/TripalImporter/TraitImporterProcessValidationTest.php
carolyncaron: I ended up applying this suggestion to scenario #0 in commit [cbdff65](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/139/commits/cbdff6575e7edddcf99cf25608687c28458801d3), as I think it is more clear than using a mispelled key! Note that I kept  'failed_items' ('failedItems' is the expected key so it is technically missing) as I think this could be a common mispelling while developing validators.

laceysanderson (approved):
This looks done to me ✅ ❤️ 

Note: We now have 100% coverage of all the process sub-methods with the TraitImporterProcessValidationTest test.
Screenshot 2025-02-11 at 3 33 24 PM

Issue #152, opened 2025-03-31
G3.152 Create phenotypes backup page


reynoldtan:
### Branch

g3.153-BackupPage

### Groups

Group 3 - Search | Browse | Tools

### Describe

Setup a Drupal configuration entity to allow user to backup phenotypes data collection file. The idea is that each entity/page/instance is a single backup. By using the configuration entity API we get a lot of features for free. See [Drupal Creating a configuration entity type](https://www.drupal.org/docs/drupal-apis/configuration-api/creating-a-configuration-entity-type). **Note: There is a drush generate command to create these files for you -do not do it manually.**

### Design

Entity Type name: Phenotypic Data Backup (phenodata_backup)

Interface:
1. A basic form interface that with the following fields:

- A data file to backup. This will be a drupal file field.
- Project - to specify to which project the data file relates to. This will be an entity relationship field that references Research Experiment Tripal Content Types.
- Comment - user notes and comments. This will be a Drupal Long Text Area.

![Image](https://github.com/user-attachments/assets/49c81be0-e402-41bb-b89e-f0e2fef3933e)

2. Save the file in: `public://phenotype-backups` and rename based on projectid and date.
3. Create a view listing of the files. Sorted by date + project name (Allow filter for project name) The project filter will use the entity relationship handler which I hope supports select out of the box 🙏 

![Image](https://github.com/user-attachments/assets/5cd47a12-c1c7-4958-8a79-36e4d6283c81)

Backend:
```
Configuration entity schema:
trpcultivate_phenotypes.data_file_backup:
  type: config_entity
  label: 'Data Collection File Backup'
  mapping:
    file_id
      type: string
      label: 'File object'
    backup_date:
      type: date
      label: 'Date file was uploaded'
    project_id:
      type: int
      label: 'The project id of the project the data file is specific to'
    comments:
      type: text
      label: 'Notes or comments about the file'
   user_id:
      type: int
      label: 'The user id number'
   ```

### NOTES:

1. Discuss how to reference validation result performed to the file. Perhaps we add a field to the config entity that contains the validation results? This may need to be a custom Drupal field 🤔 Let's worry about this once we get the rest setup.
2. Location to save all backed up files. Currently we have it as `public://phenotype-backups` but we may want to take a small amount of time to see how hard it is to make these private?
3. Given recent difficulty in setting up a filter in view using select box, what should the filter field be? Hopefully using an entity reference will provide a better select filter without needing hard-coding.

[...]

PR #153, opened 2025-04-04
G3.153 Setup backup page

reynoldtan:
**Issue #152 Create Phenotypes Backup Page**
Screenshot 2025-04-09 at 2 24 09 PM

## Motivation

This PR will setup a phenotypes data file backup page. The setup includes a schema, actions to add, update, delete and list user data file backups.

## What does this PR do?

This task used drush generate to setup codebase and created the following file/class. Moreover, the command also
modified existing module files such as permissions and routing.

1. Modified schema/trpcultivate_phenotypes.schema.yml
2. Modified trpcultivate_phenotypes.routing.yml
3. Created trpcultivate_phenotypes.permissions.yml
4. Created src/Entity/PhenoBackupForm.php
5. Created src/Form/PhenodataBackupForm.php
6. Created src/PhenodataBackupInterface.php
7. Created src/PhenoBackupListBuilder.php

### Automated Testing

PhenodataBackupListTest
- Test access response code of the listing page
- Test that the list contains the backup details

### Manual Testing

1. In a fully working clone create a Research Experiment content type and an authenticated user.
2. Create roles view all and view own. View all permission will enable the user to view the entire list of backups available whereas view_own restricts backups owned/created by the user logged in. During the test, switch the roles to test this result.

Screenshot 2025-04-09 at 2 28 37 PM

4. In the permission page, grant Phenotypes Data Backup permission to either view_all or view_own roles.
5. Access my site/admin/structure/phenodata-backup to view the listing. This page will result in Access Denied if user does not have proper permissions.
6. Upload a backup. The upload backup form is in sync with the Phenoshare Importer in terms of the files extensions supported. Access my clone/admin/tripal/loaders/trpcultivate-phenotypes-share-importer and click on download a template file to use a test backup file. Alternatively, create a file with a tsv extension.
7. With project created and a test file ready, upload a backup. Provide a note or comment.
8. Upload as many desired.
9. Update a backup file by clicking the update button.
10. Filter the list by project name in the filter form panel.

Things to test:
1. The order of the backup is descending order by date - the last created is the first.
2. A user cannot update a data file of another user (in view_all permission).
4. The data file is renamed to Project Id_User id_Year-Month-Day h:m:s . extension
5. With view_own permission there is no column Created by.
6. Comment is optional and the character -- is placed.
7. The test is basic and needs more coverage.

Credits/reference:

https://github.com/drupal/drupal/blob/711e9ad4264a66f2283c7d26b5548e1f9000766d/core/modules/config_translation/src/Controller/ConfigTranslationEntityListBuilder.php

https://github.com/drupal/drupal/blob/711e9ad4264a66f2283c7d26b5548e1f9000766d/core/lib/Drupal/Core/Config/Entity/ConfigEntityListBuilder.php

https://github.com/drupal/drupal/blob/711e9ad4264a66f2283c7d26b5548e1f9000766d/core/modules/views_ui/src/ViewListBuilder.php

### DIFF
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -40,10 +40,18 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *     "id" = "id",
  *     "uuid" = "uuid",
  *     "file_id" = "file_id",
+ *     "project_id" = "project_id",
+ *     "comments" = "comments",
+ *     "backup_date" = "backup_date",
+ *     "user_id" = "user_id",
  *   },
  *   config_export = {
  *     "id",
  *     "file_id",
+ *     "project_id",
+ *     "comments",
+ *     "backup_date",
+ *     "user_id",
  *   },
  * )
  */
@@ -58,5 +66,25 @@ final class PhenodataBackup extends ConfigEntityBase implements PhenodataBackupI
    * The file object file id of the data file.
    */
   protected string $file_id;
+
+  /**
+   * The project id of the project the data file is specific to.
+   */
+  protected string $project_id;
+
+  /**
+   * User notes or comments about the data file.
+   */
+  protected string $comments;
+
+  /**
+   * The data the data file was uploaded.
+   */
+  protected string $backup_date;
+
+  /**
+   * The user id of the user who uploaded the data file.
+   */
+  protected string $user_id;
 
 }
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -24,6 +24,7 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *     "list_builder" = "Drupal\trpcultivate_phenotypes\PhenodataBackupListBuilder",
  *     "form" = {
  *       "add" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
+ *       "edit" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
  *       "delete" = "Drupal\Core\Entity\EntityDeleteForm",
  *     },
  *   },
@@ -32,6 +33,7 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *   links = {
  *     "collection" = "/admin/structure/phenodata-backup",
  *     "add-form" = "/admin/structure/phenodata-backup/add",
+ *     "edit-form" = "/admin/structure/phenodata-backup/{phenodata_backup}",
  *     "delete-form" = "/admin/structure/phenodata-backup/{phenodata_backup}/delete",
  *   },
  *   entity_keys = {
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -1,0 +1,62 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\trpcultivate_phenotypes\Entity;
+
+use Drupal\Core\Config\Entity\ConfigEntityBase;
+use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
+
+/**
+ * Defines the phenotypic data backup entity type.
+ *
+ * @ConfigEntityType(
+ *   id = "phenodata_backup",
+ *   label = @Translation("Phenotypic Data Backup"),
+ *   label_collection = @Translation("Phenotypic Data Backups"),
+ *   label_singular = @Translation("phenotypic data backup"),
+ *   label_plural = @Translation("phenotypic data backups"),
+ *   label_count = @PluralTranslation(
+ *     singular = "@count phenotypic data backup",
+ *     plural = "@count phenotypic data backups",
+ *   ),
+ *   handlers = {
+ *     "list_builder" = "Drupal\trpcultivate_phenotypes\PhenodataBackupListBuilder",
+ *     "form" = {
+ *       "add" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
+ *       "edit" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
+ *       "delete" = "Drupal\Core\Entity\EntityDeleteForm",
+ *     },
+ *   },
+ *   config_prefix = "phenodata_backup",
+ *   admin_permission = "administer phenodata_backup",
+ *   links = {
+ *     "collection" = "/admin/structure/phenodata-backup",
+ *     "add-form" = "/admin/structure/phenodata-backup/add",
+ *     "edit-form" = "/admin/structure/phenodata-backup/{phenodata_backup}",
+ *     "delete-form" = "/admin/structure/phenodata-backup/{phenodata_backup}/delete",
+ *   },
+ *   entity_keys = {
+ *     "id" = "id",
+ *     "uuid" = "uuid",
+ *     "file_id" = "file_id",
+ *   },
+ *   config_export = {
+ *     "id",
+ *     "file_id",
+ *   },
+ * )
+ */
+final class PhenodataBackup extends ConfigEntityBase implements PhenodataBackupInterface {
+
+  /**
+   * The configuation entity ID number.
+   */
+  protected string $id;
+
+  /**
+   * The file object file id of the data file.
+   */
+  protected string $file_id;
+
+}
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -5,7 +5,6 @@ declare(strict_types=1);
 namespace Drupal\trpcultivate_phenotypes\Entity;
 
 use Drupal\Core\Config\Entity\ConfigEntityBase;
-use Drupal\trpcultivate_phenotypes\Entity\PhenodataBackupInterface;
 
 /**
  * Defines the phenotypic data backup entity type.
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -24,7 +24,6 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *     "list_builder" = "Drupal\trpcultivate_phenotypes\PhenodataBackupListBuilder",
  *     "form" = {
  *       "add" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
- *       "edit" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
  *       "delete" = "Drupal\Core\Entity\EntityDeleteForm",
  *     },
  *   },
@@ -33,7 +32,6 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *   links = {
  *     "collection" = "/admin/structure/phenodata-backup",
  *     "add-form" = "/admin/structure/phenodata-backup/add",
- *     "edit-form" = "/admin/structure/phenodata-backup/{phenodata_backup}",
  *     "delete-form" = "/admin/structure/phenodata-backup/{phenodata_backup}/delete",
  *   },
  *   entity_keys = {
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 namespace Drupal\trpcultivate_phenotypes\Entity;
 
 use Drupal\Core\Config\Entity\ConfigEntityBase;
-use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
+use Drupal\trpcultivate_phenotypes\Entity\PhenodataBackupInterface;
 
 /**
  * Defines the phenotypic data backup entity type.
@@ -21,7 +21,7 @@ use Drupal\trpcultivate_phenotypes\PhenodataBackupInterface;
  *     plural = "@count phenotypic data backups",
  *   ),
  *   handlers = {
- *     "list_builder" = "Drupal\trpcultivate_phenotypes\PhenodataBackupListBuilder",
+ *     "list_builder" = "Drupal\trpcultivate_phenotypes\ListBuilder\PhenodataBackupListBuilder",
  *     "form" = {
  *       "add" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
  *       "edit" = "Drupal\trpcultivate_phenotypes\Form\PhenodataBackupForm",
--- a/trpcultivate_phenotypes/src/Entity/PhenodataBackupInterface.php
+++ b/trpcultivate_phenotypes/src/Entity/PhenodataBackupInterface.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Drupal\trpcultivate_phenotypes;
+namespace Drupal\trpcultivate_phenotypes\Entity;
 
 use Drupal\Core\Config\Entity\ConfigEntityInterface;
 
--- a/trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
+++ b/trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
@@ -4,14 +4,103 @@ declare(strict_types=1);
 
 namespace Drupal\trpcultivate_phenotypes\Form;
 
+use Drupal\Core\Config\ConfigFactoryInterface;
 use Drupal\Core\Entity\EntityForm;
+use Drupal\Core\Entity\EntityTypeManagerInterface;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\file\Entity\File;
+use Drupal\Core\Session\AccountInterface;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+use Drupal\tripal_chado\Controller\ChadoProjectAutocompleteController;
 
 /**
  * Phenotypic Data Backup form.
  */
 final class PhenodataBackupForm extends EntityForm {
+
+  /**
+   * Entity type manager service.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected EntityTypeManagerInterface $service_EntityTypeManager;
+
+  /**
+   * Configuration.
+   *
+   * @var \Drupal\Core\Config\ConfigFactoryInteface
+   */
+  protected ConfigFactoryInterface $service_ConfigFactory;
+
+  /**
+   * Users.
+   *
+   * @var \Drupal\Core\Session\AccountInterface
+   */
+  protected AccountInterface $user;
+
+  /**
+   * Constructor.
+   *
+   * @param \Drupal\Core\Entity\EntityTypeManagerInterface $entity_type_manager
+   *   Entity Type manager service.
+   * @param \Drupal\Core\Config\ConfigFactoryInterface $config_factory
+   *   Configuration factory service.
+   * @param \Drupal\Core\Session\AccountInterface $user
+   *   Drupal users.
+   */
+  public function __construct(
+    EntityTypeManagerInterface $entity_type_manager,
+    ConfigFactoryInterface $config_factory,
+    AccountInterface $user,
+  ) {
+
+    $this->service_EntityTypeManager = $entity_type_manager;
+    $this->service_ConfigFactory = $config_factory;
+    $this->user = $user;
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public static function create(ContainerInterface $container) {
+
+    return new static(
+      $container->get('entity_type.manager'),
+      $container->get('config.factory'),
+      $container->get('current_user'),
+    );
+  }
+
+  /**
+   * {@inheritDoc}
+   */
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+
+    $project_name = $form_state->getValue('project_name');
+
+    // Pull the project id from the project value.
+    preg_match('/\((\d+)\)$/', $project_name, $matches);
+
+    $invalid_project = FALSE;
+
+    if (!isset($matches[1])) {
+      $invalid_project = TRUE;
+    }
+    else {
+      $project_id = (int) trim($matches[1]);
+      $project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
+
+      if (empty($project_name)) {
+        $invalid_project = TRUE;
+      }
+    }
+
+    if ($invalid_project) {
+      $form_state->setErrorByName('project_name', 'The project is not recognized. Please select a project and try again.');
+    }
+
+    return $form;
+  }
 
   /**
    * {@inheritdoc}
@@ -27,11 +116,16 @@ final class PhenodataBackupForm extends EntityForm {
       '#value' => uniqid(),
     ];
 
+    $config_backup_dir = $this->service_ConfigFactory
+      ->get('trpcultivate_phenotypes.settings')
+      ->get('trpcultivate.phenotypes.directory.data_backup');
+    $backup_dir = $config_backup_dir ?? 'public:://phenotype-backups/';
+
     $form['backup_file'] = [
       '#type' => 'managed_file',
       '#title' => 'Data File',
       '#description' => $this->t('Select data file to backup. Only tsv and txt file extensions are allowed.'),
-      '#upload_location' => 'public://phenotype-backups/',
+      '#upload_location' => $backup_dir,
       '#multiple' => FALSE,
       '#required' => TRUE,
       '#upload_validators' => [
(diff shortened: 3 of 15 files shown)


carolyncaron (changes requested):
I tested this as an administrator, and it worked flawlessly! 👏  I checked:
- [x] Backups are sorted by newest first
- [x] The downloadable backup is renamed in the specified format
- [x] Comments were optional
- [x] Comments were updated if I changed them using the "Update" button
- [x] The filter by Project or Experiment name worked as expected

There are 2 issues I found (or I may just need clarification):
1. I need help to properly test the user role functionality. 😅 I did create a new user and tried to assign one of the 2 view roles, but after that, I'm confused how to access the backup form without being an administrator (I get Access Denied trying to go to `site/admin/structure/phenodata-backup`). I assume we'll want backups to be made available to non-administrative users?

2. On the form to add a phenotypic data backup, the field title for experiment is "Project/Experiment Name". So, I created a Project content type (**not** Research Experiment) and I expected to be able to use it in the form, but I couldn't. Do we want to allow backups associated with **Project** or exclusively with **Research Experiment**? If that's the case, I think we need to remove all mention of the word "Project" to avoid confusion.

laceysanderson (changes requested):
## Setup

I created two users (Kirstin + Alex) and gave Kirstin the content editor role and Alex no role. Then I assigned the view all backups permission to the content editor role (ie.e. Kirstin) and the view own backup to the authenticated user (i.e. Alex).

Screenshot 2025-04-10 at 4 01 59 PM
Screenshot 2025-04-10 at 4 02 48 PM

## Manual Testing

✅ I was able to confirm that both users were able to see the backups page and create backups. 
✅ Alex could only see his own backups as intended.
Screenshot 2025-04-10 at 3 44 12 PM

✅ Kirstin could see both hers and Alexs backups as intended. It is also clear who each backup belongs to.
Screenshot 2025-04-10 at 3 44 22 PM

✅ Both were able to edit their own backups but only the comment as intended.
✅ The backup files were renamed according to the pattern discused and saved in public::// as designed although I'm reconsidering the format now that I see it 🙈 
Screenshot 2025-04-10 at 3 51 40 PM

✅ The select was restricted to Research Experiments as intended however, the naming in the UI is incorrect. It should say "Research Experiment" in all places (list, form, filter) instead of "Project/Experiment".
❌ The list has a link to the project to view it... however, clicking on it results in "Access Denied". This is making the assumption that the user has access to view the experiment pages.
✅ Updating the comment works as expected.
✅ Downloading the backup works as expected.

## Tasks

- [ ] Change the field name to "Research Experiment" in all places (list, form, filter) instead of "Project/Experiment"
- [ ] Change filename format to PhenotypeBackup-EXPERIMENT-USER-DATETIME.tsv (e.g. `PhenotypeBackup_2_3_2025Apr10_21_41_44.tsv`)

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,327 @@
...
+      '#value' => uniqid(),
+    ];
+
+    if ($entity->isNew()) {
+      // Synchronize the file types defined by the importer that will be the
+      // main source of future backup data collection file.
+      $importer_plugin_definition = $this->tripal_importer
+        ->getDefinition('trpcultivate-phenotypes-share-importer');
+      $importer_file_extension = implode(' ', $importer_plugin_definition['file_types']);
+
```
laceysanderson: This creates a circular dependency between trpcultivate_phenotyes and trpcultivate_share. I see why you took this approach but in this case we are better off to hardcode the file types.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,327 @@
...
+    }
+    else {
+      // Ensure that user can only modify data file that they own.
+      if ($entity->get('user_id') != $this->user->id()) {
+        $form['not_my_backup'] = [
+          '#markup' => 'This data file backup does not belong to this account.',
+        ];
+
+        return $form;
+      }
```
laceysanderson: This is never reached as drupal route permissions should control who accesses the edit form.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,327 @@
...
+        '#description' => $this->t('Select data file to backup. Only [@ext] file extensions are allowed.', ['@ext' => $importer_file_extension]),
+        '#upload_location' => $backup_dir,
+        '#multiple' => FALSE,
+        '#required' => TRUE,
+        '#upload_validators' => [
+          'FileExtension' => [
+            'extensions' => $importer_file_extension,
+          ],
+        ],
+      ];
```
laceysanderson: This should be reused rather then having the if isNew() and essentially two forms. The following will allow this form to show the file uploaded on edit but for it to be disabled (cannot remove or change).

```suggestion
      $fids = $entity->get('file_id');
      $fids = (is_string($fids)) ? [$fids] : $fids;
      $form['backup_file'] = [
        '#type' => 'managed_file',
        '#title' => 'Data File',
        '#description' => $this->t('Select data file to backup. Only [@ext] file extensions are allowed.', ['@ext' => $importer_file_extension]),
        '#upload_location' => $backup_dir,
        '#multiple' => FALSE,
        '#required' => TRUE,
        '#upload_validators' => [
          'FileExtension' => [
            'extensions' => $importer_file_extension,
          ],
        ],
        '#value' => [
          'fids' => $fids,
        ],
        '#disabled' => (empty($fids)) ? FALSE : TRUE,
      ];
```

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,327 @@
...
+        '#autocomplete_route_name' => 'tripal_chado.generic_autocomplete',
+        '#autocomplete_route_parameters' => [
+          'type_id' => $type_id,
+          'match_limit' => 5,
+          'base_table' => 'project',
+          'column_name' => 'name',
+          'type_column' => 'type_id',
+          'property_table' => 'project',
+        ],
+      ];
```
laceysanderson: Same as with the file.


```suggestion
      $project_id = (int) $entity->get('project_id');
      $project_name = ($project_id) ? ChadoProjectAutocompleteController::getProjectName($project_id): '';
      $form['project_name'] = [
        '#type' => 'textfield',
        '#title' => 'Project/Experiment Name',
        '#description' => $this->t('Select the project name the data file is specific to.'),
        '#description_display' => 'after',
        '#required' => TRUE,
        '#attributes' => ['placeholder' => 'Project/Experiment Name'],
        '#autocomplete_route_name' => 'tripal_chado.generic_autocomplete',
        '#autocomplete_route_parameters' => [
          'type_id' => $type_id,
          'match_limit' => 5,
          'base_table' => 'project',
          'column_name' => 'name',
          'type_column' => 'type_id',
          'property_table' => 'project',
        ],
        '#value' => ($project_id) ? "$project_name ($project_id)" : NULL,
        '#disabled' => ($project_id) ? TRUE : FALSE,
      ];
```

trpcultivate_phenotypes/src/PhenodataBackupListBuilder.php
```diff
@@ -0,0 +1,339 @@
...
+  public function buildRow(EntityInterface $entity): array {
+
+    $values = [];
+
+    $key = 'project_id';
+    $project_id = (int) $entity->get($key);
+    $project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
+    $values[$key]['data'] = [
+      '#markup' => '<a target="_blank" href="/experiment/' . str_replace(' ', '-', $project_name) . '">' . $project_name . '</a>',
+    ];
```
laceysanderson: This makes an assumption on the url for the experiment which may not always hold true :grimaces. Instead we actually need to get the entity id for the project and check that the user has permission to it before showing it as a link. Additionally, you should use a link render array here rather then `#markup`.

trpcultivate_phenotypes/trpcultivate_phenotypes.links.action.yml
```diff
@@ -0,0 +1,5 @@
+entity.phenodata_backup.add_form:
+  route_name: 'entity.phenodata_backup.add_form'
+  title: 'Add phenotypic data backup'
```
laceysanderson:
```suggestion
  title: 'Backup Phenotypic Data'
```

trpcultivate_phenotypes/trpcultivate_phenotypes.links.menu.yml
```diff
@@ -9,4 +9,9 @@ trpcultivate_phenotypes.rrules:
 
 # trpcultivate_phenotypes.watermark:  
 # trpcultivate_phenotypes.ontology:
-  
\ No newline at end of file
+  
+entity.phenodata_backup.overview:
+  title: Phenotypic Data Backups
+  parent: system.admin_structure
+  description: 'List of phenotypic data backups to extend site functionality.'
```
laceysanderson:
```suggestion
  description: 'The collection of phenotypic data backups available in this site that you have permission to access.'
```

trpcultivate_phenotypes/trpcultivate_phenotypes.routing.yml
```diff
@@ -34,3 +34,36 @@ trpcultivate_phenotypes.settings_ontology:
     _form: '\Drupal\trpcultivate_phenotypes\Form\TripalCultivatePhenotypesOntologySettingsForm'
   requirements:
     _permission: 'administer tripal'
+
+# Phenodata Backup Config Entity Routes:
+entity.phenodata_backup.collection:
+  path: '/admin/structure/phenodata-backup'
+  defaults:
+    _entity_list: 'phenodata_backup'
+    _title: 'Phenotypic Data Backup configuration'
```
laceysanderson:
```suggestion
    _title: 'Phenotypic Data Backups
```

laceysanderson (dismissed):
## Code Review

- [x] Move `PhenodataBackupInterface.php` into `src/Entity` and change the namespace accordingly.
- [x] Move `PhenodataBackupListBuilder.php` into a new directory `src/ListBuilders` and change the namespace accordingly.
- [x] Form should reuse the file and project form elements but disable them on edit (see suggestions).
- [x] Update the experiment link to use the entity id.

trpcultivate_phenotypes/src/PhenodataBackupListBuilder.php
laceysanderson: This has been moved to #154. Instead for this PR lets just use the project name and not make it a link. :-)

trpcultivate_phenotypes/trpcultivate_phenotypes.routing.yml
reynoldtan: This suggestion did not cause any Problems and is committed.

trpcultivate_phenotypes/trpcultivate_phenotypes.links.menu.yml
reynoldtan: The suggestion did not cause any Problem is committed.

trpcultivate_phenotypes/trpcultivate_phenotypes.links.action.yml
reynoldtan: The suggestion did not cause Problem and is committed.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
reynoldtan: I have modified this block to just a reminder to a user with view_all permission that the file which is about to be updated belongs to another user. Just like we talked about. Thanks to @carolyncaron for this brilliant idea. However, I discovered a potential bug where a user with view_own permission can modify someone's backups. This is by hacking the url. I added a condition to prevent this.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
reynoldtan: I decoupled phenotype share and extension can be set at the beginning line of the class.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
reynoldtan: Reused the field, this is awesome.. I modified the value key a bit since using this setup the file field seems to upload the file into the void

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
reynoldtan: implemented

trpcultivate_phenotypes/src/PhenodataBackupListBuilder.php
reynoldtan: Thanks for setting up issue. I implemented the additional part and made all markup use a render array.

carolyncaron (commented):
I manually tested this again and I wanted to report that I was able to create an authorized user with the "View Own" permission and confirmed that this user can only see their own uploads. Whereas a user with the "View All" permission does, in fact, see all the backups! 🎉  I have not been able to break this great functionality 😁 

I am also happy to see the removal of "Project" from most places- however I noticed there is one spot that was missed:
Screenshot 2025-04-22 at 11 47 06 AM

trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
```diff
@@ -0,0 +1,90 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\trpcultivate_phenotypes\Entity;
+
+use Drupal\Core\Config\Entity\ConfigEntityBase;
+use Drupal\trpcultivate_phenotypes\Entity\PhenodataBackupInterface;
```
carolyncaron: I see the warning "Unused use statement" in VS code

trpcultivate_phenotypes/src/Entity/PhenodataBackupInterface.php
```diff
@@ -0,0 +1,14 @@
...
+
+namespace Drupal\trpcultivate_phenotypes\Entity;
+
+use Drupal\Core\Config\Entity\ConfigEntityInterface;
+
+/**
+ * Provides an interface defining a phenotypic data backup entity type.
+ */
+interface PhenodataBackupInterface extends ConfigEntityInterface {
+
```
carolyncaron, line 13: I can't help but question why this Interface is left empty - I think either it should be removed, or the properties in PhenodataBackup.php could also be added here?

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,321 @@
...
+    ];
+
+    $form['comments'] = [
+      '#type' => 'textarea',
+      '#title' => 'Notes/Comments',
+      '#description' => $this->t('Notes or comments about the data file.'),
+      '#default_value' => $entity->get('comments'),
+      '#rows' => '5',
+      '#resizable' => FALSE,
+      '#maxlength' => 200,
```
carolyncaron: I think 200 chars is too limited. It also looks misleading to the user because the text box is much bigger than the amount of text that can fit in it. I would remove this line completely if possible.
```suggestion
```

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,321 @@
...
+
+    $project_id = (int) $entity->get('project_id');
+    $project_name = ($project_id) ? ChadoProjectAutocompleteController::getProjectName($project_id) : '';
+    $form['project_name'] = [
+      '#type' => 'textfield',
+      '#title' => 'Research Experiment',
+      '#description' => $this->t('Select the project name the data file is specific to.'),
+      '#description_display' => 'after',
+      '#required' => TRUE,
+      '#attributes' => ['placeholder' => 'Project/Experiment Name'],
```
carolyncaron:
```suggestion
      '#attributes' => ['placeholder' => 'Experiment Name'],
```

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,321 @@
...
+      // 0 value will suggest all projects.
+      $type_id = $entity_type_term_internalid;
+    }
+
+    $project_id = (int) $entity->get('project_id');
+    $project_name = ($project_id) ? ChadoProjectAutocompleteController::getProjectName($project_id) : '';
+    $form['project_name'] = [
+      '#type' => 'textfield',
+      '#title' => 'Research Experiment',
+      '#description' => $this->t('Select the project name the data file is specific to.'),
```
carolyncaron:
```suggestion
      '#description' => $this->t('Select the Research Experiment that this data file was generated for.'),
```

trpcultivate_phenotypes/src/ListBuilder/PhenodataBackupListBuilder.php
```diff
@@ -0,0 +1,348 @@
...
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+
+    $project_id = (int) $form_state->getValue('project_id');
+
+    if ($project_id > 0) {
+      $project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
+
+      if (empty($project_name)) {
+        $form_state->setErrorByName('project_id', $this->t(
+          'The @project is not recognized. Please select an option and try again.',
```
carolyncaron:
```suggestion
          'The @project is not recognized. Start by typing slowly, select the correct one from the dropdown and try saving again.',
```

carolyncaron (changes requested):
Here's some suggestions/comments so far, I'll revisit the tests a bit later.

trpcultivate_phenotypes/src/Entity/PhenodataBackupInterface.php
laceysanderson: I think the properties should be added here :-) Since Reynold is on holidays, I will do that.

laceysanderson: Apparently VS code does not like properties in interfaces :sweat: Since there are no methods in the class implementing this interface, I guess it should just remain empty :shrugs: Technically it is not needed but I do like to keep them in place to ensure that we indicate the methods we expect to exist in all versions of PhenodataBackup.

trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
laceysanderson: Interesting.  I am not seeing a warning for this which makes sense because it is used below on line 58. :thinking: I wonder why you are seeing it.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
laceysanderson: I agree, I've removed the limit in commit 5405e7173ddffcc8fffa9b9be5e47c45fc8ea41b and confirmed that it can now save very long text with no issues :-) It does expand the listing if they go overboard but I think that's fine.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,320 @@
...
+/**
+ * Phenotypic Data Backup form.
+ */
+final class PhenodataBackupForm extends EntityForm {
+
+  /**
+   * File types supported.
+   */
+  private const array BACKUP_FILE_TYPES = [
+    'tsv',
```
laceysanderson, line 28: The share importer can only handle tsv so this makes sense... but I wonder if we should allow txt and xlsx formats here in recognition that while the data is being generated it may be in these other formats?

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
laceysanderson: > However, I discovered a potential bug where a user with view_own permission can modify someone's backups. This is by hacking the url. I added a condition to prevent this.

The condition to prevent this ended up being essentially the same behaviour I commented on here. 
Screenshot 2025-04-23 at 2 44 47 PM

They should instead get Access Denied in this case. This change record indicates how we can trigger access denied within a page: https://www.drupal.org/node/1616360. I did this in commit 63d0c53fe9731101983ef348fb8abe0862375d21.
Screenshot 2025-04-23 at 2 45 51 PM

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
laceysanderson: After a discussion with Carolyn we decided to move forward with expanding the formats supported. This was done in 2f389f84646eb084f9956dc6bd9b545a868f0b92.

laceysanderson:
Automated Testing completed:

Screenshot 2025-04-25 at 2 40 37 PM

Screenshot 2025-04-25 at 2 41 38 PM

trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
laceysanderson: I just removed the use statement and am now seeing the warning. I think this is because the interface and this class are in the same namespace so it can find it without the use statement.

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
carolyncaron: I tested it as well, looks great to me!

trpcultivate_phenotypes/src/Entity/PhenodataBackup.php
carolyncaron: Confirmed that the use statement is now removed 👍  We realized that it is unnecessary since `PhenodataBackupInterface` is in the same namespace.

trpcultivate_phenotypes/src/ListBuilder/PhenodataBackupListBuilder.php
```diff
@@ -0,0 +1,348 @@
...
+  public function validateForm(array &$form, FormStateInterface $form_state) {
+
+    $project_id = (int) $form_state->getValue('project_id');
+
+    if ($project_id > 0) {
+      $project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
+
+      if (empty($project_name)) {
+        $form_state->setErrorByName('project_id', $this->t(
+          'The @project is not recognized. Start by typing slowly, select the correct one from the dropdown and try saving again.',
```
carolyncaron: While mannually testing, I've realized that I can't trigger this message. I should have made my code suggestion in `PhenodataBackupForm:validateForm()` instead of here.

I think this message should be updated to be accurate in the event that it does get triggered. Perhaps something like:
```suggestion
          'The @project is not recognized or you do not have permission to see backups for it.',
```

trpcultivate_phenotypes/src/Form/PhenodataBackupForm.php
```diff
@@ -0,0 +1,335 @@
...
+        $project_id = (int) trim($matches[1]);
+        $project_name = ChadoProjectAutocompleteController::getProjectName($project_id);
+
+        if (empty($project_name)) {
+          $invalid_project = TRUE;
+        }
+      }
+
+      if ($invalid_project) {
+        $form_state->setErrorByName('project_name', 'The Research Experiment name is not recognized. Please select an option and try again.');
```
carolyncaron:
```suggestion
        $form_state->setErrorByName('project_name', 'The Research Experiment name is not recognized. Start by typing slowly, select the correct one from the dropdown and try saving again.');
```

carolyncaron (changes requested):
I manually tested with 4 different users and the PR reflected the expected functionality for each:

- [x] Administer pheno backups permission
- [x] View All permission
- [x] View Own permission
- [x] No phenodata backup permissions

I also love that more file extensions are being supported, and my tests for this worked flawlessly 🎉 

However, while I was trying to trigger the error message that I previously made a code suggestion for (see my new suggestions below), I found this behaviour when trying to access a filtered phenodata backup listing **via URL** (ex: http://localhost:50950/admin/structure/phenodata-backup?project_id=999)
Screenshot 2025-04-29 at 1 22 48 PM
The 2 scenarios that trigger this are:
- Trying to filter for a project_id that doesn't exist
- Trying to access an existing project_id that the user does not have existing backups for and they only have the "View Own" permission.

**I don't think this is dire enough to address in this PR**, as it's unlikely to trigger and it ultimately does not show any backups that the user doesn't have access to. But it's something to keep in mind since I could see a user potentially sharing the URL to their own backups with another user in the future. 

Lastly, I have a suggestion for the upload form. I think we should include a warning to remind the user to carefully select their Research Experiment, since it cannot be changed later. If the user does not yet have a Research Experiment created before they want to backup data, the message can also remind them to reach out to the data curator or administrator to setup their RE page first. While the user still has the power to select the wrong RE, this prevents them from feigning ignorance 😅 

Final Note: @laceysanderson and I talked about this briefly and we should probably sort the RE listing in both the backup form and the listing to make it easier to find the correct one.

laceysanderson (commented):
> Lastly, I have a suggestion for the upload form. I think we should include a warning to remind the user to carefully select their Research Experiment, since it cannot be changed later. If the user does not yet have a Research Experiment created before they want to backup data, the message can also remind them to reach out to the data curator or administrator to setup their RE page first. While the user still has the power to select the wrong RE, this prevents them from feigning ignorance 😅

I ended up changing the experiment form element description. It now looks like the first screenshot when creating a backup and the second screenshot when updating a backup.
Screenshot 2025-04-29 at 5 29 08 PM
Screenshot 2025-04-29 at 5 28 53 PM

> Final Note: @laceysanderson and I talked about this briefly and we should probably sort the RE listing in both the backup form and the listing to make it easier to find the correct one.

Apparently the autocomplete is already sorted (in the create/update backup form; see first screenshot). I have now added a sort to the backup listing experiment filter (second screenshot).
Screenshot 2025-04-29 at 3 39 54 PM
Screenshot 2025-04-29 at 3 44 55 PM

I think that addresses the things meant to be fixed in this PR?

carolyncaron (approved):
Those updates look great @laceysanderson! I do think this looks ready to merge 😁

Issue #158, opened 2025-05-07
G4.158 Migrate the process failures method for the ProjectGenusMatch validator


carolyncaron:
### Branch

g4.158-migrateProjectGenusMatchProcessor

### Groups

Group 4 - API | Services | Plugins

### Describe

As part of our process to simplify our validator plugins (see the following issue in TripalCultivate: https://github.com/TripalCultivate/TripalCultivate/issues/59), we want to move our `process[VALIDATOR_NAME]failures()` methods to the validator class itself. The first validator we want to try this with is `ProjectGenusMatch`, since it will be easy to test manually once PR #150 has been merged.

### Design

We want to move this code:
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/4e28652a18937f6a9e322c4d631c03afcf38e217/trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoShareImporter.php#L1057C3-L1127C1

Into this class: `trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php`

laceysanderson:
I have added some docs for the new process* methods we've been discussing (one of which will be implemented in this issue) in [this PR](https://github.com/TripalCultivate/docs/pull/6). Once the PR is finished and merged these pages will be accessed at https://tripalcultivate.github.io/docs/docs/administration/extend/validators.

carolyncaron:
This has been completed in [G4.158 Migrate the message processing method for ProjectGenusMatch to its validator class #159](https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159)

PR #159, opened 2025-05-09
G4.158 Migrate the message processing method for ProjectGenusMatch to its validator class

carolyncaron:
**Issue #158**

## Motivation
As part of our process to simplify our validator plugins (see the following issue in TripalCultivate: https://github.com/TripalCultivate/TripalCultivate/issues/59), we want to move our process[VALIDATOR_NAME]failures() methods to the validator class itself. The first validator we want to try this with is ProjectGenusMatch, since it is exclusive to the TripalCultivate Phenotypes module. This means we can more easily test this PR manually since we have access to the Share Importer, and no code has to move to the TripalCultivate base module, which would require 2 separate PRs.

## What does this PR do?
- [x] Creates the method `processSimpleList()` in the `ProjectGenusMatch` validator class.
- [x] Creates a separate test class for automated testing
- [x] Removes `processProjectGenusMatchFailures()` from the Share Importer and modifies its existing test.

### Automated Testing
**ValidatorProjectGenusMatchProcessTest->testProcessSimpleList()**
- [x] Tests the 3 valid cases without any tokens
- [x] Tests the token functionality
   - [x] 1 + 2 different tokens provided 
   - [x] Using a custom case message
   - [x] A custom case message including tokens

**ValidatorProjectGenusMatchProcessTest->testProcessSimpleListExceptions()**
- [x] The case string implying passed validation
- [x] A non-recognizable case string

### Manual Testing
1. Start a fresh docker image/container on this branch.
2. Navigate to your site and create a Research Experiment.
3. Now create 2 different organisms, making sure the Genus is different.
4. Configure Trait, Method, Unit and DB for both genus at `site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology`
5. Go to Admin > Devel > PHP Execute and submit the following code to attach **ONE** genus to your research experiment:
```
$m = \Drupal::service('trpcultivate_phenotypes.genus_project');
$m->setGenusToProject(1, GENUS, FALSE);
```
6. Navigate to the Open Science Phenotype Importer (can go to `site/admin/tripal/loaders/trpcultivate-phenotypes-share-importer`)
7. Select your Research experiment from the autocomplete, and then one of your genus from the dropdown.
8. Upload and submit a test file for upload. It does not matter if it will pass validation or not, only that it is a `.tsv` file.
9. Note the validation result for `Research Experiment exists and has been configured with selected genus` and make sure the message makes sense for your case (ie. whether you selected the genus that is paired to your research experiment or not).
10. Repeat steps 7 - 9 for the other genus.

NOTE: These steps test 2 out of the 3 possible failure cases for the ProjectGenusMatch validator. It is difficult to test a non-existent project in this form but if you know a way to do it, feel free to try it!


reynoldtan:
Here is my initial review.

1. Since the process message is now with the main logic of the validator, you could centralize the case messages. For example:

```
private $case_message = [
  ‘case-message-1’ => ‘Project does not exist’,
  ‘case-message-2’ => ‘Project has no genus set and could not compare with the genus provided’, 
  ‘case-message-3’ => ‘Genus does not match a genus set to the project’,
];

Main logic:
if ($project_rec <= 0 || empty($project_rec)) {
      // The project provided, whether the name or project id, does not exist.
      $case = $this->case_message[‘case-message-1’];
      $valid = FALSE;
      $failed_items = ['project_provided' => $project];
    }


Process method:
    // Check for one of the expected cases. Use the message stored in the
    // provided tokens array if set, otherwise use our default case message.
    if ($failure['case'] == $this->case_message[‘case-message-1’]) {
      $message = $combined_tokens['case-message-1'];
      $items = [
        '[project]: ' . $failure['failedItems']['project_provided'],
      ];
    }
```

2. Use dependency injection to inject the token service. It would be useful if this service is injected at a global scope (ie. make available in validator base) so we don’t need to inject this service for every validator. Alternatively, the replacement is specific enough that I think the token resolution can be achieved using simple str_replace().

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/86db8b6e66c1889e366abf8755082ccb76779603/trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php#L255

3. In the test, an unused property $genus at line 44.
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/86db8b6e66c1889e366abf8755082ccb76779603/trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchProcessTest.php#L44

laceysanderson:
> Since the process message is now with the main logic of the validator, you could centralize the case messages.

We can't use a typical variable since this is a static method but we **could use a static variable** in both cases. That said the `case-message-1` is less readable to me in the validate method but it's needed in the process method 🤔 

> Use dependency injection to inject the token service. It would be useful if this service is injected at a global scope (ie. make available in validator base) so we don’t need to inject this service for every validator. 

We cannot use dependency injection here because its a static method. That said, I don't think it will provide a performance hit to include it multiple times since Drupal should just pull the instance from the container 🤔 We could looking into caching it if needed though.

> Alternatively, the replacement is specific enough that I think the token resolution can be achieved using simple str_replace().

I would greatly prefer to use the service as compared to `str_replace()` in case we have need for more complicated tokens in other validators/future and for error checking/testing reasons.

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -154,4 +177,107 @@ public function validateMetadata(array $form_values) {
...
+
+    // Tripal Token Service.
+    $service_TripalTokensParser = \Drupal::service('tripal.token_parser');
+
+    // Now replace any tokens that are in our message.
+    $replaced_message = $service_TripalTokensParser->replaceTokens($message, $combined_tokens);
+    // Iterate through our items and replace any tokens there.
+    foreach ($items as $key => $item) {
+      $items[$key] = $service_TripalTokensParser->replaceTokens($item, $combined_tokens);
+    }
```
laceysanderson: For ease since we will be repeating this in every validator lets condense this and use [TripalTokenParser->replaceTokensArray()](https://github.com/tripal/tripal/blob/4.x/tripal/src/Services/TripalTokenParser.php#L61) to replace the items. Maybe we could also add a comment about why we're not using dependency injection here as a reminder to ourselves?

Something like:

```suggestion
    // Now replace any tokens that are in our message or items.
    // We use the Tripal Token Parser service to ensure we support more complicated tokens.
    // NOTE: we can't use depenendency injection since this is a static method.
    $service_TripalTokensParser = \Drupal::service('tripal.token_parser');
    $replaced_message = $service_TripalTokensParser->replaceTokens($message, $combined_tokens);
    $items = $service_TripalTokensParser->replaceTokensArray($items, $combined_tokens);
```

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -154,4 +177,107 @@ public function validateMetadata(array $form_values) {
...
+   *   - 'case-message-1': "Project does not exist"
+   *   - 'case-message-2': "Project has no genus set and could not compare with
+   *     the genus provided"
+   *   - 'case-message-3': "Genus does not match a genus set to the project".
+   *
+   * @return array
+   *   A render array of type unordered list which is used to display feedback
+   *   to the user about the case that failed and the failed items from the
+   *   input file. Each item in the list contains either the project name or the
+   *   genus, whichever one caused the failure.
```
laceysanderson: Technically both project/genus are included for some cases :thinking: plus it's actually and item containing an unordered list. What about something like:

> A render array used to display feedback to the user about the failure. Specifically, we return an item where the title describes the case triggered and the value is an unordered list providing the genus and/or project name (each as their own list item). The items in the list are chosen based on the case in order to provide applicable context to the failure.

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -154,4 +177,107 @@ public function validateMetadata(array $form_values) {
...
+   *   The following tokens can be specfied as keys, with value as the
+   *   replacement value for the token. These apply to all failure cases.
+   *   - 'project': replaces the word "project".
+   *   - 'contact-admin': replaces the phrase "contact your administrator".
+   *   The following token keys will substitute the entire existing case message
+   *   with the value of that token.
+   *   - 'case-message-1': "Project does not exist"
+   *   - 'case-message-2': "Project has no genus set and could not compare with
+   *     the genus provided"
+   *   - 'case-message-3': "Genus does not match a genus set to the project".
```
laceysanderson: Since these are actually in the default_tokens variable, we could add an `@see` to reference that. That way if these change we don't end up with the docblock not matching the values. Then we can define these tokens more free-text like. For example, 
`- 'case-message-1': the message to provide to the user when the project entered does not exist.`.

laceysanderson (commented):
Here are some comments after doing a code review :-) PS. The tests look perfect to me and I love some of the messages ;-p

carolyncaron:
Replying to @reynoldtan:

> 3. In the test, an unused property $genus at line 44.
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/86db8b6e66c1889e366abf8755082ccb76779603/trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchProcessTest.php#L44

I think I intended to make things easier for myself, and then I didn't 😆 This has been addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/e880a36078b7ef0583aba70b083fd155aa72bf0b!

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: Oooh, this is much simpler! Thanks for catching that. Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/8ddf34b21be7278507f9e04e50cb75559966f4d5

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: I broke down your suggested text into bullet points in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/c4583ed2de1ff495adc9723f431e59f7c87a3cb9. Please let me know if you like it!

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: This is a great suggestion - I also applied it to the 'project' and 'contact-admin' tokens above this. This is also addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/c4583ed2de1ff495adc9723f431e59f7c87a3cb9

laceysanderson: Yes, love these updates!

laceysanderson (dismissed):
✅ Code looks good to me now :-)
✅ Manually tested where the genus was paired to the project and got the pass I expected.
Screenshot 2025-05-23 at 4 06 57 PM

✅ Manually tested where the genus WAS NOT paired to the project and got the failure I expected.
Screenshot 2025-05-23 at 4 09 17 PM

reynoldtan:
> We can't use a typical variable since this is a static method but we **could use a static variable** in both cases. That said the `case-message-1` is less readable to me in the validate method but it's needed in the process method 🤔

Opps not the best example from me using the key case-message-1 but the goal is to eliminate the redundancy of encoding exactly the same text in two places. Yes for **we could use a static variable** 👍 

I appreciate the note regarding the token_parser 👍 👍 👍 

One more comment...
1. In many of our doc comment relating to class property of type array, we describe each key in the the doc block. For example https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/85098f37b5df6e758d05a429649a2dc26d3caac4/trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php#L38

carolyncaron:
> > We can't use a typical variable since this is a static method but we **could use a static variable** in both cases. That said the `case-message-1` is less readable to me in the validate method but it's needed in the process method 🤔
> 
> Opps not the best example from me using the key case-message-1 but the goal is to eliminate the redundancy of encoding exactly the same text in two places. Yes for **we could use a static variable** 👍
> 
> I appreciate the note regarding the token_parser 👍 👍 👍
> 
> One more comment...
> 1. In many of our doc comment relating to class property of type array, we describe each key in the the doc block. For example https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/85098f37b5df6e758d05a429649a2dc26d3caac4/trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php#L38

I addressed your last comment in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/412ee833f356a3eeb4869fb8b900860f1aaa5f61 and I decided to convert the token keys for the full messages from 'case-message-1' to something more specific (in this case, 'case-no-project'), which I like a lot more!

That said, I'm really hung up on whether it makes sense to add a new static property to map the developer case messages between the validate and process methods. I feel it takes up so much room in the class, whereas if we need to modify messages in the future, we can just use a find-and-replace (this is the benefit of keeping the process method in the same class as the validate).

Here is what I think the property could look like:
```
  /**
   * An array of developer case messages for ProjectGenusMatch.
   *
   * The keys are shorthand referrences to the case, whereas the value is the
   * full length developer case message. NOTE: These messages are NOT the same
   * as the one that gets displayed to the user after processing validation.
   * See $default_tokens property for an idea of what those look like.
   *
   * @var array
   */
  protected static array $dev_cases = [
    'case-no-project' => 'Project does not exist',
    'case-no-paired-genus' => 'Project has no genus set and could not compare with the genus provided',
    'case-project-genus-mismatch' => 'Genus does not match a genus set to the project',
    'case-valid' => 'Project exists and project-genus match the genus provided',
  ];
```
Using it would look like:
`$case = ProjectGenusMatch::$dev_cases['case-no-project'];`
But, for me, this isn't any more readable than just assigning the developer string directly.

Lastly, this would tempt me to just replace all developer case messages from a string to a shorter hyphenated version in ALL validators, and I don't think the effort of doing that is worth it. 🙈  But this situation should serve as something to think about in the future the next time we need something similar to a 'developer string', because I think keeping them very short and hyphenated opens up the opportunity to use them as array keys.

reynoldtan:
Thanks for addressing my other comments and well done on making the keys more specific, I like it a lot too.

To me, adding the dev case messages makes perfect sense. It not only saves us from duplicating the same text multiple places but also helps us future-proof our validators.

Right now, typing text in two places feels a bit fragile. If we change a word, we have to remember every location. The risk of inconsistency is also likely if we misspelled especially as we add more methods down the line.

> I feel it takes up so much room in the class

I am not sure what you mean it takes up room - if you mean readability? I think using the array actually improves clarity. Thanks to your meaningful keys, it is very readable and easy to follow. 

> we can just use a find-and-replace

That is a fair point but I think now they are all in one file, this is an opportunity where PHP give us more powerful options to centralize these values so that we can change once and safely.

> this would tempt me to replace all developer messages

I think that is a great direction to go.. doing the cleanup once with consistent keys would make our validators easier to maintain over time - here find-and-replace really shines.


```

  /**
   * An array of developer case messages for ProjectGenusMatch.
   *
   * The keys are shorthand referrences to the case, whereas the value is the
   * full length developer case message. NOTE: These messages are NOT the same
   * as the one that gets displayed to the user after processing validation.
   * See $default_tokens property for an idea of what those look like.
   *
   * @var array
   */
  protected static array $dev_cases = [
    'case-no-project' => 'Project does not exist',
    'case-no-paired-genus' => 'Project has no genus set and could not compare with the genus provided',
    'case-project-genus-mismatch' => 'Genus does not match a genus set to the project',
    'case-valid' => 'Project exists and project-genus match the genus provided',
  ];
```

Gorgeous 😍

carolyncaron:
We have settled on combining the above example property ($dev_cases) with the existing tokens property ($default tokens). Ie:
```
$mapping = [
  'case-no-project' => [
    'token' => 'case-no-project',
    'dev-case' => 'Project does not exist'
    'default-msg' => 'The selected [project] does not have a genus paired to it. Please [contact-admin] to have this set up'
  ],
  
  ...

];

// Example use
$validation_result['case'] = self::$mapping['case-no-project']['dev-case'];
```

This allows us to future-proof to prevent further duplication of the developer case strings, but we are not REQUIRED (at least, at this moment) to use the mapping for that purpose. This property will, of course, be needed for accessing the tokens and their values.

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -26,6 +27,76 @@ class ProjectGenusMatch extends TripalCultivatePhenotypesValidatorBase implement
...
+   *  The following tokens are implemented for this mapping, with the following
+   *  descriptions for their 'default-msg' values:
+   *  - 'project': the word to use when referring to the project.
+   *  - 'contact-admin': the phrase to use when the user needs a privileged
+   *    administrator to fix the problem.
+   *  - 'case-no-project': the message when a project does not exist.
+   *  - 'case-no-paired-genus': the message when a project has no genus set to
+   *    it.
+   *  - 'case-project-genus-mismatch': the message when the genus selected by
+   *    the user is not configured to the selected project.
```
laceysanderson: This one feels not ideal right now and while I know we have the `@todo` to move it into base... perhaps there is more we can do now? What about the following? 

I've essentially moved your class-specific docs directly under `@var` with a short pre-amble to what this variables is. My thoughts are this part would stay and provide an example for how to document each specific mapping var. Then I kept your original generic docs which I like under the @todo. That way once we make the move we can just remove everything below/including the `@todo` and replace it with an `@see`.

```suggestion
   * @var array
   *  An associative array mapping tokens to their details where each element
   *  may contain a default message but must contain the dev-case and token.
   *  The following tokens are implemented for this mapping, with the following
   *  descriptions for their 'default-msg' values:
   *  - 'project': the word to use when referring to the project.
   *  - 'contact-admin': the phrase to use when the user needs a privileged
   *    administrator to fix the problem.
   *  - 'case-no-project': the message when a project does not exist.
   *  - 'case-no-paired-genus': the message when a project has no genus set to
   *    it.
   *  - 'case-project-genus-mismatch': the message when the genus selected by
   *    the user is not configured to the selected project.
   *
   * @todo Move this generic documentation to ValidatorBase in TripalCultivate.
   * This mapping starts with all of the potential cases for this validator,
   * followed by additional tokens which are subsitutable within the message(s)
   * provided to the user when validation fails.
   *
   * For each case, the array keys are the substitutable tokens for the entire
   * case message, and MUST contain the prefix of 'case-', and contain the
   * following key-value pairs:
   * - 'token': the same token (same as the parent key- this can helpful for
   *   code readability). Recall that it must contain the prefex 'case-'.
   * - 'dev-case': The short, developer-focussed string describing the case.
   * - 'default-msg': An informative message that gets displayed to the user
   *   when validation fails for this particular case. This can contain any
   *   number of smaller, non case-specific tokens contained in square brackets.
   * NOTE: The token 'case-valid' is reserved for the valid case for this
   * validator, and does not have a corresponding default message.
   *
   * For all remaining tokens, the array key is the substitutable token, with
   * the following key-pairs:
   * - 'token': the substitutable text in a message. This text would become
   *   flanked by brackets within a message. For eg. [token]
   * - 'default-msg': A string that would substitute the associated token
   *   within a case message.
   *

```

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -152,4 +229,112 @@ public function validateMetadata(array $form_values) {
...
+   *   An associative array that was returned by the ProjectGenusMatch validator
+   *   in the event of failed validation. It contains the following keys:
+   *   - 'case': a developer-focused string describing the case checked.
+   *   - 'valid': FALSE to indicate that validation failed.
+   *   - 'failedItems': an array of items that failed with the following keys.
+   *     - 'project_provided': The name of the project provided.
+   *     - 'genus_provided': The name of the genus provided.
+   * @param array $tokens
+   *   [OPTIONAL] An array of values to use for token replacement.
+   *   @see ProjectGenusMatch::$default_tokens
```
laceysanderson: Update to match the new variable.

```suggestion
   *   @see ProjectGenusMatch::$mapping
```

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
```diff
@@ -152,4 +229,112 @@ public function validateMetadata(array $form_values) {
...
+   *     provided as context depending on the case that was triggered:
+   *     - The project name selected by the user
+   *     - The genus selected by the user
+   *
+   * @throws \Exception
+   *   - If the failure parameter was not formatted properly.
+   *   - If the case string returned by the validator implied validation passed.
+   *   - If the case string returned by the validator is not recognized.
+   */
+  public static function processSimpleList(array $failure, array $tokens = []) {
```
laceysanderson: In hindsight based on recent conversations we could be more specific with this method name if we want to althought this one is likely fine. A more specific suggestion might be `processItemWithContextList()` to make it clear that we are returning an item form element with a nested list containing context on the case that was triggered.

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchProcessTest.php
```diff
@@ -0,0 +1,437 @@
...
+
+    // Check the render array here.
+    $selected_message_title = $this->cssSelect('div.tcp-project-genus-match-failures label');
+    $provided_message = (string) $selected_message_title[0];
+    $this->assertStringContainsString($expectations['expected_message'], $provided_message, 'The message expected from processing ProjectGenusMatch failures for this scenario did not match the one in the rendered output.');
+
+    // Next, check for expected items. Make sure we have the expected 1 item.
+    $selected_list_items = $this->cssSelect('div.tcp-project-genus-match-failures ul li');
+    $list_item_count = count($selected_list_items);
+    $this->assertEquals($expectations['expected_item_count'], $list_item_count, 'We expected ' . $expectations['expected_item_count'] . ' list items in the render array from processing ProjectGenusMatch failures, but instead found ' . $list_item_count . '.');
```
laceysanderson: Maybe use assertCount() instead of assertEquals() here?

laceysanderson (dismissed):
✅ Code looks good. Minor suggestions for improvements inline.
✅ Code coverage is 100% for the ProjectGenusMatch validator.
✅ Manually tested where the genus was paired to the project and got the pass I expected. Tested when two genus were paired and both succeeded.
Screenshot 2025-06-09 at 11 58 59 AM

✅ Manually tested where the genus WAS NOT paired to the project and got the failure I expected.
Screenshot 2025-06-09 at 11 59 23 AM

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: I like this suggestion, but I'm now wondering if I should go ahead and remove the @todo block in this PR now? This text has since been copied over to `GermplasmNameExists` in PR https://github.com/TripalCultivate/TripalCultivate/pull/63 and is already diverging slightly to improve upon the wording. 🙈 In that PR, we can finalize moving it to the ValidatorBase class in TripalCultivate and thus, would not need a new PR in this module simply to remove it here.

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: I like the use of "ItemWith"! Not so much Context (as you predicted 😆). Updating this to `processItemWithSimpleList` in the next commit!

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchProcessTest.php
carolyncaron: Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/0b42dcd83ac3a4e3a97e22327d04696ccf54dbcd

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
carolyncaron: Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/159/commits/0b42dcd83ac3a4e3a97e22327d04696ccf54dbcd!

reynoldtan (dismissed):
This code looks great. Well done, everyone!

trpcultivate_phenotypes/src/Plugin/Validators/ProjectGenusMatch.php
laceysanderson: Ahhh, yes! That is perfect!

laceysanderson (approved):
Looks great!

PR #161, opened 2025-05-14
G4.161 - Allow multiple genus to be set to a project

### LINKED ISSUE #156: G4.156 Allow for multiple genus to be configured with a single Research Experiment
### Branch

g4.156-multipleGenusProjectMatches

### Groups

Group 4 - Services

### Describe

While reviewing PR #150 , I was unable to configure more than one genus for a single Research Experiment in the Phenotypes Share importer form.

### Steps to Reproduce
1. Clone this repo, and if PR #150 has not yet been merged, checkout branch `g0.149-ImplementMoreValidators`
2. Start a fresh docker image/container of this module. 
3. Navigate to your site and create a Research Experiment.
4. Now create 2 different organisms, making sure the Genus is different.
5. Configure Trait, Method, Unit and DB for each genus at `site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology`
6. Go to Admin > Devel > PHP Execute and submit the following code to attach both genus to the same research experiment:
```
$m = \Drupal::service('trpcultivate_phenotypes.genus_project');
$m->setGenusToProject(1, GENUS1, FALSE);
$m->setGenusToProject(1, GENUS2, FALSE);
```
7. Navigate to the Open Science Phenotype Importer (can go to `site/admin/tripal/loaders/trpcultivate-phenotypes-share-importer`)
8. Select your Research experiment from the autocomplete, and then one of your genus from the dropdown.
9. Upload and submit a test file for upload. It does not matter if it will pass validation or not, only that it is a `.tsv` file.
10. Note the validation result for `Research Experiment exists and has been configured with selected genus`. 
11. Repeat steps 8 - 10 for the other genus.
12. For extra testing, try step 6 again but switch the order of the genus being set, then repeat steps 7 - 11.

### Expected Behaviour
I would expect that both genus would be considered valid since both were supposedly set to the same research experiment.

### Design

I think this is most likely a bug or assumption(s) made in `TripalCultivatePhenotypesGenusProjectService.php`. Particularly, method `setGenusToProject` implies that multiple genus can be set:

Image

reynoldtan:
**Issue #156 Allow multiple genus to be assigned to a project**

## Motivation

This PR revises the Genus-Project service to allow assignment of one or many genus to a project.

## What does this PR do?

1. Update genus-project service setter method to insert a genus instead of overriding.
2. To keep the method simple, the replace option is no longer supported since it requires search and replace logic.
3. Update project genus match validator to account for multiple genus.
4. Update automated tests.
5. Disable auto-select genus in the Importer Share form

### Automated Testing

ServiceGenusProjectTest->testGenusProjectService
- Test one genus
- Test 2 genus
- Test 5 genus
- Test getter returned correct set of genus
- Test that the rank value assigned to each genus is a sequence order

### Manual Testing

1. In a fully configured clone, create a research experiment Tripal content type in my site/bio_data/add/research_experiment
2. Create organism in my site/bio_data/add/organism. Insert as many to test multiple genus.
3. Configure every genus added in step 2 in my site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology
4. In your sites deve/php page, paste the following code:

```
$m = \Drupal::service('trpcultivate_phenotypes.genus_project');
$project_id = 1; // In a fresh clone with one project, this is 1.

// Add one.
$genus = REPLACE WITH A GENUS NAME ADDED IN STEP 2
$m->setGenusToProject($project_id, $genus);
// Pull the genus set.
$p_genus = $m->getGenusOfProject($project_id);
dpm($p_genus);
```

Repeat to add additional genus.

To inspect from a database level:
```
$genus = \Drupal::service('tripal_chado.database')
  ->select('1:projectprop', 'pp')
  ->fields('pp', ['project_id', 'value', 'rank'])
  ->condition('pp.project_id', $project_id, '=')
  ->execute()
  ->fetchAll();

dpm($genus);

```

### DIFF
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
@@ -103,7 +103,7 @@ class TripalCultivatePhenotypesGenusProjectService {
         foreach ($project_genus as $row) {
           if ($row->value == $genus) {
             $project_has_genus = TRUE;
-            $break;
+            break;
           }
         }
 
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
@@ -61,18 +61,17 @@ class TripalCultivatePhenotypesGenusProjectService {
   /**
    * Assign a genus to an experiment/project.
    *
+   * Each genus-project relationship is an entry in projectprop table.
+   *
    * @param int $project
    *   Project (project id number) the parameter $genus will be assigned to.
    * @param string $genus
    *   Genus name/title.
-   * @param bool $replace
-   *   True to replace existing genus of a project with a different genus.
-   *   Default to False.
    *
    * @return bool
    *   True, genus was set successfully or false on error/fail.
    */
-  public function setGenusToProject($project, $genus, $replace = FALSE) {
+  public function setGenusToProject($project, $genus) {
     $error = 0;
 
     if (empty($project) || $project <= 0) {
@@ -135,13 +134,13 @@ class TripalCultivatePhenotypesGenusProjectService {
   }
 
   /**
-   * Get genus of an experiment/project.
+   * Get all genus assigned to a project.
    *
    * @param int $project
    *   Project (project_id number) to search.
    *
    * @return array
-   *   Key is genus/organism id number and value is the genus name/title.
+   *   An array of all the genus assigned to a project.
    */
   public function getGenusOfProject($project) {
     $genus_project = [];
@@ -154,34 +153,15 @@ class TripalCultivatePhenotypesGenusProjectService {
 
       // Fetch genus paired to a project. If multiple genus have been set prior,
       // restrict search to genus that are active/configured using this module.
-      $result = $this->chado_connection->query("
-        SELECT organism_id AS id, genus FROM {1:organism} WHERE genus IN (
-          SELECT value::VARCHAR FROM {1:projectprop} WHERE
-            project_id = :project_id AND
-            type_id = :type_id AND
-            LOWER(value) IN (:active_genus[])
-        ) ORDER BY genus ASC
-        ", [
-          ':project_id' => $project,
-          ':type_id' => $this->sysvar_genus,
-          ':active_genus[]' => $active_genus,
-        ]
-      )
-        ->fetchAllKeyed(0, 1);
+      $result = $this->chado_connection->select('1:projectprop', 'pp')
+        ->fields('pp', ['value'])
+        ->condition('pp.project_id', $project, '=')
+        ->condition('pp.type_id', $this->sysvar_genus, '=')
+        ->where('LOWER(pp.value) IN (:active_genus[])', [':active_genus[]' => $active_genus])
+        ->orderBy('value', 'ASC')
+        ->execute();
 
-      if ($result) {
-        foreach ($result as $organism_id => $genus) {
-          $item = ['id' => $organism_id, 'genus' => $genus];
-
-          if (count($result) == 1) {
-            $genus_project = $item;
-            break;
-          }
-          else {
-            $genus_project[] = $item;
-          }
-        }
-      }
+      $genus_project = $result->fetchCol();
     }
 
     return ($genus_project) ? $genus_project : [];
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
@@ -152,7 +152,7 @@ class TripalCultivatePhenotypesGenusProjectService {
         ->condition('pp.project_id', $project, '=')
         ->condition('pp.type_id', $this->sysvar_genus, '=')
         ->where('LOWER(pp.value) IN (:active_genus[])', [':active_genus[]' => $active_genus])
-        ->orderBy('value', 'ASC')
+        ->orderBy('rank', 'ASC')
         ->execute();
 
       $genus_project = $result->fetchCol();
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
@@ -151,19 +151,33 @@ class TripalCultivatePhenotypesGenusProjectService {
       // Fetch genus paired to a project. If multiple genus have been set prior,
       // restrict search to genus that are active/configured using this module.
       $result = $this->chado_connection->query("
-        SELECT organism_id AS id, genus FROM {1:organism}
-        WHERE genus IN (SELECT value::VARCHAR FROM {1:projectprop}
-          WHERE project_id = :project_id AND type_id = :type_id AND LOWER(value) IN (:active_genus[]))
-      ", [':project_id' => $project, ':type_id' => $this->sysvar_genus, ':active_genus[]' => $active_genus]);
+        SELECT organism_id AS id, genus FROM {1:organism} WHERE genus IN (
+          SELECT value::VARCHAR FROM {1:projectprop} WHERE
+            project_id = :project_id AND
+            type_id = :type_id AND
+            LOWER(value) IN (:active_genus[])
+        ) ORDER BY genus ASC
+        ", [
+          ':project_id' => $project,
+          ':type_id' => $this->sysvar_genus,
+          ':active_genus[]' => $active_genus,
+        ]
+      )
+        ->fetchAllKeyed(0, 1);
 
-      $genus_project = $result->fetchAll();
-    }
+      if ($result) {
+        foreach ($result as $organism_id => $genus) {
+          $item = ['id' => $organism_id, 'genus' => $genus];
 
-    if (count($genus_project) == 1) {
-      $genus_project = [
-        'id' => $genus_project[0]->id,
-        'genus' => $genus_project[0]->genus,
-      ];
+          if (count($result) == 1) {
+            $genus_project = $item;
+            break;
+          }
+          else {
+            $genus_project[] = $item;
+          }
+        }
+      }
     }
 
     return ($genus_project) ? $genus_project : [];
--- a/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
+++ b/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
@@ -89,36 +89,40 @@ class TripalCultivatePhenotypesGenusProjectService {
       $is_active_genus = (in_array($g, array_keys($this->sysvar_genusontology))) ? TRUE : FALSE;
 
       if ($is_active_genus) {
-        $result = $this->chado_connection->query("
-          SELECT projectprop_id AS id FROM {1:projectprop}
-          WHERE project_id = :project_id AND type_id = :type_id LIMIT 1
-        ", [':project_id' => $project, ':type_id' => $this->sysvar_genus]);
+        // Pull all genus assigned to the project.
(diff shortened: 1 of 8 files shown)


trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
```diff
@@ -408,6 +443,18 @@ public function testProjectGenusInput() {
...
+    $project = $this->test_project['project-with-2-configgenus']['id'];
+    $genus = $this->test_genus['configured-secondary'];
+
+    $form_values = ['project' => $project, 'genus' => $genus];
+    $validation_status = $instance->validateMetadata($form_values);
+
+    $this->assertEquals('Project exists and project-genus match the genus provided', $validation_status['case'],
+      'Project genus match validator case title does not match expected title for a valid project+genus (project with multiple genus).');
+    $this->assertTrue($validation_status['valid'], 'A valid project+genus must return a TRUE valid status.');
+    $this->assertEmpty($validation_status['failedItems'], 'A valid project+genus does not return a failed item value.');
```
carolyncaron: I wonder if this PR is a good opportunity to separate out `testProjectGenusInput()` from MetadataInputTest.php? Some code was added to `setUp()` in this PR which I assume is not needed for `testGenusInput()`. This PR could create a new test file specific for testing the ProjectGenusMatch validator and leave the pre-existing test and setup for `GenusExists()` in this file for a future PR to deal with.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -97,27 +120,46 @@ protected function setUp(): void {
...
+        2,
+        [
+          'project_genus' => [
+            'Genus1',
+            'Genus2',
+          ],
+        ],
+      ],
+
+      // #3: A project with 5 genus.
```
carolyncaron:
```suggestion
      // #2: A project with 5 genus.
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -150,17 +147,18 @@ public function getGenusOfProject($project) {
...
-        LIMIT 1
-      ", [':project_id' => $project, ':type_id' => $this->sysvar_genus, ':active_genus[]' => $active_genus]);
-
-      $genus_project = $result->fetchObject();
+      $result = $this->chado_connection->select('1:projectprop', 'pp')
+        ->fields('pp', ['value'])
+        ->condition('pp.project_id', $project, '=')
+        ->condition('pp.type_id', $this->sysvar_genus, '=')
+        ->where('LOWER(pp.value) IN (:active_genus[])', [':active_genus[]' => $active_genus])
+        ->orderBy('value', 'ASC')
```
carolyncaron: Would it be more fitting here to also order by rank, to be consistent with the setter?

carolyncaron (changes requested):
My manual test for this worked flawlessly! Well done! 🎉 

I've made a few suggestions having to do with tests below. I also think the following cases should also be covered in the class `ServiceGenusProjectTest.php`:
- [x] Testing the getter with a non-configured genus.
- [x] Testing the setter when no genus have been set yet.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
reynoldtan: Result sorted by rank in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/161/commits/bf8f5fac26f0ab47c0583ee47ddecda3c41dc0e8

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -128,72 +206,213 @@ public function provideGenusProjectForGenusProjectService() {
...
+      $this->assertTrue($is_set, 'Project Genus Service failed to set a genus to project in scenario: ' . $scenario);
+    }
+
+    $set_genus = $this->service_PhenoGenusProject->getGenusOfProject($this->project);
+
+    $this->assertEquals(
+      count($set_genus),
+      count($expected['project_genus']),
+      'The number of genus set does no match expected genus count in scenario: ' . $scenario
+    );
```
laceysanderson: I would use assertCount here with the first param being $genus_count and the second being $set_genus. This will give more feedback on failure.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -128,72 +206,213 @@ public function provideGenusProjectForGenusProjectService() {
...
+   *   input values, respectively.
+   * @param array $expected
+   *   An array of expected values, with the following keys:
+   *     - 'is_set': a boolean value returned by setGenusToProject() method to
+   *       indicate the success or failure of the set genus to project request.
+   *     - 'log_message': the Tripal log error message about the failed value.
+   *
+   * @dataProvider provideInvalidValuesToGenusProjectService
+   */
+  public function testGenusProjectServiceWithInvalidValues($scenario, $input_value, $expected) {
```
laceysanderson, line 362: It would also be good to check the projectprop table to ensure that there are no records indicating genus for each scenario. That way we are not just relying on accurate return value.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
laceysanderson: I agree with Carolyn on this and would like to see this moved into it's own file in this PR.

laceysanderson (changes requested):
✅ Manually tested adding 5 genus to a project using `setGenusToProject()` and confirmed that they were set with the correct rank by checking the database table.
✅ Manually added a genus to a Research Experiment through the create content page and then used that project/genus combination in the importer without ever using `setGenusToProject()` and confirmed that the two approaches are equivalent. The importer correctly indicated that this pair was configured :-)

I do have code comments which are documented inline.

trpcultivate_phenotypes/tests/src/Kernel/Validators/MetadataInputTest.php
reynoldtan: Genus Exists test and Project Genus Match test have parted ways. I also renamed the MetadataInputTest to ValidatorGenusExistsTest since it is more about the genus exists more than metadata test. Revisions in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/161/commits/b3e6b50a3c50fa7810f7783d4e9fd5770e94068e

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
reynoldtan: Applied in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/161/commits/b3e6b50a3c50fa7810f7783d4e9fd5770e94068e

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
reynoldtan: Applied in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/161/commits/b3e6b50a3c50fa7810f7783d4e9fd5770e94068e

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchTest.php
```diff
@@ -0,0 +1,432 @@
...
+        'A project-genus match',
+        'with-config-genus',
+        [
+          'case' => 'Project exists and project-genus match the genus provided',
+          'valid' => TRUE,
+          'failedItems' => [],
+        ],
+      ],
+
+      // #5: Project and genus matched from a project with multiple genus set.
```
carolyncaron:
```suggestion
      // #4: A project set with a non-configured genus.
      [
        'Unconfigured genus to a project',
        'with-unconfig-genus',
        [
          'case' => 'Project has no genus set and could not compare with the genus provided',
          'valid' => FALSE,
          'failedItems' => [
            'genus_provided' => 'genus',
          ],
        ],
      ],

      // #5: Project and genus matched.
      [
        'A project-genus match',
        'with-config-genus',
        [
          'case' => 'Project exists and project-genus match the genus provided',
          'valid' => TRUE,
          'failedItems' => [],
        ],
      ],

      // #6: Project and genus matched from a project with multiple genus set.
```

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php
```diff
@@ -131,16 +128,16 @@ public function setGenusToProject($project, $genus, $replace = FALSE) {
...
   /**
-   * Get genus of an experiment/project.
+   * Get all genus assigned to a project.
    *
    * @param int $project
    *   Project (project_id number) to search.
    *
    * @return array
-   *   Key is genus/organism id number and value is the genus name/title.
+   *   An array of all the genus assigned to a project.
```
carolyncaron:
```suggestion
   *   A list of all the genus assigned to a project.
```

carolyncaron (dismissed):
Looks great to me! I just made a couple of very nit-picky suggestions in comments, but consider this approved by me 😄

trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchTest.php
reynoldtan: my gosh, that was a great catch! I seem to always mess up this count.

carolyncaron: It happens to me too which is why I always check! 😆

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
```diff
@@ -97,27 +156,46 @@ protected function setUp(): void {
...
-   *   - A string, the name or title of a project.
-   *   - An array, the genus and species of the organism that will be
-   *     paired with the project. Keyed by 'genus' and 'species'.
+   *   - An integer indicating the number of genus to set to the project.
    *   - An array of expected values, with the following keys:
    *     - 'project_genus': the expected genus returned by the method
    *     getGenusOfProject().
-   *     - 'alternative_genus': a genus returned after setting a genus.
    */
   public function provideGenusProjectForGenusProjectService() {
```
laceysanderson: It looks like we are missing the case where we try to set a genus for a project which is already set. For example, if the genus `Tripalus` is already configured for `Project A` and then we call `setGenusToProject('Project A', 'Tripalus)`. This should just return TRUE without changing the prop record that already exists.

This is also needed to cover line 105 in the service:
```php
        // Determine if the project already had the genus.
	        foreach ($project_genus as $row) {
	          if ($row->value == $genus) {
	            $project_has_genus = TRUE;
	            $break;
	          }
	        }
```

NOTE: When copying the above code sample I also spotted another bug. It should be `break;` not `$break;`.

laceysanderson (changes requested):
Code-wise this is looking really good. We still have 100% coverage for the two validators and the split tests are nice to see <3 We are only missing coverage of one line in the service which is meantioned in my inline comments and is the result of a missing scenario.

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusProjectTest.php
reynoldtan: Good catch 👍  I added a test case in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/161/commits/52aed5ee5c54dce58fc46250eb9fc98b4c32a614

laceysanderson (approved):
Nice, I like the new test case, the service has 100% coverage and this all looks good to me now. Once we confirm tests pass after the merge update, this should be ready to merge!

Issue #164, opened 2025-06-02
G5.164 Improve Test Coverage


laceysanderson:
### Branch

g5.164-improveTestCoverage

### Groups

Group 5 | Automated Testing | Performance

### Describe

We need to improve our test coverage in this module to ensure that we catch and Drupal/Tripal deprecation and to improve stability/maintenance. This is a subtask of #70.

We are missing coverage in the following:

- trpcultivate_phenotypes/trpcultivate_phenotypes.module trpcultivate_phenotypes_install_ontologyterms()
- trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusOntologyService.php saveGenusOntologyConfigValues($config_values)
- trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTermsService.php loadTerms($schema = NULL)
- trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTermsService.php saveTermConfigValues($config_values)
- trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php almost all methods

NOTE: the following are being fixed in current PRs:
- trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusProjectService.php setGenusToProject($project, $genus, $replace = FALSE)

NOTE: the lack of coverage in the Forms and controllers is due to these being covered by Functional tests which do not contribute to the coverage. We will address this in a different subissue.

## Design

Update existing tests where possible and design new ones as needed. More specifically,

### trpcultivate_phenotypes/trpcultivate_phenotypes.module trpcultivate_phenotypes_install_ontologyterms()

This method is called when the module is installed. It acts to setup the Drupal site by installing the default controlled vocabulary terms.

We don't even have an existing test for anything close to this 🙈 We will want to create a new Kernel test.

### trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusOntologyService.php saveGenusOntologyConfigValues($config_values)

See trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusOntologyTest.php

This method works to configure the vocabulary and databases to use for a given genus. We call this configuring a genus for use with this module since it indicates where new phenotypic traits should be created.

This method already has coverage by the kernel test ServiceGenusOntologyTest. We will want to expand this test to include the scenarios where

1. We call `saveGenusOntologyConfigValues($config_values)` before we call `defineGenusOntology()`.
[...]

PR #166, opened 2025-06-04
G5.164 Improve Test Coverage

vidu-wil:
**Issue #164**

## Motivation

Adds additional automated test coverage for the `saveGenusOntologyConfigValues($config_values)` method in TripalCultivatePhenotypesGenusOntologyService.php, `saveTermConfigValues($config_values)` in TripalCultivatePhenotypesTermsService.php,`TripalCultivatePhenotypesTraitsService` class and `trpcultivate_phenotypes_install_ontologyterms()` method in trpcultivate_phenotypes.module. This addresses the issue #164; a subtask of https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/70.

## What does this PR do?

1. Implements the test method `testInvalidGenusOntologyService` and its associated data provider `provideInvalidValuesForGenusOntologyService`.
2. Implements the test method `testSaveTermConfigValuesMethodInvalidKey`, `testSaveTermConfigValuesMethodFailure` to complete the test coverage for `saveTermConfigValues($config_values)` and their data providers.
3. Adds full code coverage for the the methods setTraitGenus(), getTraitMethod(), getTraitMethodUnit(), getMethodUnitDataType() and getPhenoCvTerm(), and partial coverage for the insertTrait() method in class `TripalCultivatePhenotypesTraitsService`. 
4. Implements the test class `PhenotypeTermInstallTest` to test the method `trpcultivate_phenotypes_install_ontologyterms()` that's being called at install.
5. Fixes some syntax errors on the files `trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesGenusOntologyService.php` and `trpcultivate_phenotypes/tests/src/Unit/ConfigWatermarkFormTest.php` to follow coding standards.

### Automated Testing

Automated tests for following scenarios:
1. Call saveGenusOntologyConfigValues($config_values) before we call defineGenusOntology().
2. The config passed in includes a config_name that is not present in $this->genus_ontology to confirm that we get the correct logged error.
3. The config passed in includes a genus that is not present in the acceptable genus keys to confirm that we get the expected log message.
4. Additionally for scenario 2 + 3 test whether the saveGenusOntologyConfigValues() method returns FALSE and getGenusOntologyConfigValues() does not return the invalid values we tried to set.
6. Failure case for the method `saveTermConfigValues($config_values)` when invalid term is given and when the config_values are invalid.
7. Triggering all the exception cases in the method `setTraitGenus()`, no genus set case in `insertTrait()` method, invalid trait case in `getTraitMethod()` method, invalid method case in `getMethodUnit()` method, invalid unit case and unit with multiple data types case in `getMethodUnitDataType()`, and no genus set case, invalid type case and duplicate term error case in `getPhenoCvterm()` method.
8. The method `trpcultivate_phenotypes_install_ontologyterms()` which is called at install to insert terms used by configuration page default field values.


laceysanderson:
These failing tests should be fixed by updating your branch by using the "Update branch" button above.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -486,6 +486,7 @@ public function getMethodUnit($method) {
    *   getPhenoCvterm()
    */
   public function getMethodUnitDataType($unit) {
+    $genus_config = $this->config;
```
reynoldtan: The variable $genus_config is used by the single if statement of the method. I would suggest moving the assignment statement in the if statement or just pass the required array value.

```
throw new \Exception(...
      $this->config['unit']['name'] . '. Multiple data types found for the same unit.');

```

trpcultivate_phenotypes/tests/src/Kernel/ServiceGenusOntologyTest.php
```diff
@@ -284,4 +307,112 @@ public function testGenusOntologyService($scenario, $genus, $species, $expected)
...
+   *   - A string, the genus name.
+   *   - A string, the species name.
+   *   - An array containing an invalid input for saveGenusOntologyConfigValues.
+   *   - An array of expected values, with the following keys:
+   *     - 'is_saved': a boolean value returned by
+   *        saveGenusOntologyConfigValues() method to indicate
+   *        the success or failure of saving the genus.
+   *     - 'log_message': the Tripal log error message about the failed value.
+   */
+  public function provideInvalidValuesForGenusOntologyService() {
```
reynoldtan:
```suggestion
  public static function provideInvalidValuesForGenusOntologyService() {
```

Test data provider must be declared public and static.

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -175,6 +198,83 @@ public static function provideTermsForSaveTermConfigValuesMethod() {
...
+   *   - An array, term input with the following keys:
+   *     - 'name': the name of the term.
+   *     - 'cv': the cv vocabulary the term will be associated.
+   *   - A string, the term identifier of the module the term array will
+   *     be saved and mapped to.
+   *   - An array of expected values, with the following keys.
+   *     - 'is_saved': the expected return value of the method.
+   *     - 'log_messages': the expected error log message.
+   */
+  public function provideInvalidTermsForSaveTermConfigValuesMethod() {
```
reynoldtan:
```suggestion
  public static function provideInvalidTermsForSaveTermConfigValuesMethod() {
```

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -175,6 +198,83 @@ public static function provideTermsForSaveTermConfigValuesMethod() {
...
+   * Data Provider: provides invalid config values to saveTermConfigValues().
+   *
+   * @return array
+   *   Each term test scenario is an array with the following values:
+   *   - A string, human-readable short description of the test scenario.
+   *   - An array or a string, invalid input for the config values.
+   *   - An array of expected values, with the following keys.
+   *     - 'is_saved': the expected return value of the method.
+   */
+  public function provideInvalidConfigForSaveTermConfigValuesMethod() {
```
reynoldtan:
```suggestion
  public static function provideInvalidConfigForSaveTermConfigValuesMethod() {
```

reynoldtan:
Impressive work! Big thanks for finding other non-standard pieces and fixing them in this PR. I just have this one suggestion...

1. In ServiceTermTest, relocate the test method closer to the data provider it is using to reduce scrolling required when cross-checking values.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -200,12 +222,140 @@ public function testTraitsServiceDatabaseExpectations() {
...
+      'and set the controlled vocabulary associated with the term.',
+      $exception_caught,
+      "We expected an exception to be caught for this scenario, but one wasn't thrown.",
+    );
+  }
+
+  /**
+   * Test the insertTrait() method.
+   */
+  public function testInsertTrait() {
```
carolyncaron: This test looks like it's only triggering an exception when a genus hasn't been set, but the name implies it's testing the whole method. I would suggest renaming to something such as `testInsertTraitUnsetGenus()` and update the description accordingly.

carolyncaron (changes requested):
Great work on improving test coverage in the module! 👏 

For the asserts that you added in `ValidTestTrait.php`, I would suggest you update the assert message to be a little more specific to the situation. This will make it a bit easier to debug in case any of these automated tests fail in the future. The assert messages in `ServiceGenusOntologyTest.php` and `ServiceTermTest.php`, however, look great!

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -131,6 +154,31 @@ public static function provideTermIdentifierForGetTermIdMethod() {
...
+  /**
+   * Test Term Service getTermId() method.
+   *
+   * @param string $scenario
+   *   A string, human-readable short descriptionn of the test scenario.
+   * @param string $input_term_identifier
+   *   A string, term identifier input.
+   * @param array $expected
+   *   An array of expected values, with the following keys.
+   *     - 'term_exists': boolean value to idicate if a term identifier exits
```
reynoldtan:
```suggestion
   *     - 'term_exists': boolean value to indicate if a term identifier exits
```

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -176,35 +224,105 @@ public static function provideTermsForSaveTermConfigValuesMethod() {
...
+   * @dataProvider provideTermsForSaveTermConfigValuesMethod
    */
-  public function testGetTermId($scenario, $input_term_identifier, $expected) {
-    $term_id = $this->service_PhenoTerms->getTermId($input_term_identifier);
-    $term_exists = ($term_id > 0) ? TRUE : FALSE;
+  public function testSaveTermConfigValuesMethod($scenario, $input_term, $term_identifier, $expected) {
+    // Create or fectch input term.
+    $term_exists = $this->chado_connection->select('1:cvterm', 'cvt')
+      ->fields('cvt', ['cvterm_id'])
+      ->condition('cvt.name', $input_term['name'])
```
reynoldtan:
```suggestion
      ->condition('cvt.name', $input_term['name'], '=')
```
I added the comparison operator to make it easier to read.

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -234,18 +353,80 @@ public function testSaveTermConfigValuesMethod($scenario, $input_term, $term_ide
...
+        ],
+      ],
+      // #1: Is not an array
+      [
+        'a string',
+        '',
+        [
+          'is_saved' => FALSE,
+        ],
+      ],
```
reynoldtan, line 434:
```suggestion
      ],
      // #2: Not a registered config name.
      [
        'not a config name',
        [
          'not_a_config_name' => 1,
        ],
        [
          'is_saved' => FALSE,
        ]
      ],
```
An Additional invalid config test case to test the inner if statement of the method.
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/b5b0450bb086e1318b5854394c2de1873bc40ef4/trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTermsService.php#L291

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -200,12 +222,141 @@ public function testTraitsServiceDatabaseExpectations() {
...
     $this->assertEquals($data_type->value, 'Quantitative', 'Unit property - additional type does not match expected value (Quantitative).');
   }
 
+  /**
+   * Test setTraitGenus() method for error cases.
+   */
+  public function testSetTraitGenus() {
+    // Test case for where the genus is not configured.
+    $exception_caught = FALSE;
+    $exception_message = 'NONE';
```
reynoldtan: This duplicate variable can be deleted.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -200,12 +222,141 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    catch (\Exception $e) {
+      $exception_caught = TRUE;
+      $exception_message = $e->getMessage();
+    }
+    $this->assertEquals(
+      'Exception: The genus Test Genus was not configured for' .
+      'use with Tripal Cultivate Phenotypes. To configure this genus, go to ' .
+      Url::fromRoute('trpcultivate_phenotypes.settings_ontology')->toString() .
+      ' and set the controlled vocabularies associated with this genus.',
+      $exception_caught,
```
reynoldtan:
```suggestion
      $exception_message,
```
Since you're capturing values for $exception_caught and $exception_message, add an assert statement to verify respective values. Here's an excellent example from Carolyn https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/b5b0450bb086e1318b5854394c2de1873bc40ef4/trpcultivate_phenotypes/tests/src/Kernel/Validators/ValidatorProjectGenusMatchProcessTest.php#L416

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -200,12 +222,141 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    ];
+
+    $exception_caught = FALSE;
+    try {
+      $this->service_traits->insertTrait($trait);
+    }
+    catch (\Exception $e) {
+      $exception_caught = TRUE;
+    }
+    $this->assertEquals(
```
reynoldtan, line 343: Same comment for line 251

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -176,35 +224,105 @@ public static function provideTermsForSaveTermConfigValuesMethod() {
...
 
+    if ($term_exists) {
+      $cvterm_id = $term_exists;
+    }
+    else {
+      $cvterm = chado_insert_cvterm($input_term, [], $schema = NULL);
+      $cvterm_id = $cvterm->cvterm_id;
+    }
+
+    $is_saved = $this->service_PhenoTerms->saveTermConfigValues([$term_identifier => $cvterm_id]);
```
reynoldtan: Replace chado_insert_cvterm() calls with ChadoCvtermBuddy service to handle creation of a term

trpcultivate_phenotypes/tests/src/Kernel/ServiceTermTest.php
```diff
@@ -176,35 +224,105 @@ public static function provideTermsForSaveTermConfigValuesMethod() {
...
-      'getTermId() should return ' . $expected['term_exists'] . ' for the input indentifier in scenario ' . $scenario
+      $expected['is_saved'],
+      $is_saved,
+      'saveTermConfigValues() should return ' . $expected['is_saved'] . ' in scenario ' . $scenario
+    );
+
+    // Test to see if the configuration value is the input term saved.
+    $this->assertEquals(
+      $this->service_PhenoTerms->getTermId($term_identifier),
+      $cvterm_id,
```
reynoldtan: I think this should be 0, on failed attempt, getTermId() will return 0 instead of the actual cvterm_id

carolyncaron (approved):
The updates made based on my suggestions look great to me! Nothing else is jumping out to me, so I am approving this PR! 🎉

laceysanderson (approved):
I gave this one a code review and it looks great to me. Once the tests finish, this is ready to merge! 🎉

Issue #169, opened 2025-06-12
G0.169 Replace Chado API functions with ChadoBuddy


vidu-wil:
### Branch

g0.169-ReplaceChadoAPIFunctions

### Groups

Group 0 - General

### Describe

In class `TripalCultivatePhenotypesTraitsService` chado API function `chado_insert_cvterm()` is being used and should be updated with ChadoBuddy system.

### Design

- In the method `insertTrait` in class `TripalCultivatePhenotypesTraitsService`, the legacy `chado_insert_cvterm()` is being used to insert the terms. This has to be replaced with the ChadoBuddy system.

PR #170, opened 2025-06-17
G0.169 Replace Chado API functions with ChadoBuddy

vidu-wil:
**Issue #169**

## Motivation

In class TripalCultivatePhenotypesTraitsService chado API function chado_insert_cvterm() is being used and should be updated with ChadoBuddy system.

## What does this PR do?

1. Replaces the legacy `chado_insert_cvterm()` in the `insertTrait()` method of class `TripalCultivatePhenotypesTraitsService` with `upsertCvterm()` method from `ChadoCvtermBuddy` class. 

### Automated Testing

- The Automated testing is complicated for the changes since it's related to mocking the `ChadoBuddyPluginManager`'s `createInstance()` method which similar to the issue mentioned on this comment. https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/164#issuecomment-2959552114

### Manual Testing

2. Open the port on a browser and sign in
3. Create organism in `/bio_data/add/organism`. (Genus: Lens)
4. Configure the genus added in step 3 in `/admin/tripal/extension/tripal-cultivate/phenotypes/ontology`.
5. In your terminal open drush php:cli and execute the following:

```
$genus = 'Lens';

$traits_service = \Drupal::service('trpcultivate_phenotypes.traits');

$traits_service->setTraitGenus($genus);

$trait = 'Trait Name';
$method = 'My Method';
$unit = 'cm';

$trait_combo = [
  'Trait Name' => $trait,
  'Trait Description' => 'Trait Description',
  'Method Short Name' => $method,    
  'Collection Method' => 'Some collection method', 
  'Unit' => $unit,
  'Type' => 'Quantitative'
];

$trait_assets = $traits_service->insertTrait($trait_combo);

$my_combo = $traits_service->getTraitMethodUnitCombo($trait, $method, $unit);
print_r($my_combo);
```


trpcultivate_phenotypes/trpcultivate_phenotypes.services.yml
```diff
@@ -14,4 +14,4 @@ services:
 
   trpcultivate_phenotypes.traits:
     class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTraitsService'
-    arguments: ['@trpcultivate_phenotypes.genus_ontology', '@trpcultivate_phenotypes.terms', '@tripal_chado.database', '@tripal.logger']
+    arguments: ['@trpcultivate_phenotypes.genus_ontology', '@trpcultivate_phenotypes.terms', '@tripal_chado.database', '@tripal_chado.chado_buddy', '@tripal.logger']
```
reynoldtan: @tripal.logger is not being used in the TripalCultivatePhenotypesTraitsService and can be excluded from the arguments list.

reynoldtan:
I reviewed the replacement and performed the manual test.

- [x] Updates did not cause code syntax/formatting Problems
- [x] Used proper injection of service dependencies
- [x] Manual test worked
- [x] Overall test results showed 0 failed item and no unresolved deprecation notices

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -229,20 +257,21 @@ public function insertTrait(array $trait, ?string $schema = NULL) {
...
+          'cvterm.definition' => $values['description'],
         ];
 
-        $ins = chado_insert_cvterm($rec, [], $schema);
+        $ins = $this->cvterm_buddy->upsertCvterm($rec, []);
         if (!$ins) {
           // Could not insert cvterm.
           throw new \Exception('A database error occurred while inserting a term.
-          Failed to insert term ' . $type . ' : ' . $term);
+          Failed to insert term ' . $type . ' : ' . $rec);
```
carolyncaron: This won't print as expected since `$rec` is an array. If $type = 'MyType', then you would see printed:
```
A database error occurred while inserting a term. Failed to insert term MyType : Array
```
I think what you want here is the cvterm name instead of the whole record, in which case, I would replace `$rec` with `$rec['cvterm.name']`.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
```diff
@@ -193,6 +215,12 @@ public function setTraitGenus(string $genus) {
   public function insertTrait(array $trait, ?string $schema = NULL) {
     // Configuration settings of the genus.
     $genus_config = $this->config;
+
+    if ($schema) {
+      $this->cvterm_buddy->connection->setSchemaName($schema);
+      $this->dbxref_buddy->connection->setSchemaName($schema);
+    }
```
carolyncaron, line 222: Would it be possible to get code coverage for this block, or is it also related to issues with mocking ChadoBuddy?

carolyncaron (changes requested):

carolyncaron (changes requested):
Overall, this looks great! My manual test worked as expected. Since I used PHP Execute in my browser, I verified that the trait was in fact inserted by executing the following at the end:
```
$my_combo = $traits_service->getTraitMethodUnitCombo($trait, $method, $unit);
print_r($my_combo);
```
Screenshot 2025-06-19 at 10 59 25 AM

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
laceysanderson: We should be able to get test coverage for this line by adding the test schema as a parameter when we are calling insertTrait() in the tests which shouldn't be too hard. Note: the test chado schema name is available at any point after creating it using the `$this->testSchemaName` variable which is set by the chado test trait.

trpcultivate_phenotypes/tests/src/Kernel/Services/Traits/ValidTraitTest.php
```diff
@@ -206,6 +188,34 @@ public function testTraitsServiceDatabaseExpectations() {
...
+    $unit = 'cm';
+
+    $trait_combo = [
+      'Trait Name' => $trait,
+      'Trait Description' => 'Trait Description',
+      'Method Short Name' => $method,
+      'Collection Method' => 'Some collection method',
+      'Unit' => $unit,
+      'Type' => 'Quantitative',
+    ];
```
reynoldtan:
```suggestion
    // A trait to insert.
    $trait_combo = [
      'Trait Name' => 'My Trait Name',
      'Trait Description' => 'Trait Description',
      'Method Short Name' => 'My Method',
      'Collection Method' => 'Some collection method',
      'Unit' => 'cm',
      'Type' => 'Quantitative',
    ];
```

Variables can be skipped and $trait_combo array can be made simpler.

trpcultivate_phenotypes/trpcultivate_phenotypes.services.yml
```diff
@@ -14,4 +14,4 @@ services:
 
   trpcultivate_phenotypes.traits:
     class: 'Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTraitsService'
-    arguments: ['@trpcultivate_phenotypes.genus_ontology', '@trpcultivate_phenotypes.terms', '@tripal_chado.database', '@tripal.logger']
+    arguments: ['@trpcultivate_phenotypes.genus_ontology', '@trpcultivate_phenotypes.terms', '@tripal_chado.database', '@tripal_chado.chado_buddy', ]
```
reynoldtan:
```suggestion
    arguments: ['@trpcultivate_phenotypes.genus_ontology', '@trpcultivate_phenotypes.terms', '@tripal_chado.database', '@tripal_chado.chado_buddy']
```
The final comma symbol can be safely left out.

trpcultivate_phenotypes/src/Service/TripalCultivatePhenotypesTraitsService.php
carolyncaron: While commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/170/commits/c9ad39e80a56d08822947fe9146c34acc21ffced does offer code coverage, the test itself doesn't actually asserting anything, and we would want to assert that the trait has been inserted correctly.

However, if we add all the needed asserts, this test would essentially look the same as `testTraitsServiceDatabaseExpectations()`, with the only difference being the $schema parameter added. Both tests are actually testing with the same schema, because it's rather difficult to specify 2 different schemas with the current way the test is set up.

Because these are older tests and we are otherwise very confident in Chado's ability to switch between schemas (thanks to my discussion with @laceysanderson), we think the solution can be simplified by just adding the $schema parameter to the `insertTrait()` call within `testTraitsServiceDatabaseExpectations()`. This will give us our needed code coverage while still verifying that our trait was inserted properly. Additionally, it's more performant as we aren't adding a new test to the class.

carolyncaron (approved):
I confirmed that my comments have all been addressed. This looks great to me!

reynoldtan (approved):
I have verified that all my comments have been addressed

PR #185, opened 2025-08-20
G4.185 Implement GermplasmNameExists validator into Share Importer

### LINKED ISSUE #184: G4.184 Setup Germplasm Name Exists validator to share importer
### Branch

g4.184-ImplementeGermplasmNameExists

### Groups

### Describe

Implement Germplasm Name Exists Validator in Share Importer

https://github.com/TripalCultivate/TripalCultivate/blob/e5b7ffe2cd701fd88a1e26be8740b14cb9203e2f/src/Plugin/Validators/GermplasmNameExists.php#L22

### Design

- [x] Setup Germplasm Name Exists validator instance in configure validator method.
- [x] Implement static message processor and define metadata requirements in process validation message method.
- [x] Create test case to test implementation.

reynoldtan:
**Issue #184** Implement Germplasm Name Exists validator in Share Importer.

## What does this PR do?

- [x] Create instance of Germplasm Name Exists validator in configureValidators() method.
- [x] Setup germplasm_name_exists process validation message array element.
- [x] Call Germplasm Name Exists static message processor.
- [x] Update ShareImporterFormValidateTest to include Germplasm Name Exists Validator.
- [x] Create a test file fixture used in a scenario to test germplasm name.

### Automated Testing

Ensure continued success of Kernel tests
1. ShareImporterFormTest
2. ShareImporterFormValidateTest

### Manual Testing

1. In a fully setup clone, create the following test input values.
- A research experiment (project table)
- An organism (organism table)
- A germplasm (stock table)

2. Configure the genus at my_clone/admin/tripal/extension/tripal-cultivate/phenotypes/ontology

3. Pair the project and genus.
```
// Copy and paste into devel/php.   
  ->setGenusToProject(PROJECT ID IN STEP 1, GENUS IN STEP 1);
```

4. Import the test file below using the importer page at my_clone/admin/tripal/loaders/trpcultivate-phenotypes-share-importer

- Set the Research Experiment to the experiment created in Step 1
- Set Genus to the genus created in step 1

[reference_to_non_existent_germplasm.tsv](https://github.com/user-attachments/files/21943986/reference_to_non_existent_germplasm.tsv)

Screenshot 2025-08-22 at 1 40 54 PM

Upon upload of the test file, the importer will trigger a Germplasm Name Exists validator error. Modify the file
and provide the germplasm name created in STEP 1 and re-upload the file to successfully import a file.


trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoShareImporter.php
```diff
@@ -354,6 +355,11 @@ public function configureValidators(array $form_values, string $file_mime_type)
     $instance->setIndices($indices);
     $validators['data-row']['empty_cell'] = $instance;
 
+    $instance = $this->service_validatorPluginManager->createInstance('germplasm_name_exists');
+    $instance->setIndices([0]);
```
carolyncaron:
```suggestion
    $instance->setIndices($header_index['Germplasm Name']);
```

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoShareImporter.php
```diff
@@ -354,6 +355,11 @@ public function configureValidators(array $form_values, string $file_mime_type)
     $instance->setIndices($indices);
     $validators['data-row']['empty_cell'] = $instance;
```
carolyncaron: Should have a comment here to describe why the validator is needed. How about:
```suggestion
    // - All germplasm names exist in the database
```

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoShareImporter.php
```diff
@@ -998,6 +1004,11 @@ public function processValidationMessages($failures) {
         'status' => 'todo',
         'details' => '',
       ],
+      'germplasm_name_exists' => [
+        'title' => 'Germplasm Name exists in the database',
```
carolyncaron: The file might have one or multiple germplasm
```suggestion
        'title' => 'Germplasm exist(s) in the database',
```

trpcultivate_phenoshare/tests/src/Kernel/TripalImporter/ShareImporterFormValidateTest.php
```diff
@@ -534,12 +559,41 @@ public static function provideFormInputValues() {
...
             'details' => 'The following line number and column header combinations were empty, but a value is required.',
           ],
+          'germplasm_name_exists' => ['status' => 'fail'],
+        ],
+        0,
+        FALSE,
+      ],
+
+      // #11: Contains correct header and one line of correct data.
+      // 1st line has reference to a non-existent germplasm name.
```
carolyncaron:
```suggestion
      // #11: 1st line has reference to a non-existent germplasm name.
```

carolyncaron (changes requested):
In the manual test, I confirmed that a non-existent germplasm triggered a validation error, whereas adding an existing one to the test file allowed validation to pass. ✅ 

I made a few small suggestions, mostly to keep consistent with the other validators and the Traits Importer.

trpcultivate_phenoshare/src/Plugin/TripalImporter/TripalCultivatePhenoShareImporter.php
reynoldtan: The suggestion failed the test due to setIndices() expects an array and we are passing an integer. I have modified the suggestion and made it similar to how empty_cell sets indices.

trpcultivate_phenoshare/tests/src/Kernel/TripalImporter/ShareImporterFormValidateTest.php
```diff
@@ -534,12 +559,40 @@ public static function provideFormInputValues() {
...
+          ],
+        ],
+        [
+          'project_genus_match' => ['status' => 'pass'],
+          'valid_data_file' => ['status' => 'pass'],
+          'valid_delimited_file' => ['status' => 'pass'],
+          'valid_header' => ['status' => 'pass'],
+          'empty_cell' => ['status' => 'pass'],
+          'germplasm_name_exists' => [
+            'title' => 'Germplasm Name exists in the database',
```
carolyncaron: This line does not match my suggested change to the title set by the importer ("Germplasm exist(s) in the database"). Since this test didn't fail, I think there needs to be an assert added to check for the title!

carolyncaron (changes requested):
Thanks @reynoldtan! One last thing jumps out to me - I expected tests to fail when you committed my change to the title for this instance of GermplasmNameExists, but it didn't. It looks like we're missing an assert 😅

trpcultivate_phenoshare/tests/src/Kernel/TripalImporter/ShareImporterFormValidateTest.php
reynoldtan: Good catch! it exposed another outdated title text 👍. Fix is in this commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/185/commits/f758fc1c3327288351acf69a0d6611e7214f1ba5

carolyncaron (approved):
Looks great, all of my suggestions have been resolved! 👍

laceysanderson (approved):
Code review looks good ✅

Issue #192, opened 2025-10-20
G0.192 Improve grouping of methods of a trait (DO NOT CLOSE)


reynoldtan:
### Branch

g0.192 - ImproveMethodGrouping

### Groups

Group 0 - General

### Describe

<img width="2890" height="1656" alt="Image" src="https://github.com/user-attachments/assets/7271cf53-2be8-4dc6-b786-c4f2d7ae0b77" />

The case where someone has measured the same trait in different ways is a big thing to watch out for as someone using or curating the data so ideally it would be really clear that this trait was measured with two different methods.

Sort of like we discussed with the add form grouping multiple methods within the same box. That said, I can see how this would be a hard design to get right so maybe we open a new issue that it would be good to make it more obvious in the listing. That way, you can move on and maybe the right design will occur to you as you work on the add trait combo form.

reynoldtan:
Here's an Idea - a mix of tag, icons, number and help text. Tag the trait at the beginning with the number of methods that it supports. The icon next to it matches the icon in every instance of a method and numbered both to indicate part of the group.
A hover text indicate that a method is a collection method of a trait.

<img width="851" height="687" alt="Image" src="https://github.com/user-attachments/assets/d7183c81-9a1c-489b-8827-54f618115264" />

reynoldtan:
Here's another idea - a much cleaner approach without introducing additional icon.

<img width="855" height="689" alt="Image" src="https://github.com/user-attachments/assets/11fffa67-dffd-4fd8-9a04-88a7f2e05efd" />

reynoldtan:
![Image](https://github.com/user-attachments/assets/0c4292ab-b6e3-4b24-bb8e-be24f1817246)

reynoldtan:
This is a suggestion made by @vidu-wil 

> 💬 I just have one suggestion regarding displaying the trait on the table. Would it be possible to introduce a subsection for each Genus, display the genus name as a header, and then list all traits that belong to that genus under that section?

<img width="1331" height="1653" alt="Image" src="https://github.com/user-attachments/assets/ebb0ea46-6dd8-4d7f-b2b0-4bff1d9167fa" />

carolyncaron:
> This is a suggestion made by [@vidu-wil](https://github.com/vidu-wil)
> 
[...]

PR #193, opened 2025-10-21
G0.193 Setup experiment trait picker form

### LINKED ISSUE #183: G0.183 Create a project/experiment configuration page
### Branch

g0.183-ProjectConfigPage

### Groups

Group 0 - General

### Describe

A configuration page is required to allow admin to define a set of traits that a "Research Experiment" will collect data for. The form should be available on each "Research Experiment" content type page in its own tab. The traits are selected from existing traits provided by the Traits Importer.

The workflow for a user would be as follows:
1. Create their research experiment by going to Content > Tripal Content > Add Tripal Content > Research Experiment.
2. Upon submission they will be redirected to the page for their specific experiment which lists all that they entered when creating it. This info is all from the Research Experiment Content Add form and fields.
3. This issue will add a "Phenotypes" tab at the top of the page accessed in step 2. They will click on this tab to configure the phenotypes their experiment is measuring. This tab will take them to the configuration form created in this issue.
4. They will select existing trait/method/units that align with each phenotype they are measuring and assign an experiment-specific column header label for this combo.
5. The combos they configure will be listed at the top of the configuration page and be includes on the "Research Experiment" view page from step 2.
6. These configurations will also be used when validating data imported via the collect and share importers.

### Discussion

- How do we encourage them to use existing trait/method/units while also letting them know they can create their own if the measurements they are doing are not yet described in the site?
- If they go to import data and have not yet configured phenotypes, how do we handle this? Do we embed this form in the importers or just link to it and require it be filled out before continuing to the next step?
- Where are we saving these configurations? The trait/method/units are in the chado.cvterm table, the experiment is in the chado.project table. Is the label and combo a chado.projectprop? or should it be stored in a drupal table? or a chado custom table?

### Design

- [ ] Setup Phenotypes tab in project content type

Image

- [ ] Create configuration form. The add new section in the following drawing is based on the Tripal3 Raw Phenotypes importer "Describe" stage.

Image

- [ ] Others (ie. permissions)
Access permission to the Phenotypes tab is similar to a user with access rights to Edit tab.

reynoldtan:
**Issue #183 Create Experiment Configuration Page**

## Motivation

This is the second phase of the "Create Experiment Configuration Page" task. This PR introduces the Trait Picker window that enables selection of traits and assign them to experiments.

Screenshot 2025-11-03 at 8 17 59 AM

## What does this PR do?

1. Setup trait picker window.
2. Add remove handler to remove icon in the trait listing.

### Automated Testing

PhenoExperimentTraitPickerFormTest
- Test getFormId() method.
- Test buildForm() method - make sure that fields have the correct render array key - value pairs.
- Test matchTrait() AJAX callback method - test the callback method that will list traits matching a keyword.
- Test page exception scenarios.

PhenoExperimentTraitHandlerControllerTest
- Test AJAX request to assign a trait to experiment.
- Test AJAX request to remove a trait from experiment.
- Test handler exception cases.

### Manual Testing

Switch to this branch - **g0.183-ProjectConfigPageAddTrait**

1. Create the following organisms.
Content > Tripal Content > Add Tripal Content > Organism
(**my_site/bio_data/add/organism**)
- Lens culinaris
- Triticum durum

2. Configure the genus (set Trait, Method, Unit and DB) so that it is recognized by the Phenotypes Module.
Tripal > Extensions > TripalCultivate Phenotypes Configuration > Ontology Terms
(**my_site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology**)

3. Create a Research Experiment Tripal Content Type.
Content > Tripal Content > Add Tripal Content > Research Experiment
(**my_site/bio_data/add/research_experiment**)

Provide an experiment name and add the 2 organisms created in step 1 (provide only the genus into the field) into the Germplasm Genus field found in Design tab (use the button Add another item to add a second field). Click Save.

image

4. Upload traits. Run Tripal job for each genus.
Tripal > Data Loaders > Tripal Cultivate: Phenotypic Trait Importer
(**my_site/admin/tripal/loaders/trpcultivate-phenotypes-traits-importer**)

[Traits.zip](https://github.com/user-attachments/files/23304608/Traits.zip)

5. Access the Research Experiment.
Content > Tripal Content > The name of the Research Experiment > Phenotypes tab.

image

6. Click image

7. If the trait picker is launched without a specified genus from the trait listing page, the window will require to select a genus. Use the search field to search specific trait, or use the "Show All Traits" link to display all available trait for a genus.
Screenshot 2025-11-03 at 8 18 47 AM

**NOTE**: The "Suggest a Trait" link currently uses **mailto:** implementation. Further discussion is need to determine a better approach.


carolyncaron (commented):
The following is my review from doing a manual test. Overall, I found it very intuitive to use! The functionality seems to work well and the AJAX is smoooooth. 💯 I have just a few comments:
- I feel that the "Add Trait" pop-up is quite small on the screen and not utilizing the space very effectively. I know this is largely based on resolution of the monitor though but it would nice to adapt the size to take up more of the screen if possible. 🙈 
image

- I LOVE the detail where selecting a genus on the view page before clicking "Add Trait" pre-selects that genus for you when adding traits. I also noticed that once a genus is selected, you can no longer use that dropdown again unless you exit the pop-up window and reopen it. Is this by design or a limitation of AJAX? It would be nice to be able to switch back and forth between genuses if desired (though not crucial to be honest, since more than one genus for a research experiment is expected to be uncommon).
image

- I'm not sure that the form field for adding a label is clear enough about what its looking for. Perhaps we should include a short description above the traits (right below the search) guiding the user to come up with a good label and give a couple of examples. I also like that it tried to suggest a label for each trait, but in my opinion they are currently too long for what a researcher would find reasonable as column headers in their data. Instead of suggesting both "Trait Name + Method Short Name" I would stick with only Trait Name here.
image

- The tooltip should also mention that the point of the label is to make it the column header of the file - I don't think that currently comes across. How about:
> A short label that will be used as the column header for this Trait-Method-Unit combination in the data collection file for this experiment. This must be unique within this experiment.
 
- Lastly, I agree with finding a better approach for "Suggest a Trait". I clicked on it once and somehow ended up with eight email pop-ups 🤣 
Screenshot 2025-11-06 at 10 18 18 AM
I think I pictured a basic form for adding a custom trait (that an admin can approve maybe), but if we want to keep it very simple in the meantime, we could just direct them to the Phenotype Traits Importer?

reynoldtan:
Thanks @carolyncaron for the suuuuuuuper thorough review 👍 I made a few revisions to address the window size, help text, and the redirect to trait importer in this commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/c1b8a10c57bab2e286c43c56eb0e9bc39f73e4b2

As for the select genus field being disabled, this is intentional and by design to ensure users remain deliberate and aware of the genus to which they are working with.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+   * @var \Drupal\Core\Database\Connection
+   */
+  protected Connection $db;
+
+  /**
+   * Request array.
+   *
+   * @var array
+   */
+  private array $request;
```
laceysanderson: We should be very careful when using private scope! I found private scope caused issues with dependency injection and serialized objects and had to remove it in a number of places in Tripal core. Additionally it can make testing more difficult then needed since private scoped variables will not be accessible in mock objects.

We should only use it if there is a very good reason we don't want any extending classes to touch a variable. The only case I can think of is related to sensitive data and security.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+   *
+   * @param Symfony\Component\HttpFoundation\Request $request
+   *   Request.
+   * @param string $action
+   *   The requested action - assign or remove a trait.
+   *
+   * @return Symfony\Component\HttpFoundation\JsonResponse
+   *   JSON response array message and status code.
+   */
+  public function handleAction(Request $request, string $action) {
```
laceysanderson: I think these docs could use some expanded to make how this is happening more clear. Maybe something like:


```suggestion
   * Handle AJAX request to assign/remove trait to/from experiment.
   *
   * The "trait" and "experiment" this action will apply to is encoded in the
   * request object and will be saved to self::request.
   *
   * @param Symfony\Component\HttpFoundation\Request $request
   *   Request.
   * @param string $action
   *   The requested action to be peformed for a trait. Supported actions are:
   *   - assign: assigns the trait to the experiment.
   *   - remove: removes the trait from the experiment.
   *
   * @return Symfony\Component\HttpFoundation\JsonResponse
   *   JSON response array message and status code.
   */
  public function handleAction(Request $request, string $action) {
```

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+   * Drupal database connection.
+   *
+   * @var \Drupal\Core\Database\Connection
+   */
+  protected Connection $db;
+
+  /**
+   * Request array.
+   *
+   * @var array
```
laceysanderson: Expand this docblock to indicate that this is the data from the current request and document the keys that we expect and what they are. For example, 

The following keys are expected:

- project (int): the project_id for the experiment the action should be applied to.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+    if (!in_array($action, array_keys(self::ACTION_PARAMS))) {
+      throw new AccessDeniedHttpException('Invalid request: Unsupported handler action.');
+    }
+
+    if ($request->getMethod() != 'POST') {
+      throw new AccessDeniedHttpException('Invalid request: Unsupported request method.');
+    }
+
+    $this->request = $request->request->all();
+    $request_params_keys = is_null($this->request) ? [] : array_keys($this->request);
```
laceysanderson: Add a comment documenting what these lines are doing so they stand out from the validation checks and so that the developer can follow the logic. Also, add an @see to indicate where they can find documentation on the expected keys.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+      }
+    }
+
+    foreach ($this->request as $value) {
+      if (trim($value) == '') {
+        throw new AccessDeniedHttpException('Invalid request: Invalid data provided.');
+      }
+    }
+
+    $method = $action . 'Trait';
```
laceysanderson: Why not make the action = `assignTrait` to begin with? What would make it easier to support other configuration actions in the future that may be related to non-trait parts like the combo label or method/unit.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * Handle trait callback: assign a trait.
+   *
+   * @return array
+   *   An array with the following key:
+   *   - 'message': the relevant status message.
+   *   - 'status_code': the response status code ie. 200 for Ok.
+   */
+  public function assignTrait() {
```
laceysanderson: Document that the trait and experiment being acted on was encoded in the request and is available in self::request. Also indicate which self::request keys specifically this method requires are present.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
```diff
@@ -0,0 +1,257 @@
...
+
+    return [
+      'message' => 'Ok',
+      'status_code' => self::GOOD_REQUEST,
+    ];
+  }
+
+  /**
+   * Handle trait callback: remove a trait.
+   *
```
laceysanderson: Same as above, please document where the experiment and trait info is and what keys from self::request are required by this method. Especially since the combo_id is required to fit a very specific format!

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -280,7 +283,19 @@ public function buildForm(array $form, FormStateInterface $form_state) {
       }
 
       // Use the trait status to disable the remove option.
-      $remove = 'x';
+      $remove = [
+        '#type' => 'link',
+        '#title' => '',
+        '#url' => Url::fromUri('internal:#'),
```
laceysanderson: Drupal ethos is to never make core functionality javascript only in case someone has it turned off. In this case, that means that we want this URL to point to a route and we want the javascript to use that route.

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -318,7 +333,9 @@ public function buildForm(array $form, FormStateInterface $form_state) {
               ],
             ],
           ],
-          $remove,
+          [
+            'data' => $remove,
```
laceysanderson: A comment here about what is in remove at this point would be helpful as I am having trouble following the logic. In the controller I see that the `combo_id` should be in the request data but reading the code here it looks like $remove is either `-` or a full on link render array that contains data-combo_id as an attribute.

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
```diff
@@ -0,0 +1,452 @@
...
+   * @var \Drupal\Core\Routing\RouteMatchInterface
+   */
+  protected RouteMatchInterface $service_RouteMatch;
+
+  /**
+   * The genus used to filter the traits table and show only related traits.
+   *
+   * @var string
+   */
+  private string $filter_genus;
```
laceysanderson: Does this really need to be private?

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
```diff
@@ -0,0 +1,452 @@
...
+        ->load($tripal_entity_id);
+    }
+    else {
+      $research_experiment = $this->service_RouteMatch
+        ->getParameter('tripal_entity');
+
+      if (!$research_experiment->id()) {
+        // Research experiment entity does not exist.
+        throw new NotFoundHttpException();
+      }
```
laceysanderson: I think this should be outside the if/else shouldn't it? Don't we also want an exception if the id is in the form state and we are unable to load it via the entity type manager? This should also log a watchdog error letting the admin know that someone tried to configure traits for an experiment we were unable to load!

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,452 @@
...
+      }
+
+      $form_state->set('tripal_entity_id', $research_experiment->id());
+    }
+
+    $experiment = $research_experiment->get('exp_name')->getValue();
+    if (!$experiment) {
+      // Research experiment entity does not exist contain exp_name field.
+      throw new NotFoundHttpException();
+    }
```
laceysanderson: This should also log a watchdog error so that the admin knows that we expect there to be an exp_name field!

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
```diff
@@ -0,0 +1,452 @@
...
+    }
+
+    $experiment = $research_experiment->get('exp_name')->getValue();
+    if (!$experiment) {
+      // Research experiment entity does not exist contain exp_name field.
+      throw new NotFoundHttpException();
+    }
+
+    $project_id = (int) $experiment[0]['record_id'];
+    $form_state->set('project_id', $project_id);
```
laceysanderson: I think this should be saved via a hidden form element (i.e. #type => value) rather then set directly to the form state. That forces it to be available to validation, etc.

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
```diff
@@ -0,0 +1,452 @@
...
+      ],
+    ];
+
+    // This settings array is used to construct AJAX request parameters.
+    $form['#attached']['drupalSettings']['tcpCombo'] = [
+      'route' => 'bio_data/experiment/handle_trait/assign',
+      'project' => $project_id,
+      'genus' => $genus,
+      'user' => $this->currentUser()->id(),
+    ];
```
laceysanderson: This should work via the form submit alone without javascript needing to be triggered.

laceysanderson (dismissed):
Wow this is the most javascript heavy form I have ever seen :see_no_evil: I think it's really important we stick to the form api to process everything and javascript is just used to limit page reloads. Otherwise there is a chance we are bypassing important drupal form api security, etc.

reynoldtan:
This update introduces revisions to Phenotypes Experiment Configuration form suite, with key enhancements focused on these areas.

- Removal of custom CSS and Javascript libraries.
- Unnecessary routing.
- Additional functionality, drop down buttons, and toolbars.
- Use of render array and properties #state and #ajax to manage form behaviours.

and other with the goal of strengthening the overall codebase and ensure it is solid, robust and compliant to Drupal framework.

Below are sources of test data used in the following screenshots:

Organism: Lens _culinaris_ and Triticum _durum_

**Lens Traits:** 
[sample traits-Lens.tsv.zip](https://github.com/user-attachments/files/24483757/sample.traits-Lens.tsv.zip)

**Triticum Traits:**
[sample traits-Triticum.tsv.zip](https://github.com/user-attachments/files/24483759/sample.traits-Triticum.tsv.zip)


**Summary Table**
image

**Trait Selector**
image

reynoldtan:
Please don't invest significant time reviewing the current summary table, as a more refined version is on the way 

image


- https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/192

vidu-wil:
I did the manual testing for this and it looks great! 👍 All the functionality mentioned in the description works as expected. 
Screenshot 2026-01-26 at 2 43 40 PM

✅ I see that the pop-window size issue that Carolyn mentioned above has been fixed and the pop-up window is adapting according to the resolution and window size now.
✅ Also, I like the fact that "Suggest a Trait" is directing to the "Tripal Cultivate: Phenotypic Trait Importer" rather than linking to mailto: option.
❔I just have a minor suggestion about the wording of the "Close & Update Traits" button. I think it makes more sense to have it worded as "Update Traits & Close" or "Update and Close" instead.
Screenshot 2026-01-26 at 3 38 02 PM

✅ The summary page shows everything as expected and I tried setting some items as required and the drop down next to "Remove" on each trait shows either "Set Require" or "Set Optional" correctly and adapts accordingly after changing.
Screenshot 2026-01-26 at 3 02 59 PM

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -334,7 +484,123 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
+          $this->tripal_logger->error($msg = 'Invalid request: Failed to update trait status.');
+          $this->messenger()->addStatus($msg);
+        }
+
+        break;
+    }
+
+    if ($ok === 1) {
+      $this->messenger()
+        ->addStatus('The trait combo operation ' . ucfirst($action) . ' completed successfully.');
```
vidu-wil:
```suggestion
        ->addStatus("The trait combo operation '" . ucfirst($action) . "' completed successfully.");
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+   *   Drupal database connection.
+   * @param \Drupal\tripal_chado\Database\ChadoConnection $chado_connection
+   *   The connection to the Chado database.
+   * @param \Drupal\tripal\Services\TripalLogger $tripal_logger
+   *   Tripal logger service.
+   * @param \Drupal\trpcultivate_phenotypes\Service\TripalCultiavtePhenotypesGenusOntologyService $service_PhenoGenusOntology
+   *   TripalCultivate Phenotypes Genus-Ontology.
+   * @param \Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesGenusProjectService $service_PhenoGenusProject
+   *   TripalCultivate Phenotypes Genus-Project service.
+   * @param \Drupal\trpcultivate_phenotypes\Service\ripalCultivatePhenotypesTraitsService $service_PhenoTraits
```
vidu-wil:
```suggestion
   * @param \Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTraitsService $service_PhenoTraits
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+   */
+  public function buildForm(array $form, FormStateInterface $form_state) {
+
+    $tripal_entity = $this->getRouteMatch()->getParameter('tripal_entity');
+
+    // Route validates instance of tripal_entity and presence of exp_name field.
+    $experiment = $tripal_entity->get('exp_name')->getValue();
+    ['record_id' => $experiment_id, 'value' => $experiment_name] = $experiment[0];
+
+    // Update the title to show which reseach experiment is being configured.
```
vidu-wil:
```suggestion
    // Update the title to show which research experiment is being configured.
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+      '#id' => 'tcp-trait',
+    ];
+
+    $form[$form_dialog_wrapper]['search_tooltips'] = [
+      '#type' => 'container',
+      '#children' => [
+        'a_tip' => [
+          '#type' => 'html_tag',
+          '#tag' => 'p',
+          '#value' => 'Start typing part of the trait name into the search field to search for specific traits, or click - Show all Traits button, to explore all available traits for the selected genus.',
```
vidu-wil:
```suggestion
          '#value' => "Start typing part of the trait name into the search field to search for specific traits, or click 'Show all Traits' button, to explore all available traits for the selected genus.",
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+        'invisible' => [
+          ':input[name="genus"]' => ['value' => 0],
+        ],
+      ],
+    ];
+
+    // Capture trigger element responsible for staring trait selection request.
+    $trigger_el = $form_state->getTriggeringElement() ?? 0;
+
+    // Prepare tratis that matched the search key.
```
vidu-wil:
```suggestion
    // Prepare traits that matched the search key.
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+          ->condition('tc.label', $label, '=')
+          ->condition('tc.project_id', $experiment_id, '=')
+          ->execute()
+          ->fetchField();
+
+        if ($label_exists) {
+          $form[$form_dialog_wrapper]['duplicate_label'] = [
+            '#theme' => 'status_messages',
+            '#message_list' => [
+              'warning' => [
```
vidu-wil:
```suggestion
              'error' => [
```
I believe this should be an error instead of warning. (The '#status_headings' below is defined as an 'error' and the heading does not appear due to this)
Screenshot 2026-01-27 at 11 39 14 AM

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -272,60 +386,96 @@ public function buildForm(array $form, FormStateInterface $form_state) {
...
-      $rows[] = [
-        'data' => [
-          [
-            'data' => [
-              '#markup' => ($this->filter_genus) ? $trait_row->label : $trait_row->label . '<br /><small>' . $genus . '</small>',
+      // Table column: trait combo.
+      $form[$summary_table_name][$i]['trait_combo'] = [
+        '#type' => 'component',
+        '#component' => 'trpcultivate_phenotypes:trait_combo',
+        '#slots' => [],
```
vidu-wil:
```suggestion
```
'slots' has been removed from trait_combo.component.yml

trpcultivate_phenotypes/tests/src/Kernel/Forms/PhenoExperimentTraitSelectorFormTest.php
```diff
@@ -0,0 +1,894 @@
...
+        ],
+        [
+          'message' => '',
+        ],
+      ],
+    ];
+  }
+
+  /**
+   * Test valid reoute parameter and page exceptions.
```
vidu-wil:
```suggestion
   * Test valid route parameter and page exceptions.
```

trpcultivate_phenotypes/tests/src/Kernel/Forms/PhenoExperimentConfigurationFormTest.php
```diff
@@ -736,114 +885,54 @@ public function testPageAccess(string $scenario, array $user, array $expected) {
...
-        [
-          'message' => '',
-        ],
-      ],
     ];
   }
 
   /**
-   * Test page exceptions.
+   * Test valid reoute parameter and page exceptions.
```
vidu-wil:
```suggestion
   * Test valid route parameter and page exceptions.
```

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
reynoldtan: good catch 👍

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
reynoldtan: great catch, here's an example
image

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
reynoldtan: Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/a93bf15ccacb3678842234c22f1c8172dcf43a6e. I flipped the quotes so that the text to be highlighted is in double quotes instead of single, as it is familiar to user to denote important string

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
reynoldtan: Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/a93bf15ccacb3678842234c22f1c8172dcf43a6e. I flipped the quotes so that the text to be highlighted is in double quotes instead of single, as it is familiar to user to denote important string

reynoldtan:
> ❔I just have a minor suggestion about the wording of the "Close & Update Traits" button. I think it makes more sense to have it worded as "Update Traits & Close" or "Update and Close" instead.

I set the label to "Close and Update Traits" to reflect the actual sequence of events triggered when the button is clicked - window closes first, then as second event the summary table reloads (update traits). I initially tried the alternative "Update & Close", but found it ambiguous, while still on the selector form, I kept questioning what was being updated. It is good you highlighted this as it can now serve a focus point for other reviewers to ensure interface is clear.

ruobinLiu:
Everything worked as expected! Nice work Reynold!

Except I observed some minor differences during my test than the screenshots, have them both circled in red:

**Step 5**
Seems like I can add trait by clicking on the "Add Trait" botton
543112069-715fc2a8-38a4-4025-9bd6-192da75bc338

**Step 7**
There is a toggle that could allow me to disable a trait?
543112726-4f0cdbdb-9c6b-4479-b36a-e6b617fcb0cf

reynoldtan:
Thanks for the review @ruobinLiu 👍 

> Everything worked as expected! Nice work Reynold!
> 
> Except I observed some minor differences during my test than the screenshots, have them both circled in red:
> 
> **Step 5** Seems like I can add trait by clicking on the "Add Trait" botton 543112069-715fc2a8-38a4-4025-9bd6-192da75bc338
> 

I added a quick link, with the same action as the main + Add Trait button, if there is not trait combo in the summary table as an alternative way to add trait.

> **Step 7** There is a toggle that could allow me to disable a trait? 543112726-4f0cdbdb-9c6b-4479-b36a-e6b617fcb0cf

The toggle/checkbox field is to indicate that the combo your adding to the experiment, is going to be a required trait (it must exists in the data file).

Tip: if you hover on the star ⭐  icon, it will show a brief description.
Screen Shot 2026-02-02 at 10 36 37 AM

Adding the trait with the toggle on/checked will attach the same star ⭐  icon in the summary table.
Screen Shot 2026-02-02 at 10 39 18 AM

carolyncaron:
> > ❔I just have a minor suggestion about the wording of the "Close & Update Traits" button. I think it makes more sense to have it worded as "Update Traits & Close" or "Update and Close" instead.
> 
> I set the label to "Close and Update Traits" to reflect the actual sequence of events triggered when the button is clicked - window closes first, then as second event the summary table reloads (update traits). I initially tried the alternative "Update & Close", but found it ambiguous, while still on the selector form, I kept questioning what was being updated. It is good you highlighted this as it can now serve a focus point for other reviewers to ensure interface is clear.

How about "Close and Refresh Table"? I think the "Update" word is confusing because it makes it sound like it's "Saving" the traits, but in practice, they are already saved when the "Add" button is clicked. Plus, "Refresh Table" is already used on the summary page, so this would be keep it consistent.

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -2,29 +2,47 @@
...
+ * Class definition of Experiment Phenotypes Configuration page.
  */
 class PhenoExperimentConfigurationForm extends FormBase {
 
+  /**
+   * Drupal database connection.
+   *
+   * @var \Drupal\Core\Database\Connection
+   */
+  protected Connection $database_connection;
```
carolyncaron: I recommend calling this `$drupal_connection` instead to keep consistency with Tripal and make it explicit what type of database connection this is.

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
```diff
@@ -2,29 +2,47 @@
...
    * @var \Drupal\tripal_chado\Database\ChadoConnection
    */
   protected ChadoConnection $chado_connection;
 
+  /**
+   * Tripal logger service.
+   *
+   * @var \Drupal\tripal\Services\TripalLogger
+   */
+  protected $tripal_logger;
```
carolyncaron:
```suggestion
  protected TripalLogger $tripal_logger;
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+ * Class definition of Experiment Add Trait.
+ */
+class PhenoExperimentTraitSelectorForm extends FormBase {
+
+  /**
+   * Drupal database connection.
+   *
+   * @var \Drupal\Core\Database\Connection
+   */
+  protected Connection $database_connection;
```
carolyncaron:
```suggestion
  protected Connection $drupal_connection;
```

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
```diff
@@ -0,0 +1,670 @@
...
+   * @var \Drupal\tripal_chado\Database\ChadoConnection
+   */
+  protected ChadoConnection $chado_connection;
+
+  /**
+   * Tripal logger service.
+   *
+   * @var \Drupal\tripal\Services\TripalLogger
+   */
+  protected $tripal_logger;
```
carolyncaron:
```suggestion
  protected TripalLogger $tripal_logger;
```

carolyncaron (changes requested):
Thanks for implementing these changes! I did a quick code review, and here are some things that I noticed during the manual review:

- For my test, I added some traits for Lens but none for Triticum yet. If I filter the summary table for Triticum, it says "No traits found for this research experiment" which isn't true because there are some for Lens. Is it possible to mention the genus being filtered in the error message? Something like "No traits for Triticum found for this research experiment. I acknowledge this is a bit tricky because there is a filter for "All Genus", so the current error message works for that case specifically.

- I think it makes more sense now to rename "Suggest a Trait" to be "Import Traits" instead since it is being redirected to the Traits Importer. In the future, we may want to add the ability to "Suggest a trait" using a form, but that's outside the scope of this PR so I think it's best to rename it to accurately reflect what it does.

- When closing the trait picker form, is it possible to make the "X" (Close) option in the top right of the window perform the same action as "Close & Update Traits" currently does? Right now, it just closes the form and the user might wonder what happened to the traits they just added. 
   Image

- The suggested default for Label is still WAY too long in my opinion. I would just go with the Trait Name since that it most realistic in practice. Perhaps @ruobinLiu could confirm here, but the "Method Short Name" in most cases won't be as brief as what your example files show (ex. PLHT). Regardless, combining the Trait Name and Method name still makes this too long in my opinion.

- This is a very minor visual critique, but it looks odd that the 2 options next to the "Remove" button don't have the same size background when the mouse is overlaid on them. This also occurs when the Trait is already required and the text says "Set Optional".
Set Require: Image
Set Archive: Image

- Related to above, I would change "Set Require", "Set Optional" and "Set Archive" to be "Set as required", "Set as optional", and "Archive Trait", respectively.

reynoldtan:
> * For my test, I added some traits for Lens but none for Triticum yet. If I filter the summary table for Triticum, it says "No traits found for this research experiment" which isn't true because there are some for Lens. Is it possible to mention the genus being filtered in the error message? Something like "No traits for Triticum found for this research experiment. I acknowledge this is a bit tricky because there is a filter for "All Genus", so the current error message works for that case specifically.

Addressed in https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/19ec6e3e2e305d6dae5d697981bf1944d4a03bab

Some examples:
Screen Shot 2026-02-03 at 4 46 19 PM
Screen Shot 2026-02-03 at 4 46 47 PM

reynoldtan:
> * I think it makes more sense now to rename "Suggest a Trait" to be "Import Traits" instead since it is being redirected to the Traits Importer. In the future, we may want to add the ability to "Suggest a trait" using a form, but that's outside the scope of this PR so I think it's best to rename it to accurately reflect what it does.

Addressed in commit: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/32555f1f9f80de1169e8079387a7dd5771786834

Screen Shot 2026-02-03 at 4 50 33 PM

reynoldtan:
> * When closing the trait picker form, is it possible to make the "X" (Close) option in the top right of the window perform the same action as "Close & Update Traits" currently does? Right now, it just closes the form and the user might wonder what happened to the traits they just added.

Addressed in commit: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/7282d6676de075aabde3d2bfa922086c04da795d

reynoldtan:
> * The suggested default for Label is still WAY too long in my opinion. I would just go with the Trait Name since that it most realistic in practice. Perhaps @ruobinLiu could confirm here, but the "Method Short Name" in most cases won't be as brief as what your example files show (ex. PLHT). Regardless, combining the Trait Name and Method name still makes this too long in my opinion.


I see your point on the label length. The reason we combine Trait Name and Method Short Name is to ensure uniqueness. With just the Trait Name, different methods of Plant Height would end up with identical labels - 'Plant Height', and the interface would catch this.

reynoldtan:
> * This is a very minor visual critique, but it looks odd that the 2 options next to the "Remove" button don't have the same size background when the mouse is overlaid on them. This also occurs when the Trait is already required and the text says "Set Optional".

Not sure how to rectify this but I looks like this is admin theme issue also in content type view.
Screenshot 2026-02-04 at 8 01 14 AM

Screenshot 2026-02-04 at 8 00 32 AM

reynoldtan:
> * Related to above, I would change "Set Require", "Set Optional" and "Set Archive" to be "Set as required", "Set as optional", and "Archive Trait", respectively.

Addressed in commit : https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/d6cc18797eb03c8571a34e750314aed02981b17a

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
reynoldtan: Addressed in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/112810af3fb7ba4c11f5feb5d2e2a5d48c041bc5

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
reynoldtan: addressed in commit: https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/112810af3fb7ba4c11f5feb5d2e2a5d48c041bc5

ruobinLiu:
> Thanks for the review @ruobinLiu 👍
> 
> > Everything worked as expected! Nice work Reynold!
> > Except I observed some minor differences during my test than the screenshots, have them both circled in red:
> > **Step 5** Seems like I can add trait by clicking on the "Add Trait" botton 543112069-715fc2a8-38a4-4025-9bd6-192da75bc338
> 
> I added a quick link, with the same action as the main + Add Trait button, if there is not trait combo in the summary table as an alternative way to add trait.
> 
> > **Step 7** There is a toggle that could allow me to disable a trait? 543112726-4f0cdbdb-9c6b-4479-b36a-e6b617fcb0cf
> 
> The toggle/checkbox field is to indicate that the combo your adding to the experiment, is going to be a required trait (it must exists in the data file).
> 
> Tip: if you hover on the star ⭐ icon, it will show a brief description. Screen Shot 2026-02-02 at 10 36 37 AM
> 
> Adding the trait with the toggle on/checked will attach the same star ⭐ icon in the summary table. Screen Shot 2026-02-02 at 10 39 18 AM

Really like the ⭐ icon allowing customized "Required Traits" 👍

ruobinLiu:
Step 2 Configure the genus
Screenshot 2026-02-06 at 3 45 19 PM

I notice the Trait Picker allows new traits to be added. Do I get to describe this new trait with Pgenotypic Trait File later on?

reynoldtan:
> I notice the Trait Picker allows new traits to be added. Do I get to describe this new trait with Pgenotypic Trait File later on?

This option is for traits not defined (or added through the trait picker) and detected in the file upload process, much like the new traits detected in rawphenotypes which required user to describe. This option is not fully implemented in our new phenotypes importer.

trpcultivate_phenotypes/tests/src/Kernel/Forms/PhenoExperimentConfigurationFormTest.php
```diff
@@ -73,16 +81,11 @@ class PhenoExperimentConfigurationFormTest extends ChadoTestKernelBase {
   const ROUTE_NAME = 'trpcultivate_phenotypes.experiment_configuration';
 
   /**
-   * Expected research experiment entity id to describe a setup.
+   * The table name.
```
carolyncaron: Which table? Is this the name of custom table for storing phenotype trait/method/unit combinations?

carolyncaron (changes requested):
Thank you for those changes, the form looks like its just about there!

Just a couple more suggestions:
- Next to "Import Traits", can you add the icon that indicates the link will open in a new tab? I think it's a nice touch when the user knows to expect a new tab will be opened. This is the class I use on RE pages when I link out to JBrowse: `<i class="fa-solid fa-arrow-up-right-from-square"></i>`
- For the status messages when traits are set as required, optional, archived, and restored, I think the message can be more simple and exclude "The trait combo operation". We should avoid using the word "combo" when it is user-facing. 

For example:
Image
This message could read as `"Set as Required" completed successfully.`

reynoldtan:
> * For the status messages when traits are set as required, optional, archived, and restored, I think the message can be more simple and exclude "The trait combo operation". We should avoid using the word "combo" when it is user-facing.

Addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/7ac157d8cf99d56c9e808d00430dbb16ec4f1b31

trpcultivate_phenotypes/tests/src/Kernel/Forms/PhenoExperimentConfigurationFormTest.php
reynoldtan: expanded the table name in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/7ac157d8cf99d56c9e808d00430dbb16ec4f1b31

reynoldtan:
> * Next to "Import Traits", can you add the icon that indicates the link will open in a new tab? I think it's a nice touch when the user knows to expect a new tab will be opened. This is the class I use on RE pages when I link out to JBrowse: `<i class="fa-solid fa-arrow-up-right-from-square"></i>`

image

Addressed in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/193/commits/f2ed04b4c1efd5117b5cc54dbcad239afdc195b9

carolyncaron (approved):
Thanks for those changes! PR approved ✅

reynoldtan:
I have reviewed Lacey's past comments and either it is no longer relevant as the file referenced is no longer part of the codebase or I have addressed it fully. Ready to merge.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: Noted and this item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
reynoldtan: This item is no longer relevant as the property is no longer part of the codebase

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
reynoldtan: Watchdog implemented

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitSelectorForm.php
reynoldtan: Watchdog implemented

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
reynoldtan: This item is no longer relevant as this render array is no longer part of the codebase.

trpcultivate_phenotypes/src/Form/PhenoExperimentTraitPickerForm.php
reynoldtan: This item is no longer relevant as this variable is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Controller/PhenoExperimentTraitHandler.php
reynoldtan: This item is no longer relevant as the referenced file is no longer part of the codebase.

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
reynoldtan: This item is no longer relevant as the array is no longer part of the codebase.

trpcultivate_phenotypes/src/Form/PhenoExperimentConfigurationForm.php
reynoldtan: Noted and this item is no longer relevant as the variable is no longer part of the codebase.

PR #207, opened 2026-01-30
G3.207 - Setup Backups Tab in Experiment Page

### LINKED ISSUE #205: G3.205 Feature requests for phenotype backup
### Branch

g3.205-backupTab

### Groups

Group 3 - Search | Browse | Tools

### Describe

Image

1. It would be nice to have a specific tab for backup directly on the research experiment page.
2. On the phenotypic data backup page, to change the default sort "from recent" to "oldest" or at least make this configurable.

### Design

See above.

reynoldtan:
# Issue #205

## Motivation

This PR will setup a listing of phenotypes data backup specific to the research experiment being viewed. A button to backup data file is provided in the same page as the summary table.

Screen Shot 2026-01-30 at 12 30 46 PM

## What does this PR do?

1. Adds a links task item in the links.task.yml
2. Define a route using the same permission settings as phenotypes configuration
3. Create a controller for the summary table.
4. Setup automated test.

### Automated Testing

ClassName->PhenoExperimentPhenoBackupControllerTest
- Test summary table has the correct header and in the expected order.
- Test that page has the expected page title text.
- Test that each backup detail falls in the correct column header.
- Test that the backup order is showing the latest backup first.

### Manual Testing

1. Switch to branch **g3.205-backupTab**
2. In a fully working clone, create a Research Experiment Tripal Content Type.

**Content/Tripal Content/+ Add Tripal Content/Research Experiment**
or access **my_clone/bio_data/add**

3. Click the Phenotypes Backup tab

Screen Shot 2026-01-30 at 1 17 55 PM

4. Use the in-page Backup Phenotypic Data button to add test backups.
Screen Shot 2026-01-30 at 1 19 54 PM

5. Verify that backups are downloadable and arranged so that the latest backup appears on top.

Note: the button to backup file does not auto-select the current experiment due to required major form changes.

Updates required.
1. Register a route slug or query string variable.
2. Pull request and perform validation and sanitation of the value.
3. Determine the full project, name and project id to replicate the autocomplete field ie. AGILE (1) value.
4. Implement a logic that will determine if slug/query is present and perform steps 2-3.
5. Implement automated tests.


ruobinLiu:
[like]  Liu, Ruobin reacted to your message:
________________________________

ruobinLiu:
Screenshot 2026-02-02 at 4 31 58 PM
Regarding Step 5: the latest backup still appears on bottom for me 🤔

reynoldtan:
> Screenshot 2026-02-02 at 4 31 58 PM Regarding Step 5: the latest backup still appears on bottom for me 🤔

reynoldtan:
> Screenshot 2026-02-02 at 4 31 58 PM Regarding Step 5: the latest backup still appears on bottom for me 🤔

Thanks for the review @ruobinLiu. This screenshot looks like the table part of the backup functionality. Try to switch back to the table where you find the menu options shown

image

ruobinLiu:
> > Screenshot 2026-02-02 at 4 31 58 PM Regarding Step 5: the latest backup still appears on bottom for me 🤔
> 
> Thanks for the review @ruobinLiu. This screenshot looks like the table part of the backup functionality. Try to switch back to the table where you find the menu options shown
> 
> image

Thanks for the instruction Reynold! I find the right table now, yes, my latest record is now sitting on the top 🎉 
Screenshot 2026-02-04 at 11 23 25 AM

ruobinLiu (dismissed):
Looking great!

carolyncaron:
> > Screenshot 2026-02-02 at 4 31 58 PM Regarding Step 5: the latest backup still appears on bottom for me 🤔
> 
> Thanks for the review @ruobinLiu. This screenshot looks like the table part of the backup functionality. Try to switch back to the table where you find the menu options shown
> image

Would it be possible to fix the sort of this table as well in this PR? Ideally, the table would be sorted the same regardless of where we might be accessing it.

reynoldtan:
> Would it be possible to fix the sort of this table as well in this PR? Ideally, the table would be sorted the same regardless of where we might be accessing it.

Sorting fixed:

Sort order in backup page matches order in pheno tab.
Screenshot 2026-03-05 at 9 14 53 AM

Screenshot 2026-03-05 at 9 14 47 AM


Quick note - the tabs are now replaced with a dropdown menu.
Screenshot 2026-03-05 at 9 22 36 AM

ruobinLiu:
Yes, sort order in backup page matches order in pheno tab 😄 great job @reynoldtan 🎉 
Screenshot 2026-03-12 at 11 38 14 AM
Screenshot 2026-03-12 at 11 39 36 AM

ruobinLiu (dismissed):

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+   * Retrieve backups for a research experiment.
+   */
+  public function loadBackups() {
+
+    ['record_id' => $exp_id, 'value' => $exp_name] = $this->service_RouteMatch
+      ->getParameter('tripal_entity')
+      ->get('exp_name')
+      ->getValue()[0];
```
laceysanderson: This block is making two assumptions:
1. That there is a TripalEntity in the Route.
2. That this entity has a field named `exp_name` with the two properties mentioned.

 The first assumption is not too bad since that is enforced by the route but a quick check here would be prudent.

 The second assumption needs to be addressed though as admin can easily remove this field through the UI.

 - Instead I would use the entity title instead of the experiment name in the title for this page as that is garunteed through the Drupal UI.
 - For the chado project_id I thought we had a Tripal service 🤔 but I can't find it. In the meantime it would be best to use the first ChadoField you come across.
 - I would create a `getExperimentId()` helper method for in this class to make it easy to update later as services become available.

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
```diff
@@ -0,0 +1,118 @@
...
+        ->getStorage('file')
+        ->load($backup->get('file_id'));
+
+      if ($file_obj) {
+        $file_download['data'] = [
+          '#type' => 'button',
+          '#value' => 'Download',
+          '#button_type' => 'primary',
+          '#attributes' => [
+            'onClick' => 'window.location.href="' . $file_obj->createFileUrl() . '"; return false;',
```
laceysanderson: This requires javascript for someone to download their file, doesn't it? I think this is why we usually use links styled as buttons. 🙈

laceysanderson: We should update the same thing in the ListBuilder.

laceysanderson (changes requested):
Just those two inline comments regarding the field assumption and javascript dependence. The rest of the code looks very clean / solid. I didn't do a manual review but am extremely excited by the screenshots!

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
reynoldtan: Addressed in the recent commit. For the last item, creating the temporary helper method did not feel worthwhile. For the moment I added a @todo marker to describe our intent and once this service becomes available we can refactor cleanly.

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
reynoldtan: Addressed in both places, in the recent commit.

laceysanderson: Perfect --thanks!

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
laceysanderson: The new approach uses the label to determine the experiment ID. However, this makes the assumption that the title/label is the same as the name which is not always the case as it can be configured using any tokens 🙈 

I think we stick with the original approach of using the `record_id` from a field but I think we use a helper method so we can pick the first required tripal field on the base `project` table. That way we don't make assumptions about field names and we are guaranteed the record of the project table.

I'm actually going to create an issue in Tripal core to make such functionality there as this is a common need. See https://github.com/tripal/tripal/issues/2445.

laceysanderson: Using the label in the title is perfect though! Even better then using the experiment name in my opinion.

reynoldtan: The Tripal entity getter method works like a charm 👍 addressed in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/207/commits/93ff911b935612b4709cb90a32555fff74ccdd97

carolyncaron (approved):
The functionality works flawlessly for me! ✅ I confirmed that backups are now sorted in descending order regardless if they're accessed via an experiment or via the main backups page.

It's worth noting that new changes in Drupal 11.4-dev removed the tabs we normally see on pages 😢 So now, the way to navigate to backups from a research experiment is via settings in the top right:
Image

This unfortunately makes the tabs that appear when navigating to backups for a specific experiment inconsistent with the rest of the "Drupal experience". HOWEVER, I prefer the customized tabs a lot more than this recent change in Drupal! We may want to start a discussion in a new issue regarding this change. But for now, I feel that this PR has accomplished what it set out to do.

trpcultivate_phenotypes/src/Controller/PhenoExperimentPhenoBackupController.php
laceysanderson: Perfect! 👍

laceysanderson (approved):
✅ Good error checking for presence of entity and perfect solution for the title (i.e. using the entity label) and the experiment_id (using the new core getter method).
✅ Both buttons are now links that are styled to look like buttons removing the javascript dependency!
✅ Manual testing showed the same screenshots as everyone else 😍 but darker ;-p 

Image

Image

PR #213, opened 2026-02-13
Devcontainer not building on Mac Silicon

laceysanderson:
**Issue https://github.com/tripal/tripal/pull/2427**

## Description

Indicates the platform `linux/amd64` in both the devcontainer pull command and dockerfile FROM. This will work on all platforms (i.e. both `linux/amd64` and `linux/arm64`) since it's just specifying which version of the base image to pull and there is only one option anyway. On Mac OSX this ensures that docker doesn't complain that it can't find the nonexistent `linux/arm64` version and just fail 🙄 

## Testing?

Test that you can use devcontainer with VSCode on as many different platforms as you have available.

### DIFF
--- a/.devcontainer.json
+++ b/.devcontainer.json
@@ -8,7 +8,7 @@
       "postgresqlversion": "18"
     }
   },
-  "initializeCommand": "docker pull knowpulse/tripalcultivate-base:drupal11.x-dev-php8.5-pgsql18",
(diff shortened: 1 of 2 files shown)


vidu-wil:
I tested building the devcontainer on VSCode on Mac, and it built successfully. ✅ 🎉

Screenshot 2026-02-13 at 2 25 23 PM
Screenshot 2026-02-13 at 2 25 41 PM

vidu-wil (dismissed):

PR #222, opened 2026-04-01
G0.222 - Replace hook to attach form_validate callback with entity field constraint.

### LINKED ISSUE #220: G0.220 - Implement a field constrait to organism-related field to maintain genus-experiment relationship
### Branch

g0.220-FieldConstraints

### Groups

Group 0 - General

### Describe

Phenotypes module implements a hook_alter() to attach a form validator callback to ensure that organism with phenotypes in Research Experiment entity is protected from deletion and alterations.

A newer and preferred approach using constraints on entities and/or fields is more suitable in this application hence this issue is to update the said hook to implement field-entity-constraints.

See - https://www.drupal.org/docs/drupal-apis/entity-api/entity-validation-api/defining-constraints-validations-on-entities-andor-fields

### Design

1. Define a constraint
phenotypes\src\Plugin\Validation\Constraint

```
  /**
   * Checks that the genus with phenotypes cannot be altered or removed.
   *
   * @Constraints(
   *   id = "ExperimentGenus",
   *   label = @Translation("Unique genus", context = "Validation"),
   *   type = "string",
   * )
   */
class ExperimentGenusConstraint extend Constraint {
 
  // The message that will be shown if the genus is missing or not unique.
  public $genus_missing = 'The "%value" of this research experiment is linked to the Phenotypes module and must be a unique entry in the Germplasm Genus field
}

```

2. Create a validator callback 
phenotypes\src\Plugin\Validation\Constraint

```
class ExperimentGenusConstraintValidator extends ConstraintsValidator {
  
  // Re-use the existing field value validation logic found in the hook_alter().
}

```

3. Apply the constraints to a field using  MODULENAME_entity_type_alter() hook.
Note, this constraints must not be attached to a field via the field name as this can be altered or removed in the UI. Instead target the field that reference specific table in Chado (ie. organism). For added precision, target field with specific table reference and in a project/experiment setup.

See using Options in the said reference.

reynoldtan:
**Issue #220 Implement entity field constraint**

**Dependant on #241**

## Motivation

**NOTE:** Tests are failing on this PR. The fix is in another PR currently under review https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/244

This PR replaces the hook implementation that attaches a form validate callback to use entity field constraints instead.

## What does this PR do?

1. Implements Constraint plugin
2. Implements ConstraintValidator plugin
3. Implements hook_entity_type_alter() and attach constraint at entity type level.
4. Updates automated test.

### Automated Testing

- Test experiment genus with phenotype must exists
- Test experiment genus with phenotype is unique
- Test experiment genus with phenotype cannot be altered

### Manual Testing

**Part 1 - Test where the genus field is a textfield**

1. Switch to branch **g0.220-FieldConstraints**
2. In a fully configured clone, create organism in Tripal > Content > Add Tripal Content > **Organism**. Use **Lens** for Genus and **culinaris** for Species.
3. Configure the Genus in Tripal > Extensions > Tripal Cultivate Phenotypes Configuration.
4. Create Research Experiment Tripal Content. Tripal > Content > Add Tripal Content > Research Experiment. Click Save.
5. Edit the Research Experiment and add Lens as value to Design Tab - Germplasm Genus. You may add other Genus by clicking **Add another item**. Click Save.
6. Edit the Research Experiment and verify that you can ALTER or REMOVE the genus Lens from the list of genus. Once confirmed, add back Lens genus and click Save.
7. Add phenotypes to Lens in Tripal > Data Loaders > Tripal Cultivate Phenotypes Trait Importer (use the data file below). Run Tripal Job when instructed.
[sample traits-Lens.tsv.zip](https://github.com/user-attachments/files/26631467/sample.traits-Lens.tsv.zip)
Access the Experiment configuration page and insert at least 1 trait to Lens
Screenshot 2026-04-10 at 7 07 34 AM
8. Edit the Reseach Experiment once more and verify that with phenotypes, Lens genus cannot be ALTERED, REMOVED or DUPLICATED.

Screenshot 2026-04-10 at 7 08 54 AM

**Part 2 - Test where the genus is a dropdown**
1. Clone the Base repository and switch to branch **g1.37-organismProjectField**
2. Start a container and login.
3. In the terminal execute the commands 1 through 5 to setup Phenotypes module.

Screenshot 2026-05-26 at 2 25 40 PM

- CMD 1: cd ..
- CMD 2: git clone https://github.com/TripalCultivate/TripalCultivate-Phenotypes.git
- CMD 3: drush en trpcultivate_phenotypes trpcultivate_phenoshare trpcultivate_phenocollect -y
- CMD 4: cd TripalCultivate-Phenotypes
- CMD 5: git checkout g0.220-FieldConstraints
- OPTIONAL to clear cache: drush cr;

4. Repeat Steps 2 - 8 in Part 1. Please note that in Base, genus field is a dropdown field element and cannot alter the value therefore test that genus cannot be REMOVED or DUPLICATED.

If the warning below is shown when configuring a genus, click the provided link (Click here to manage Tripal Jobs) and rerun the job titled - **Tripal Cultivate Phenotypes: Install Ontology and Terms**. Execute the job when prompted and reload configuration page.

Screenshot 2026-06-03 at 8 11 25 AM


ruobinLiu:
Success 🎉  I confirm when there are traits associate to a Genus, it can no longer be altered, removed or duplicated. Awesome job Reynold!

Screenshot 2026-04-10 at 10 22 08 AM

ruobinLiu (dismissed):

trpcultivate_phenotypes/src/Hook/TripalCultivatePhenotypesAlterHooks.php
```diff
@@ -126,71 +95,37 @@ public function menuLocalTasksAlter(&$data, $route_name, RefinableCacheableDepen
...
+   * Implements hook_entity_type_alter().
    *
    * Enforces the genus-experiment-phenotype relationship by ensuring uniqueness
-   * and preventing modification or removal of the genus entry once phenotypic
+   * and preventing modifications or removal of the genus entry once phenotypic
    * data has been associated.
+   *
+   * NOTE: this is adding constraint to Entity Type not defined by Phenotypes at
+   * the entity level. The constraint validator rebuilds the violations to
+   * target a specific field entity to apply field-highlighting on failure.
```
laceysanderson: What do you think of this wording?


```suggestion
   * We use this hook to alter the Tripal Core tripal_entity type 
   * research_experiment bundle. We are adding an entity-level
   * constraint which will in turn target a specific field.
   * 
   * NOTE: We are not adding a field level constraint for a couple of reasons:
   * 1. We do not want to make any assumptions regarding the name or type of 
   *   the field targeting the projectprop type > genus.
   * 2. We need to ensure this constraint targets the field after the field
   *   collections have been added which prevents us from doing this in 
   *   the module install phase.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+
+  /**
+   * {@inheritdoc}
+   *
+   * NOTE: the constraint is attached at entity level and parameter #1 is
+   * a tripal_entity entity type.
+   */
+  public function validate($tripal_entity, Constraint $constraint): void {
+
+    // Determine the feild entity with the required term properties.
```
laceysanderson:
```suggestion
    // Determine the field entity with the required term properties.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+    $constraint_fieldname = '';
+
+    foreach ($tripal_entity->getFields() as $field) {
+      $field_settings = $field->getSettings();
+
+      $base_table = $field_settings['storage_plugin_settings']['base_table'] ?? NULL;
```
laceysanderson:
```suggestion
      // If the base table is not set, this field is not ready to have
      // the constraint added.
      $base_table = $field_settings['storage_plugin_settings']['base_table'] ?? NULL;
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * NOTE: the constraint is attached at entity level and parameter #1 is
+   * a tripal_entity entity type.
+   */
+  public function validate($tripal_entity, Constraint $constraint): void {
+
+    // Determine the feild entity with the required term properties.
+    $constraint_fieldname = '';
+
+    foreach ($tripal_entity->getFields() as $field) {
```
laceysanderson: It's more performant here to get the field definitions which are cached. Note: the field definition is in `$field` which can have `FieldDefinition->getSettings()` called on it just as you did before.

```suggestion
    foreach ($tripal_entity->getFieldDefinitions() as $field_name => $field) {
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+      if ($base_table != self::CONSTRAINT_FIELD_BASETABLE) {
+        continue;
+      }
+
+      $idspace = $field_settings['termIdSpace'] ?? NULL;
+      $accession = $field_settings['termAccession'] ?? NULL;
+
+      if ($idspace == self::CONSTRAINT_TERM_REQUIREMENT['idspace'] && $accession == self::CONSTRAINT_TERM_REQUIREMENT['accession']) {
+        // Field name with all constraint requirements met.
+        $constraint_fieldname = $field->getName();
```
laceysanderson: Now you already have the field name :-)

```suggestion
        $constraint_fieldname = $field_name;
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+   * validator during the value validation.
+   *
+   * @var array
+   *
+   * @see TripalCultivate Base/config/install/tripal.tripalfield_collection.trpcultivate_experiments.yml (genus)
+   */
+  private const CONSTRAINT_TERM_REQUIREMENT = [
+    'idspace' => 'TAXRANK',
+    'accession' => '0000005',
+  ];
```
laceysanderson: I don't think this is going to be specific enough. Administrators can easily configure the field term to whatever they like. I'll think about it more as I look through the rest.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,155 @@
...
+      }
+    }
+
+    // Validate the value of the field entity identified.
+    if ($constraint_fieldname !== '') {
+      $pheno_configgenus = $this->service_PhenoGenusOntology->getConfiguredGenusList();
+
+      $field_values = array_filter(
+        array_column($tripal_entity->get($constraint_fieldname)->getValue(), 'value')
+      );
```
laceysanderson: This is not going to work for Vidusha's field as it does not save the genus in the value property but rather the `genus_value` property since she is managing two properties.

Again back to my comment about choosing the field by term. I think we need to instead choose the field via its property types and find fields with the following properties:

- projectprop.type_id: look for `path` set to `project.project_id>projectprop.project_id;type_id`
- projectprop.value: look for `path` set to `project.project_id>projectprop.project_id;value`

Note: the `path` can contain aliases so you will have to substitute to real table names.
Note2: there may be more then one of each of these per field (e.g. Vidusha's field has two of each.

I think there is code in an unmerged PR of mine that will help this a lot... so hold off on implementing.

laceysanderson (changes requested):
We're going to have to put this on hold for a bit. See comments below.

trpcultivate_phenotypes/src/Hook/TripalCultivatePhenotypesAlterHooks.php
reynoldtan: perfecly detailed as always, though It is worth mentioning the field-highlighting (frontend) on failure since that is a big usability benefit/win of this implementation.
image

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: No longer applicable as the constant is no longer in use.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: This approach has been implemented in the recent commit.

trpcultivate_phenotypes/src/Hook/TripalCultivatePhenotypesAlterHooks.php
reynoldtan: Doc block updated with this expanded comment

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypes.php
```diff
@@ -0,0 +1,43 @@
...
+   * The message that will be shown if the genus has failed validation.
+   *
+   * Placeholder %genus and @reload are interpolated at runtime, with the user
+   * entered genus value and the reload link, respectively.
+   *
+   * @var string
+   *
+   * @see src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
+   */
+  public string $genus_failed = 'Update failed: Genus "%genus" of this research experiment is linked to the Phenotypes module and must be a unique entry in the Germplasm Genus field. Click @reload to restore form values if you have removed or altered a genus.';
```
laceysanderson: This is being shown to users so I would make it less module-specific and more data-specific. Also, we should take into account this may not always be specific to research experiments 😉


```suggestion
  public string $genus_failed = 'Update failed: Genus "%genus" of this %content-type is linked to phenotypic data and must remain present and unique. Click @reload to restore form values if you tried to remove or change the genus.';
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypes.php
```diff
@@ -0,0 +1,43 @@
...
+  label: new TranslatableMarkup('Lock Experiment Genus with Phenotypes'),
+  type: 'tripal_entity'
+)]
+class LockExperimentGenusWithPhenotypes extends Constraint {
+
+  /**
+   * The message that will be shown if the genus has failed validation.
+   *
+   * Placeholder %genus and @reload are interpolated at runtime, with the user
+   * entered genus value and the reload link, respectively.
```
laceysanderson: Expanded for easier reading and to include the content type.
```suggestion
   * Placeholders will be interpolated at runtime. Specifically, 
   * - %genus: genus failing validation.
   * - %content-type: the label of the content type being validated.
   * - @reload: a link to force reloading of the form.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypes.php
```diff
@@ -0,0 +1,43 @@
...
+   * The message that will be shown if all genus have been removed.
+   *
+   * Placeholder @reload are interpolated at runtime, with the user entered and
+   * the reload link.
+   *
+   * @var string
+   *
+   * @see src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
+   */
+  public string $all_genus_failed = 'Update failed: Research experiment is linked to the Phenotypes module and must maintain organism/genus values. Click @reload to restore form values if you have removed all genus.';
```
laceysanderson: Same as before.

```suggestion
  public string $all_genus_failed = 'Update failed: %content-type has phenotypic data registered and and, thus, linked genus must remain present. Click @reload to restore form values if you have removed all genus.';
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypes.php
```diff
@@ -0,0 +1,43 @@
...
+   *
+   * @see src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
+   */
+  public string $genus_failed = 'Update failed: Genus "%genus" of this research experiment is linked to the Phenotypes module and must be a unique entry in the Germplasm Genus field. Click @reload to restore form values if you have removed or altered a genus.';
+
+  /**
+   * The message that will be shown if all genus have been removed.
+   *
+   * Placeholder @reload are interpolated at runtime, with the user entered and
+   * the reload link.
```
laceysanderson:
```suggestion
   * Placeholders will be interpolated at runtime. Specifically, 
   * - %content-type: the label of the content type being validated.  
   * - @reload: a link to force reloading of the form.  
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+   *   The genus ontology service.
+   * @param \Drupal\trpcultivate_phenotypes\Service\TripalCultivatePhenotypesTermsService $service_PhenoTerms
+   *   The terms service.
+   */
+  public function __construct(
+    protected ChadoConnection $chado_connection,
+    protected TripalCultivatePhenotypesGenusOntologyService $service_PhenoGenusOntology,
+    protected TripalCultivatePhenotypesTermsService $service_PhenoTerms,
+  ) {
+
```
laceysanderson: This took me a moment to realize... since this is not common practice in our modules, I think this would be helpful to comment.


```suggestion
    // Parameters are assigned to protected properties via constructor
    // property promotion.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+    );
+  }
+
+  /**
+   * {@inheritdoc}
+   *
+   * NOTE: the constraint is attached at entity level and parameter #1 is
+   * a tripal_entity entity type.
+   */
+  public function validate($tripal_entity, Constraint $constraint): void {
```
laceysanderson: Since the inherited documentation is not that halpful, I suggest expanding it like this:


```suggestion
   * NOTE: the constraint is attached at entity level in order to validate
   * all fields that reference genus values.
   *
   * @param \Drupal\tripal\Entity\TripalEntity $tripal_entity
   *   The Tripal Entity being validated containing the field(s) being
   *   checked for genus values.
   * @param \Symfony\Component\Validator\Constraint $constraint
   *   The constraint definition which includes messages.
   */
  public function validate(TripalEntity $tripal_entity, Constraint $constraint): void {
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+    $this->pheno_configgenus = $this->service_PhenoGenusOntology
+      ->getConfiguredGenusList();
+
+    // Skip field-constraint check if system has no genus configured, or if the
+    // term genus does not have a cvterm_id.
+    if ($this->project_id === NULL || !$this->config_genus_cvterm_id || !$this->pheno_configgenus) {
+      return;
+    }
+
+    // Find fields that implement path property string.
```
laceysanderson: We should document this gotcha you came across :-p Also this feels like a more clear way to indicate exactly what we are looking for here.

```suggestion
    // Find fields that manage a chado project.projectprop record using the
    // genus cvterm as it's type_id.
    // Note: If a field has no value, then it cannot be found via this helper
    // method. See the method documentation for more details on why.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+    $genus_property_fields = $this->findGenusFieldProperty($tripal_entity);
+
+    // Trigger the constraint if an experiment has no remaining genus values
+    // while still containing phenocombo records.
```
laceysanderson: Just to be clear why this needs to be validated and provide more context then the code does alone.
```suggestion
    // Trigger the constraint if an experiment has no remaining genus values
    // but still has configured experiment-trait-method-unit combinations
    // which by definition are genus-specific.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+        ->condition('combo.project_id', $this->project_id, '=')
+        ->countQuery()
+        ->execute()
+        ->fetchField();
+
+      if ($exp_has_phenocombo) {
+        $this->context
+          ->buildViolation(
+            Markup::create(strtr($constraint->all_genus_failed, [
+              '@reload' => Link::fromTextAndUrl('Restore Values', Url::fromRoute('<current>'))->toString(),
```
laceysanderson, line 147: Include %content-type replacement here due to suggestion in the constraint definition using `$tripal_entity->getBundle()->label()`.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+            ]))
+          )
+          ->addViolation();
+      }
+
+      return;
+    }
+
+    // Find the value property of fields that implement path property string.
```
laceysanderson: Maybe make it more clear this is not necessarily named `value` and what it actually is.

> Find the TripalPropertyType key storing the genus for all fields found  using findGenusFieldProperty(). This TripalPropertyType should store the chado projectprop.value column where the projectprop.type_id of the same record references the genus cvterm.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+  /**
+   * Find Tripal field(s) that reference project.project_id;type_id property.
+   *
+   * @param \Drupal\tripal\Entity\TripalEntity $tripal_entity
+   *   Tripal Entity @see \Drupal\tripal\Entity\TripalEntity.
+   *
+   * @return array
+   *   The field names that implement the path property string.
+   */
+  protected function findGenusFieldProperty(TripalEntity $tripal_entity): array {
```
laceysanderson: I think this could be made more clear. I would name the method `findFieldsWithGenusProperty()` and then make the documentation say something like:

> This method compiles a list of all fields that store a chado projectprop record with the type_id referencing the genus cvterm. This is done by looking for a TripalPropertyValue for the projectprop.type_id specifically and checking it's value matches the cvterm_id of the genus cvterm.
>
> Note: Since we need to look at the value of projectprop.type_id property and Drupal only includes values for a field if it is not empty, we cannot detect genus property fields without a genus being assigned.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+
+    $genus_property_fields = [];
```
laceysanderson, line 195: Move the comment defining this variable from down below to where the variable is actually instantiated.

```suggestion
    // This variable is a list of fields describing a property and whose 
    // property has the same type_id as `genus (TAXRANK:0000005)`. The value 
    // is the table alias in this property for the projectprop table.
    $genus_property_fields = [];
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+   * @param array $genus_property_fields
+   *   An array of field names that have been identified to implement the path
+   *   property string.
+   * @param \Drupal\tripal\Entity\TripalEntity $tripal_entity
+   *   Tripal Entity @see \Drupal\tripal\Entity\TripalEntity.
+   *
+   * @return array
+   *   An array of field names that contains the genus value.
+   */
+  protected function findGenusFieldValue($genus_property_fields, TripalEntity $tripal_entity): array {
```
laceysanderson: This field does not actually do what it was supposed to do 🙈 The purpose of this method was to find the field and **property names** which contain the genus. Instead you are still making assumptions on property names in the validateGenusField method...

I would name this method `findFieldPropertyWithGenusValue()` and describe it's purpose as:

> This method looks up the TripalPropertyType key storing the genus value for each field looked up by findFieldsWithGenusProperty(). It uses the chado projectprop table alias returned for that field to ensure the chado projectprop.value is from the same record in the projectprop table as the type_id referencing the genus term. This is needed because a single field can manage more then one projectprop record (i.e. a scientific name property and a genus property).
>
> Note: this method does not return the actual genus value, just the name of the field and property that will contain that value.

Additionally, it should return a list of properties to validate where each element is an array containing the keys `field_name`, `property_key`, `table_alias`. This can then be used by the validateGenusField method to remove the assumption on the property key.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,297 @@
...
+   *   either the string 'value' or 'genus_value'.
+   *   ie. [genus_value => Lens] or [value => Triticum].
+   * @param \Symfony\Component\Validator\Constraint $constraint
+   *   Constraint definition.
+   */
+  protected function validateGenusField(string $field_name, array $field_values, Constraint $constraint): void {
+
+    // Find the key that corresponds to the field value and create summary
+    // Count of each unique value.
+    $value_key = array_key_exists('value', reset($field_values)) ? 'value' : 'genus_value';
```
laceysanderson: This is what I mean when I say you are still assumping the name of the property 🙈 Instead this should be the `property_key` determined by `findFieldPropertyWithGenusValue()`.

laceysanderson (changes requested):
I'm going to leave this review here with the following comments and will re-review once they're addressed. I've only gone through the constraint itself at this point.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Resolved in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/2037e8e0198499747b49b7a6f21eff73589e6123

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Applied in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/37e5b7222e2057439f7359d861af7e84f370bb08

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Method renamed in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/37e5b7222e2057439f7359d861af7e84f370bb08

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Applied in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/37e5b7222e2057439f7359d861af7e84f370bb08

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Addressed in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/37e5b7222e2057439f7359d861af7e84f370bb08

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Comment no longer applicable as recent change provided a comment with more details

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: No longer applicable in the recent changes.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: No longer applicable in the recent changes.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: No longer applicable in the recent changes.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: No longer applicable with the recent changes

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+   * NOTE: the constraint is attached at entity level in order to validate
+   * all fields that reference genus values.
+   *
+   * @param mixed $tripal_entity
+   *   The Tripal Entity being validated containing the field(s) being
+   *   checked for genus values.
+   * @param \Symfony\Component\Validator\Constraint $constraint
+   *   The constraint definition which includes messages.
+   */
+  public function validate(mixed $tripal_entity, Constraint $constraint): void {
```
laceysanderson, line 100: This method currently runs on every since Tripal entity save. As such let's check the chado base table here first and abort quickly if it's not `project`. You can use `$tripal_entity->getBundle()->getThirdPartySetting('tripal', 'chado_base_table')` to get the base table.

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+
+    // Validate genus fields.
+    foreach ($field_properties_to_validate as $constraint_field) {
+      $field_values = $tripal_entity->get($constraint_field['field_name'])->getValue();
+      $this->validateGenusField($constraint_field, $field_values, $constraint);
+    }
+  }
+
+  /**
+   * This method compiles a list of all fields that store a chado projectprop
```
laceysanderson, line 175: Needs a short one-line description for coding standards:


```suggestion
   * Find fields with projectprop.type_id property of Genus (TAXRANK:0000005).
   *
   * This method compiles a list of all fields that store a chado projectprop
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+    // property has the same type_id as `genus (TAXRANK:0000005)`. The value
+    // is the table alias in this property for the projectprop table.
+    $genus_property_fields = [];
+    $chado_fields = $tripal_entity->getTripalStorageFields('chado_storage');
+
+    // Loop through all chado fields looking for those with a path including the
+    // projectprop.type_id. This variable is a list of fields describing a
+    // property and whose property has the same type_id as
+    // `genus (TAXRANK:0000005)`. The value is the table alias in this property
+    // for the projectprop table.
```
laceysanderson: Duplicated comment which is no longer near it's variable.
```suggestion
    // Loop through all chado fields looking for those with a path including the
    // projectprop.type_id.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+
+              if (is_array($table_mapping)) {
+                $genus_property_fields[$field_name] = array_search(
+                  'projectprop',
+                  $tripal_entity->getTripalFieldPropertyInfo($field_name, $property_key, 'table_alias_mapping')
+                );
+              }
+              else {
+                $genus_property_fields[$field_name] = 'projectprop';
+              }
```
laceysanderson: What about if it has a table_mapping but does not include the projectprop table? Also, you pulled the table alias 2X. I would instead do:

```suggestion
              $genus_property_fields[$field_name] = 'projectprop';
              if (is_array($table_mapping)) {
                $alias = array_search(
                  'projectprop',
                  $table_mapping
                );
                if ($alias) {
                  $genus_property_fields[$field_name] = $alias;
                }
              }
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+
+    return $genus_property_fields;
+  }
+
+  /**
+   * Find the value property of fields that implement the path property string.
+   *
+   * @param array $genus_property_fields
+   *   An array of field names that have been identified to implement the path
+   *   property string.
```
laceysanderson:
```suggestion
   * Find the projectprop.value property with the indicated table alias.
   *
   * This method is meant to be paired with findFieldsWithGenusProperty().
   * Where that method identifies the fields with a projectprop.type_id
   * referencing the genus cvterm, this method takes the table alias for
   * that property and finds the matching projectprop.value property.
   *
   * @param array $genus_property_fields
   *   An array of field names that have been identified to have a 
   *   projectprop.type_id property referencing the genus cvterm.
   *   The key is the field name and the value is the table alias
   *   for the projectprop record storing the genus.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+   * Validate the genus value of the field identified to implement genus.
+   *
+   * The genus value is validated within the context of the experiment.
+   *
+   * @param array $field
+   *   An associative array of fields containing genus. Each field element
+   *   contains the following keys:
+   *   - 'field_name': the name of the field.
+   *   - 'property_key': the property key used to reference field value.
+   *   - 'content_type': the label of the content type the field belongs to.
```
laceysanderson: This is a single field, right? Maybe this instead?

```suggestion
   *   An associative array describing the field property to validate.
   *   Keys include:
   *   - 'field_name': the name of the field.
   *   - 'property_key': the property key used to reference field value.
   *   - 'content_type': the label of the content type the field belongs to.
```

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,334 @@
...
+   * @param array $field_values
+   *   The array of genus values of the field. The value is an array keyed by
+   *   either the string 'value' or 'genus_value'.
+   *   ie. [genus_value => Lens] or [value => Triticum].
```
laceysanderson:
```suggestion
   *   An array of field values for this field where the keys are the property 
   *   names. Note: This contains more than just the genus values so make sure 
   *   to use with the genus property key.
```

laceysanderson (commented):
Some code suggestions :-)

trpcultivate_phenotypes/trpcultivate_phenotypes.module
```diff
@@ -72,3 +70,40 @@ function trpcultivate_phenotypes_form_alter(&$form, FormStateInterface $form_sta
...
+  $schema = $connection->getSchemaName();
+
+  // Create terms defined in config entity.
+  \Drupal::service('trpcultivate_phenotypes.terms')
+    ->loadTerms($schema);
+
+  // Create genus-ontology configuration.
+  \Drupal::service('trpcultivate_phenotypes.genus_ontology')
+    ->loadGenusOntology();
+}
```
laceysanderson: This should not be here. It was removed in a previous PR in favour of `\Drupal::service('trpcultivate_phenotypes.setup_module_service')->installOntologyTerms();`

reynoldtan: Outstanding catch! I'm glad you spotted that 👍

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: This has been flagged nested flow (7 levels deep) by the bot. I've adjusted the code in the most recent commit

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Applied and expanded the test coverage in the recent commit

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,347 @@
...
+              $table_mapping = $tripal_entity->getTripalFieldPropertyInfo($field_name, $property_key, 'table_alias_mapping');
+
+              $genus_property_fields[$field_name] = 'projectprop';
+              if (is_array($table_mapping)) {
+                $alias = array_search(
+                  'projectprop',
+                  $table_mapping
+                );
+                if ($alias) {
+                  $genus_property_fields[$field_name] = $alias;
```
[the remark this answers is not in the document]
reynoldtan: Addressed in the recent commit

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
laceysanderson: Code updates look good 👍

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
```diff
@@ -0,0 +1,344 @@
...
+        ->condition('vocab.cv_id', $genus_config['trait'], '=')
+        ->countQuery()
+        ->execute()
+        ->fetchField();
+
+      // Genus has phenotypes and is missing/altered/has duplicates from the
+      // list of germplasm genus of the research experiment entity.
+      $not_unique = (isset($count_bygenus[$genus]) && $count_bygenus[$genus] > 1) ? 1 : 0;
+
+      if ($has_pheno > 0 && (!in_array($genus, $field_values) || $not_unique)) {
```
laceysanderson: Right now we only check duplicates of genus once there has been phenotypic data added... However, it would be good to check this earlier. I suggest we change this block to check the two conditions independantly:
1. Is there duplicates...violation
2. If there are phenotypes and the genus is not in the field values...violation

Note: We discussed further on this in person and we are ok with duplicates not being validated if this module does not have any genus' configured. However, if there is at least one genus configured (i.e. Tripalus) then we want to catch any duplicates even if that genus itself is not configured (i.e. give me an error if I enter Lens 2X even if only Tripalus is configured).

reynoldtan: Implemented in the recent commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/7424673bb45003dcc19fe8f497977c895be07dc6

reynoldtan: Removed duplicate genus detection from the validation in the most recent commit

laceysanderson: Perfect 👍 For posterity, this was done because of [[Bug] Genus duplicated when two related organisms are added using the organism linker field](https://github.com/TripalCultivate/TripalCultivate/issues/119) which was triggering the duplicate genus when two different species of the same genus were added.

laceysanderson (dismissed):
✅ All my code changes have been addressed ❤️ 
✅ Code looks great at this point --no other comments from me!

## Manual Testing

✅ Confirmed that duplicate genus is no longer detected and that both "Lens culinaris" and "Lens ervoides" can be added to the drop down despite having the same genus. Image
✅ Confirmed that if **no traits** have been configured for the experiment then any genus can be removed/changed.
✅ Confirmed that even if one trait is configured, then the genus that trait is for cannot be removed (in my test Tripalus). Image
✅ Confirmed that even after traits are added for a configured genus, non-configured genus' can still be removed (in my test Lens).
✅ Confirmed this is no longer run on non-project-based content types (in my test organism)
✅ Confirmed that this is run on non-experiments (in my case Genome Project) but that nothing is triggered because we don't have traits defined. There are also no errors in the log. Image

ruobinLiu (dismissed):

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Resolving this on your behalf having confirmed full implementation in this line  - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/379827031c3318d38a14d6d1b5b7b22e80d9709a/trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php#L156

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Resolving this on your behalf with reference to its full implementation in this line https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/379827031c3318d38a14d6d1b5b7b22e80d9709a/trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php#L190

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: This is no longer applicable with the recent revision

trpcultivate_phenotypes/src/Plugin/Validation/Constraint/LockExperimentGenusWithPhenotypesValidator.php
reynoldtan: Doc block revised in this commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/222/commits/ff6eef874ad8d8a328c7819f4b8d35c015ab34ab

ruobinLiu (approved):

laceysanderson (approved):

PR #223, opened 2026-04-10
G4.221 - PhenoCombo Service

### LINKED ISSUE #221: G4.221 - Combo CRUD operations
### Branch

g4.221-ComboCRUD

### Groups

### Describe

Phenotypes module trait-method-unit combo requires operation to:

1. Create - insert trait method unit combo for an experiment.
2. Read/List - pull all combos of an experiment (currently being reviewed in trait service as getExperimentTraitMethodUnitCombos()).
3. Update - alter trait combo status fields (ie. is_required, was_collected)
4. Delete/Remove - disassociate from an experiment (not delete the trait record from cvterm table). 

### Design

Create additional method in the Trait Service Class:

**Create**
```
 /**
   * Assign a trait-method-unit combo to an experiment.
   *
   * @param array $trait_combo
   *   An associative array describing the trait-method-unit combination:
   *   - attr_id: Trait cvterm ID (type: trait).
   *   - observable_id: Method cvterm ID (type: method).
   *   - unit_id: Unit cvterm ID (type: unit).
   *   - label: a human-readable text label used as an alternative reference to
   *     the trait name. Labels must be unique within an experiment.
   *   - is_archived: 1 if archived, 0 otherwise.
   *   - is_required: 1 if required, 0 otherwise.
   *   - was_shared: 1 if used in Phenotypes Share module, 0 otherwise.
   *   - was_collected: 1 if measured in Phenotypes Collect module, 0 otherwise.
   * @param int|string $experiment
   *   The experiment identifier, either:
   *   - An integer project_id
   *   - A string experiment name (corresponding to 'name' column in Chado
   *    'project' table).
   *
   * @return object
   *   The trait-method-unit combo inserted as a database row object.
   */
  public function insertTraitMethodUnitComboToExperiment(array $trait_combo, int|string $experiment): \stdClass {
  
    // Ensure experiment (if int) is a number and > 0.

    // Determine what is provided. If string, use ChadoAutocompleteController to resolve name to the project id.

    // Make sure attr_id, observable_id, and unit id exist using trait service to pull trait-method-unit combo.

    // Within the experiment, inspect that the label is unique.

    // Insert the combo into the combo base table.
  }  
```

**Update**

```

  /**
   * Update trait-method-unit combo status flags.
   * 
   * @param int $combo_id
   *   The unique identifier of a trait-method-unit combo.
   * @param array $status_flags
   *   The status array key-value pair where the value is 0 or 1 (yes or no, respectively) and
   *   the keys are the following status flags. Default to empty array (nothing to update).
   *   - is_archived: indicates trait combo is archived.
   *   - is_required: indicates trait combo is a required trait and must contain a value in the data file for this column.
   *   - was_shared:  indicates trait combo was used in Phenotypes Share module.
   *   - was_collected: indicates trait was measured in Phenotypes Collect module.
   */
  public function updateTraitMethodUnitComboStatusById(int $combo_id, array $status_flags = []): void {
   
    // Ensure combo id is valid.

    // Update the combo status flags in combo_id with flags in the status flags array.
  }

``` 

**Remove**

```
  /**
   * Remove a trait-method-unit combo within the experiment.
   * 
   * @param int $combo_id
   *   The unique identifier of a trait-method-unit combo.
   */
  public function removeTraitMethodUnitComboById(int $combo_id): void {
    
    // Ensure combo id is valid.

    // Ensure combo has no associated phenotypes.

    // Only then can the combo be removed from the experiment.
  }
```

reynoldtan:
**Issue https://github.com/TripalCultivate/TripalCultivate-Phenotypes/issues/221** PhenoCombo CRUD service

**NOTE: This PR is dependent on PR https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/246 Expand SetupModuleService to alter the chado phenotype table**

## Motivation
This PR will setup a service that will handle Creation, Retrieval, Modification, and Removal of PhenoCombo.

## What does this PR do?

1. Setup each PhenoCombo CRUD operation.

| Operation | Method(s) |
| -------- | -------- |
| CREATE | assignPhenoComboToExperiment() |
| READ | getExperimentPhenoCombo() and getAllExperimentPhenoCombos() |
| UPDATE | setExperimentPhenoComboStatusFlags() |
| REMOVE | removePhenoComboFromExperiment() |

2. Create helper methods
setExperiment(), ensureExperimentIsSet(), sanitizePhenoCombo(), sanitizePhenoComboDetails(), resolvePhenoCombo(), resolveExperimentToProjectId(), labelIsUniqueInExperiment() and experimentHasCombo()
 
3. Setup automated test.

### Automated Testing

- [x] Test experiment context setter method with various experiment identifiers.
- [x] Test assign PhenoCombo to experiment. 
- [x] Test get specific PhenoCombo.
- [x] Test get all PhenoCombos.
- [x] Test set experiment PhenoCombo status
- [x] Test PhenoCombo removal
- [x] Test helper methods to sanitize PhenoCombo, sanitize PhenoCombo details, and check if experiment has combo. 
- [x] Test invalid arguments and no configured genus  

### Manual Testing

### **PART 1 - SETUP:**
----------------
1. Switch to **g4.221-ComboCRUD** branch and build a clone.
2. Setup 2 organism in Content > Tripal Content > Add Tripal Content > Organism
(**my_site/bio_data/add/organism**)

- **Lens culinaris**
- **Triticum durum**

3. Configure the genus (set Trait, Method, Unit and DB) so that it is recognized by the Phenotypes Module.
Tripal > Extensions > TripalCultivate Phenotypes Configuration > Ontology Terms
(**my_site/admin/tripal/extension/tripal-cultivate/phenotypes/ontology**)

4. Create a Research Experiment Tripal Content Type.
Content > Tripal Content > Add Tripal Content > Research Experiment
(**my_site/bio_data/add/research_experiment**)

**Experiment Name: Awesome Experiment**

Provide the experiment name and add the 2 organisms created in step 1 (provide only the genus into the field) into the Germplasm Genus field found in Design tab (use the button Add another item to add a second field). Click Save.

image

5. Upload traits. Run Tripal job for each genus.
Tripal > Data Loaders > Tripal Cultivate: Phenotypic Trait Importer
(**my_site/admin/tripal/loaders/trpcultivate-phenotypes-traits-importer**)

https://github.com/user-attachments/files/24919913/Traits.zip

### **PART 2 - MANUAL TEST:**
----------------------------

Required input to method call.

| Name | Description |
| -------- | -------- |
| $pheno_combo | an array with keys - trait, method, and unit. Value for each key can either be the name or the cvterm id. |
| $combo | can be an integer (combo_id), a string (label) or trait combo (described above) |
| $experiment_pheno_combo_details | an array with keys - label, is_archived, is_required, was_collected, was_shared. Value for each key, with the exception of label, is 0 or 1 (no/yes) |

Examples:

```
$pheno_combo = [
  'trait' => 'My Trait Name', 
  'method' => 3499,
  'unit' => 'kg',
];

// Combo as:
$combo_id = 1;
$label = 'A Label Text';
$pheno_combo = [
  'trait' => 3312, 
  'method' => 'MY METHOD',
  'unit' => 2245,
];

$experiment_pheno_combo_details = [
  'label' => 'A Label',
  'is_archived' => 1,
  'is_required' => 0,
  'was_collected' => 0,
  'was_shared' => 0,
];
```

Hint:
_Launch the trait selector window in Phenotypes section of the research experiment. Reference names shown or inspect label text field element for ids, both forms can be used to create trait-method-unit ($pheno_combo) array._

Screenshot 2026-05-11 at 8 52 43 AM

Edit research experiment form and access Phenotypes from the drop menu.
image

### **Method Calls**
Navigate to my **site/devel/php** and copy the code below, paste in the the space provided and click Execute button.

```
// This is the experiment context, all operations performed will be restricted to this experiment.
$service = \Drupal::service('trpcultivate_phenotypes.pheno_combo');
$experiment = 'Awesome Experiment';
$service->setExperiment($experiment);

// Paste subsequent test method call code below this line.
```

**1. Test assign pheno-combo**
Assign as many traits desired.

```
$pheno_combo = [
  'trait' => 'Cotyledon Colour',
  'method' => 'COTYCOLR',
  'unit' => 'colour'
];

// Do not reuse this label value for other pheno-combos. Label must be unique within the experiment.
// Change status flag values to either 0 or 1 or remove status flag entirely (will be set to default 0).
$experiment_pheno_combo_details = [
  'label' => 'Label For Cotyledon Colour',
  'is_archived' => 0,
  'is_required' => 1,
  'was_shared' => 0,
  'was_collected' => 1,
];

$combo_id = $service->assignPhenoComboToExperiment($pheno_combo, $experiment_pheno_combo_details);
dpm($combo_id);
```

To confirm success of each operation, query trpcultivate_phenocombo table, use pheno-combo getter methods, or view experiment pheno-combo summary listing page.
Screenshot 2026-05-11 at 11 05 21 AM

Edit research experiment form and access Phenotypes from the drop menu.
image

**2. Test retrieve PhenoCombo.**
NOTE: Remove previous test code before copy-pasting code block below.

```
$all_phenocombos = $service->getAllExperimentPhenoCombos();
dpm($all_phenocombos);
```

```
// Provide one of the following example $combo argument:
$combo_id = 1;
$label = 'Label For Cotyledon Colour';
$pheno_combo = [
  'trait' => 'Cotyledon Colour',
  'method' => 'COTYCOLR',
  'unit' => 'colour'
];

$a_phenocombo = $service->getExperimentPhenoCombo($combo_id);
dpm($a_phenocombo, 'by combo_id');
$a_phenocombo = $service->getExperimentPhenoCombo($label);
dpm($a_phenocombo, 'by label');
$a_phenocombo = $service->getExperimentPhenoCombo($pheno_combo);
dpm($a_phenocombo, 'by pheno_combo');
```

**3. Test set PhenoCombo status**
NOTE: Remove previous test code before copy-pasting code block below.

```
// Provide one of the following example $combo argument:
$combo_id = 1;
$label = 'Label For Cotyledon Colour';
$pheno_combo = [
  'trait' => 'Cotyledon Colour',
  'method' => 'COTYCOLR',
  'unit' => 'colour'
];

// Change values to either 0 or 1.
$experiment_pheno_combo_details = [
  'is_archived' => 1,
  'is_required' => 0,
  'was_shared' => 1,
  'was_collected' => 0,
];

// Pick one.
$service->setExperimentPhenoComboStatusFlags($combo_id, $experiment_pheno_combo_details);
$service->setExperimentPhenoComboStatusFlags($label, $experiment_pheno_combo_details);
$service->setExperimentPhenoComboStatusFlags($pheno_combo, $experiment_pheno_combo_details);
```

**4. Test PhenoCombo removal** 
NOTE: Remove previous test code before copy-pasting code block below.

```
// Provide one of the following example $combo argument:
$combo_id = 1;
$label = 'Label For Cotyledon Colour';
$pheno_combo = [
  'trait' => 'Cotyledon Colour',
  'method' => 'COTYCOLR',
  'unit' => 'colour'
];

// Pick one.
$service->removePhenoComboFromExperiment($combo_id);
$service->removePhenoComboFromExperiment($label); 
$service->removePhenoComboFromExperiment($pheno_combo);
```


trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php
```diff
@@ -0,0 +1,817 @@
...
+    'Triticum' => [
+      [
+        'Biomass',
+        'The total mass of plant-based or organic matter.',
+        'B-MASS',
+        'Use quadrant (a marked frame) to define a specific area',
+        'kilogram',
+        'Quantitative',
+      ],
+    ],
```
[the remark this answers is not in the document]
reynoldtan: This is a non-issue as the similar line is the value Quantitative.

carolyncaron (commented):
✅ I completed the manual test and was able to add 2 trait combos (1 for each genus): 
Image

✅ I confirmed that I could change details for the combos using `setExperimentPhenoComboStatusFlags()` (see that details are different for the trait combos above, both were initially inserted with the same details).

✅ I also confirmed that I could delete combos using any combination of combo_id, label, and pheno_combo array.

My code review is still to follow!

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+use Drupal\tripal_chado\Controller\ChadoProjectAutocompleteController;
+
+/**
+ * Experiment - Phenotypic Combo Service.
+ *
+ * This service handles all processes linking an experiment with an existing
+ * Phenotypic Combo (i.e. Trait - Method - Unit Combo) and setting status flags
+ * relating to that link.
+ *
+ * Definitions:
```
carolyncaron, line 22: These definitions are really helpful! I think you could also add a definition here for what you mean by "status flags".

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *
+   * @param \Drupal\tripal\Entity\TripalEntity|int|string $experiment
+   *   The experiment identifier, either:
+   *   - Experiment entity (Tripal Entity) the experiment entity object.
+   *   - An integer (int) value corresponding to the project_id number.
+   *   - A string (string) value corresponding to the project name.
+   *   Both int and string forms reference a field from the same Chado
+   *   'projects' table.
+   *
+   * @throws Exception
```
carolyncaron:
```suggestion
   * @throws \Exception
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+
+    $this->experiment_context = $project_id;
+  }
+
+  /**
+   * Assign a PhenoCombo to an experiment.
+   *
+   * @param array $pheno_combo
+   *   An associative array describing the TRAIT-METHOD-UNIT combination.
+   *   This MUST ALREADY EXISTS in Chado. The following keys are expected:
```
carolyncaron:
```suggestion
   *   This MUST ALREADY EXIST in Chado. The following keys are expected:
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    $this->ensureExperimentIsSet();
+
+    if (!isset($experiment_pheno_combo_details['label'])) {
+      throw new \InvalidArgumentException(
+        'Missing label key error. The key \'label\' must exist in the \'$experiment_pheno_combo_details\' parameter.'
+      );
+    }
+
+    $sanitized_pheno_combo = $this->sanitizePhenoCombo($pheno_combo);
+    $sanitized_experiment_pheno_combo_details = $this->sanitizePhenoComboDetails($experiment_pheno_combo_details);
```
carolyncaron: These variables are getting harder to read the longer they get. What about reassigning the original variables with the sanitized versions instead of creating new variable names? It doesn't look like the unsanitized versions are used again after this point.
```suggestion
    // Sanitize the input.
    $pheno_combo = $this->sanitizePhenoCombo($pheno_combo);
    $experiment_pheno_combo_details = $this->sanitizePhenoComboDetails($experiment_pheno_combo_details);
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *
+   * @param string|null $genus
+   *   The genus to further filter the experiment PhenoCombo results based on
+   *   genus-configuration cvterm.cv_id values for trait, method, and unit.
+   * @param array $options
+   *   Options to customize the query result by restricting the fields returned.
+   *   The following options are supported:
+   *     - format: one of 'full' (default), 'component', or 'header' depending
+   *       on the format of the return value desired. See the return value
+   *       below for more details.
```
carolyncaron: Just want it to be clear that 'format' is a key in the $options array:
```suggestion
   *   The following options are supported:
   *   - 'format': one of 'full' (default), 'component', or 'header' depending
   *     on the format of the return value desired. See the return value
   *     below for more details.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *
+   *   $options = ['format' => 'component' | 'header' | 'full']
+   *
+   *   - component: the component format is suitable of use with trait_combo
+   *   component as value to the key #props.
+   *   @see component/trait_combo
+   *
+   *   - header: the format used for data loader headers. This format includes
+   *   combo_id, name (trait cvterm.name), description (trait cvterm.definition)
+   *   and type(Required or Optional).
```
carolyncaron:
```suggestion
   *   and type (Required or Optional).
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   @see TripalCultivatePhenoTraitImporter::$headers
+   *
+   *   - full:
+   *   @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::getExperimentPhenoCombo()
+   *     The return value of this method.
+   *
+   * @throws \InvalidArgumentException
+   *   - If a genus is not a configured genus of the experiment.
+   *   - If not a valid options key provided in $options parameter.
+   *   - If not a valid fetch format value provided in the $options parameter.
```
carolyncaron:
```suggestion
   *   - If an invalid valid options key is provided to the $options parameter.
   *   - If an invalid value is provided for 'format' in the $options parameter.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+      throw new \InvalidArgumentException(
+        sprintf(
+          'Failed to get experiment PhenoCombos. Unsupported options key provided [%s]. Only keys [%s] are allowed.',
+          implode(', ', $options_diff),
+          implode(', ', $valid_options_keys)
+        )
+      );
+    }
+
+    // If format is not specified, default to full, otherwise ensure that it can
```
carolyncaron, line 407: If 'full' is the default, I would add this to the method's docblock

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+      ->condition('tbl.attr_id', $pheno_combo->attr_id, '=')
+      ->condition('tbl.observable_id', $pheno_combo->observable_id, '=')
+      ->condition('tbl.assay_id', $pheno_combo->unit_id, '=')
+      ->countQuery()
+      ->execute()
+      ->fetchField();
+
+    if ($combo_phenotypes_count) {
+      throw new \InvalidArgumentException(
+        'PhenoCombo has Phenotypes error. The PhenoCombo has associated phenotypic records.'
```
carolyncaron:
```suggestion
        'Failed to remove PhenoCombo from this experiment. The PhenoCombo has associated phenotypic records.'
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   cvterm_id number (NOT THE CVTERM RECORD).
+   *   ie. [trait => TRAIT ID, method => METHOD ID, unit => UNIT ID].
+   *
+   * @throws InvalidArgumentException
+   *   - If host site with Phenotypes module has no genus configured.
+   *   - If PhenoCombo is missing any of the key(s) trait, method, and unit.
+   */
+  public function sanitizePhenoCombo(array $pheno_combo): array {
+
+    // Phenotypes module hosted has no genus configured.
```
carolyncaron:
```suggestion
    // Check if the Phenotypes module has genus configured.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    if ($config_genus === []) {
+      throw new \Exception(
+        sprintf(
+          'The Phenotypes module is not configured with a genus. Please navigate to %s to configure a genus.',
+          Url::fromRoute('trpcultivate_phenotypes.settings_ontology')->toString()
+        )
+      );
+    }
+
+    // PhenoCombo is missing an alias.
```
carolyncaron:
```suggestion
    // Check if PhenoCombo is missing an alias.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    }
+
+    // Verify trait, method and unit exist.
+    $pheno_combo = NULL;
+
+    foreach ($pheno_combo_alias as $alias) {
+      ${$alias} = $sanitized_pheno_combo[$alias];
+    }
+
+    // Find the combination in each config-genus.
```
carolyncaron:
```suggestion
    // Look for this combination in each configured genus.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+        }
+      }
+      catch (\Exception $e) {
+        continue;
+      }
+    }
+
+    if (is_null($pheno_combo)) {
+      throw new \InvalidArgumentException(
+        'Missing PhenoCombo error. The trait-method-unit combinations does not exist.'
```
carolyncaron:
```suggestion
        'Missing PhenoCombo error. The trait-method-unit combination does not exist for any configured genus.'
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    foreach ($pheno_combo_alias as $alias) {
+      ${$alias} = $sanitized_pheno_combo[$alias];
+    }
+
+    // Find the combination in each config-genus.
+    foreach ($config_genus as $genus) {
+      $this->service_PhenoTraits->setTraitGenus($genus);
+
+      try {
+        if (($pheno_combo = $this->service_PhenoTraits->getTraitMethodUnitCombo($trait, $method, $unit)) !== NULL) {
```
carolyncaron, line 712: Does this break the foreach at the first successful combo it finds? Is that the intended result? What if multiple pheno combos exist?

If I'm misinterpretting this code, then a detailed comment would go a long way here!

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+          break;
+        }
```
carolyncaron, line 714: Does this break the foreach at the first successful combo it finds? Is that the intended result? What if multiple pheno combos exist?

If I'm misinterpretting this code, then a detailed comment would be very helpful here!

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   - If label is an empty string and is not unique label in the experiment.
+   *   - If any of the status flag has unexpected or non-integer value.
+   */
+  public function sanitizePhenoComboDetails(array $experiment_pheno_combo_details): array {
+
+    if ($experiment_pheno_combo_details === []) {
+      return [];
+    }
+
+    $sanitized_experiment_pheno_combo_details = [];
```
carolyncaron: Again, if you find a way to shorten this, I'd be grateful 🙈

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *     the experiment indicated.
+   *   - Status Flags
+   *     @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::setExperimentPhenoComboStatus()
+   *
+   * @return array
+   *   A validated and sanitized experiment PhenoCombo details where status flag
+   *   values are guaranteed 0 or 1 value and label trimmed and is a unique
+   *   label within experiment context.
+   *
+   * @throws InvalidArgumentException
```
carolyncaron:
```suggestion
   * @throws \InvalidArgumentException
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+        }
+
+        $sanitized_experiment_pheno_combo_details[$field] = $status_flag_val;
+      }
+    }
+
+    if ($unexpected_values != []) {
+      throw new \InvalidArgumentException(
+        sprintf(
+          'Unexpected PhenoCombo status flag value type error. Status flag(s) contains invalid value in status flags(s) [%s].',
```
carolyncaron:
```suggestion
          'Unexpected PhenoCombo status flag value type error. Status flag(s) contains invalid value(s): [%s].',
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   * @param array $status_flags
+   *   The status array key-value pair where the value is 0 or 1
+   *   (yes or no, respectively) and the keys are the following status flags.
+   *   - is_archived: indicates experiment PhenoCombo is archived.
+   *   - is_required: indicates experiment PhenoCombo is a required trait and
+   *     must contain a value in the data file for this column.
+   *   - was_shared: indicates experiment PhenoCombo was used in
+   *     Phenotypes Share module.
+   *   - was_collected: indicates experiment PhenoCombo was used in
+   *     Phenotypes Collect module.
```
carolyncaron, line 518: I find adding single quotes around potential keys helps my brain to see them quickly - especially when developing and I want to look at the method's definition
```suggestion
   *   The status array key-value pair where the value is 0 or 1
   *   (yes or no, respectively) and the keys are the following status flags:
   *   - 'is_archived': indicates experiment PhenoCombo is archived.
   *   - 'is_required': indicates experiment PhenoCombo is a required trait and
   *     must contain a value in the data file for this column.
   *   - 'was_shared': indicates experiment PhenoCombo was used in
   *     Phenotypes Share module.
   *   - 'was_collected': indicates experiment PhenoCombo was used in
   *     Phenotypes Collect module.
```

carolyncaron (changes requested):
Here is my code review of the service methods (haven't had a chance to look at the tests yet). These are mostly minor wording suggestions! 😁

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   * @throws \InvalidArgumentException
+   *   - If a genus is not a configured genus of the experiment.
+   *   - If not a valid options key provided in $options parameter.
+   *   - If not a valid fetch format value provided in the $options parameter.
+   */
+  public function getAllExperimentPhenoCombos(string|null $genus = NULL, array $options = []):array|null {
+
+    $this->ensureExperimentIsSet();
+
+    if ($genus !== NULL && $genus !== '') {
```
vidu-wil:
```suggestion
    if (!empty($genus)) {
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   - If failed to update a PhenoCombo status flag (database error).
+   */
+  public function setExperimentPhenoComboStatusFlags(array|int|string $combo, array $status_flags): void {
+
+    $this->ensureExperimentIsSet();
+
+    $combo_id = $this->resolvePhenoCombo($combo);
+    $sanitized_status_flags = $this->sanitizePhenoComboDetails($status_flags);
+
+    // Nothing to set if not one status was provided.
```
vidu-wil:
```suggestion
    // Nothing to set if status was not provided.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+  /**
+   * Remove a PhenoCombo from an experiment.
+   *
+   * NOTE: this does not delete the trait, method, and unit cvterm records.
+   *
+   * @param array|int|string $combo
+   *   The experiment PhenoCombo previously assigned to the experiment.
+   *   @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::getExperimentPhenoCombo()
+   *
+   * @throws InvalidArgumentException
```
vidu-wil:
```suggestion
   * @throws \InvalidArgumentException
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+
+      $this->tripal_logger->error($e->getMessage());
+      throw new Exception($e->getMessage());
+    }
+  }
+
+  /**
+   * Ensures that experiment context has been set.
+   *
+   * @throws Exception
```
vidu-wil:
```suggestion
   * @throws \Exception
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   An associative array describing the TRAIT-METHOD-UNIT combination.
+   *   @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::assignPhenoComboToExperiment()
+   *
+   * @return array
+   *   A validated, sanitized, and normalized PhenoCombo where any or all
+   *   names are converted to their cvterm ids, and the final PhenoCombo uses
+   *   cvterm_id number (NOT THE CVTERM RECORD).
+   *   ie. [trait => TRAIT ID, method => METHOD ID, unit => UNIT ID].
+   *
+   * @throws InvalidArgumentException
```
vidu-wil:
```suggestion
   * @throws \InvalidArgumentException
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   following optional keys may be provided and will be validated:
+   *   - label (string) uniquely identifying this trait-method-unit combo in
+   *     the experiment indicated.
+   *   - Status Flags
+   *     @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::setExperimentPhenoComboStatus()
+   *
+   * @return array
+   *   A validated and sanitized experiment PhenoCombo details where status flag
+   *   values are guaranteed 0 or 1 value and label trimmed and is a unique
+   *   label within experiment context.
```
vidu-wil:
```suggestion
   *   A validated and sanitized experiment PhenoCombo details where status flag
   *   values are guaranteed to be 0 or 1 in value and label is trimmed and is a 
   *   unique label within experiment context.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   * Resolve combo parameter values to experiment PhenoCombo combo_id.
+   *
+   * @param array|int|string $combo
+   *   The experiment PhenoCombo previously assigned to the experiment.
+   *   @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::getExperimentPhenoCombo()
+   *
+   * @return int
+   *   The combo unique identifier id (combo_id).
+   *
+   * @throws InvalidArgumentException
```
vidu-wil:
```suggestion
   * @throws \InvalidArgumentException
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    // Phenotypes module hosted has no genus configured.
+    $config_genus = $this->service_PhenoGenusOntology->getConfiguredGenusList();
+    if ($config_genus === []) {
+      throw new \Exception(
+        sprintf(
+          'The Phenotypes module is not configured with a genus. Please navigate to %s to configure a genus.',
+          Url::fromRoute('trpcultivate_phenotypes.settings_ontology')->toString()
+        )
+      );
+    }
```
vidu-wil: Missing test coverage for this section.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+          'Invalid label error. The label must be a string value and not empty.'
+        );
+      }
+
+      $query
+        ->condition('tbl.project_id', $this->experiment_context, '=')
+        ->condition('tbl.label', $label, '=');
+    }
+    else {
+      throw new \InvalidArgumentException('Unsupported combo input type');
```
vidu-wil: Missing test coverage for the else condition.

trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php
```diff
@@ -0,0 +1,1075 @@
...
+      ->execute()
+      ->fetchAssoc();
+
+    // Check that all status flags are set to no (0) at the beginning.
+    $this->assertEquals(
+      array_fill_keys($status_flags, 0),
+      array_filter($a_pheno_combo, function ($i) use ($status_flags) {
+        return in_array($i, $status_flags);
+      }, ARRAY_FILTER_USE_KEY),
+      'The status flags are expected to all be set to 0 (No).',
```
vidu-wil:
```suggestion
      'All the status flags are expected to be set to 0 (No).',
```

trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php
```diff
@@ -0,0 +1,1075 @@
...
+      ->execute();
+
+    try {
+      $this->service_PhenoCombo->removePhenoComboFromExperiment($combo->combo_id);
+    }
+    catch (\Exception $e) {
+      $this->assertStringContainsString(
+        'PhenoCombo has Phenotypes error',
+        $e->getMessage(), 'PhenoCombos
+        with associated phenotypic are protected from remove operation.'
```
vidu-wil:
```suggestion
        $e->getMessage(), 'PhenoCombos with associated phenotypic records are protected from the remove operation.'
```

trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php
```diff
@@ -0,0 +1,1075 @@
...
+    }
+    catch (\Exception $e) {
+      $this->assertStringContainsString(
+        'Missing PhenoCombo error',
+        $e->getMessage(),
+        'PhenoCombo not found is expected to throw an exception.'
+      );
+    }
+
+    // Using triat ids inserted as PhenoCombo, sanitize pheno_combo - the
```
vidu-wil:
```suggestion
    // Using trait ids inserted as PhenoCombo, sanitize pheno_combo - the
```

vidu-wil (changes requested):
I did the manual testing for this PR and everything works as expected. 

✅I assigned two traits, one for each genus and they were added correctly.

Image

Image

Image

✅Retrieves the PhenoCombo correctly by using combo_id, label and pheno combo.

Image

✅Sets the status correctly.

Image

✅Removes the pheno combo correctly. (The following screenshot represents the message shown when I removed the pheno combo and tried to access it after removing)

Image

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Good suggestion. It is intentionally limited to the definitions of terms that appear in the class name - Experiment and PhenoCombo. Status flags are part of the class behaviour rather than terminology used in the class name so I kept the definitions (prepared by @laceysanderson) focused on the name context. Status flags are defined in details where it is pertinent.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: I used $sanitized_* variables intentionally and followed the same pattern throughout the class. The goal is to clearly distinguish between incoming raw data and one that passed through validation/sanitation, safe and unsafe to use, raw and sanitized input, etc.  I would reuse the same variable if it undergoes simple transformation like trim or lowercase, but in this case values undergo significant operation and sometimes throws an exception, I found the distinction helpful for readability and maintenance (because method retain access to raw parameter values, it will be easy to implement future enhancements that require additional data).

Totally agree with the long variable name, shortened to

```
$sanitized_phenocombo_details
```

in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/fbe60f34302bc9b88240f39667a45aee9ca80604

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Good catch

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Good catch! I expanded doc block to include default format option in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/df545681a9520131368481ebf682786cb46f2173

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Revised in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/858b1ac6d492b0430f43398654afd6809ae3ebd1

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: No, you're not misinterpreting this code. You're exactly right it will stop at the first combo it finds within a genus context and since the DuplicateTraits validator guarantees combos are unique per genus, that is all it needs to verify existence of a combo (previously added through traits importer) and can safely be used

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: This else block is unreachable since the type hint of the parameter will capture incorrect type. I addressed it by removing this block from this method in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/c641c7fa945361634d498d8a1d127c0470a90a5a

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: This is the test for this exception - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/f9b9a9de9deb48f22e7d7672d3cd3c3be152da59/trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php#L1056

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Shortened in commit https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/fbe60f34302bc9b88240f39667a45aee9ca80604

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   Both int and string forms reference a field from the same Chado
+   *   'projects' table.
+   *
+   * @throws \Exception
+   *   - If Phenotypes module hosted has not been configured with a genus.
+   *   - If experiment is not configured with a genus in Phenotypes module.
+   */
+  public function setExperiment(TripalEntity|int|string $experiment): void {
+
+    if (empty($this->service_PhenoGenusOntology->getConfiguredGenusList())) {
```
laceysanderson:
```suggestion
    // First lets confirm that we have genus configured in the first place.
    if (empty($this->service_PhenoGenusOntology->getConfiguredGenusList())) {
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+      throw new \Exception(
+        sprintf(
+          'The Phenotypes module is not configured with a genus. Please navigate to %s to configure a genus.',
+          Url::fromRoute('trpcultivate_phenotypes.settings_ontology')->toString()
+        )
+      );
+    }
+
+    $project_id = self::resolveExperimentToProjectId($experiment);
```
laceysanderson, line 169:
```suggestion
    // Now let's resolve the project_id regardless of the context we were given. 
    // If this method is not able to resolve the project_id then it will throw an 
    // exception so there's no need to check it again here.
    $project_id = self::resolveExperimentToProjectId($experiment);
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+
+    if (empty($this->service_PhenoGenusProject->getGenusOfProject($project_id))) {
```
laceysanderson, line 173:
```suggestion
    // Now that we have a project, let's ensure it has genus' assigned
    // as we expect for use with phenotypes.
    if (empty($this->service_PhenoGenusProject->getGenusOfProject($project_id))) {
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+
+  /**
+   * Resolve experiment context to project_id.
+   *
+   * @param \Drupal\tripal\Entity\TripalEntity|int|string $experiment
+   *   The experiment identifier.
+   *   @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::setExperiment()
+   *
+   * @return int
+   *   The project id (Chado.project: project_id) of the experiment context.
```
laceysanderson:
```suggestion
   *   The project id (Chado project.project_id) of the experiment context.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   * @throws InvalidArgumentException
+   *   - If experiment context could not resolve to an existing project_id.
+   */
+  protected static function resolveExperimentToProjectId(TripalEntity|int|string $experiment): int {
+
+    $project_id = 0;
+
+    if ($experiment instanceof TripalEntity) {
+      $project_id = $experiment->getBackendRecordId('chado_storage');
+    }
```
laceysanderson: What if the TripalEntity given is not project-based? For example, what if someone gives you an analysis-based content type here... this method would return the analysis_id and pretend it is a project_id... it would be good to confirm it is a project-based content type.

This can be done using `$experiment->getBundle()->getThirdPartySetting('tripal', 'chado_base_table');`

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    elseif (is_numeric($experiment)) {
+      $project_id = ChadoProjectAutocompleteController::getProjectName((int) $experiment)
+        ? $experiment : $project_id;
+    }
+    elseif (is_string($experiment)) {
+      $project_id = ChadoProjectAutocompleteController::getProjectId($experiment);
+    }
+
+    if (is_null($project_id) || $project_id === 0 || $project_id === '') {
```
laceysanderson: It would be better to check project_id is not what you want it to be here. Specifically, NOT(is_int(project_id) AND project_id > 0). Your current approach would not throw an error if an object or array was returned 😬

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   ie. [trait => TRAIT ID, method => METHOD ID, unit => UNIT ID].
+   *
+   * @throws \InvalidArgumentException
+   *   - If host site with Phenotypes module has no genus configured.
+   *   - If PhenoCombo is missing any of the key(s) trait, method, and unit.
+   */
+  public function sanitizePhenoCombo(array $pheno_combo): array {
+
+    // Check if the Phenotypes module has genus configured.
+    $config_genus = $this->service_PhenoGenusOntology->getConfiguredGenusList();
```
laceysanderson: This is also called in setExperiment(). I would suggest storing it in a class property there so you don't need to grab it again here.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
laceysanderson: The coverage report still shows no coverage for this check: https://qlty.sh/gh/TripalCultivate/projects/TripalCultivate-Phenotypes/pull/223/coverage. The test you quoted checks setExperiment() only. You also need to call this method directly in that test to confirm it is caught here too.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+        sprintf(
+          'The Phenotypes module is not configured with a genus. Please navigate to %s to configure a genus.',
+          Url::fromRoute('trpcultivate_phenotypes.settings_ontology')->toString()
+        )
+      );
+    }
+
+    // Check if PhenoCombo is missing an alias.
+    $pheno_combo_alias = array_keys(self::PHENOCOMBO_FIELD_MAP);
+    if (array_diff($pheno_combo_alias, $input_keys = array_keys($pheno_combo))) {
```
laceysanderson: I feel like this could be more readible with some small changes and more specific comment.

```suggestion
    // Ensure all the required keys are present (e.g. 'trait', 'method', 'unit').
    $pheno_combo_alias = array_keys(self::PHENOCOMBO_FIELD_MAP);
    $input_keys = array_keys($pheno_combo);
    if (array_diff($pheno_combo_alias, $input_keys)) {
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    $sanitized_pheno_combo = [];
+    $unexpected_values = [];
+
+    foreach ($pheno_combo_alias as $alias) {
+      if (!is_int($pheno_combo[$alias]) && !is_string($pheno_combo[$alias])) {
+        array_push($unexpected_values, $alias);
+      }
+
+      $sanitized_pheno_combo[$alias] = $pheno_combo[$alias];
+    }
```
laceysanderson, line 683: It feels really wrong to add unchecked values into $sanitized_pheno_combo here. Yes, you check the data type but it's added to the array regardless. Yes, if there is any wrong data type it will throw the exception but it still feels very wrong.

Instead, I would suggest keeping the data type check and using this loop to assign the $trait, $method and $unit variables.

```suggestion
    // Detect any unexpected data type and assign trait/method/unit variables.
    $unexpected_values = [];
    foreach ($pheno_combo_alias as $alias) {
      if (is_int($pheno_combo[$alias]) || is_string($pheno_combo[$alias])) {
        ${$alias} = $pheno_combo[$alias];
      }
      else {
        array_push($unexpected_values, $alias);
      }
    }
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
laceysanderson: I agree with Carolyn that a detailed comment would go a long way here ;-p

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+        )
+      );
+    }
+
+    // Verify trait, method and unit exist.
+    $pheno_combo = NULL;
+
+    foreach ($pheno_combo_alias as $alias) {
+      ${$alias} = $sanitized_pheno_combo[$alias];
+    }
```
laceysanderson, line 699: If you take my above suggestion, this loop is not needed at all.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+          'Experiment context not set error in %s: experiment context must be set before performing operations.
+           Use setExperiment(EXPERIMENT ID or NAME) to set an experiment context.',
+          get_class($this)
+        )
+      );
+    }
+  }
+
+  /**
+   * Sanitize PhenoCombo array.
```
laceysanderson: This provides no more information then the method name. Instead, what about?

```suggestion
   * Ensure PhenoCombo array only contains valid cvterm IDs.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    // Return resolved cvterm id, regardless of the PhenoCombo value provided.
+    foreach ($pheno_combo_alias as $alias) {
+      $sanitized_pheno_combo[$alias] = $pheno_combo[$alias]->cvterm_id;
+    }
+
+    return $sanitized_pheno_combo;
+  }
+
+  /**
+   * Sanitize experiment PhenoCombo details.
```
laceysanderson:
```suggestion
   * Validate and sanitize label and status flags.
   *
   * Note: default values are not filled in.
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+   *   An associative array containing label and combo status flags. Any of the
+   *   following optional keys may be provided and will be validated:
+   *   - label (string) uniquely identifying this trait-method-unit combo in
+   *     the experiment indicated.
+   *   - Status Flags
+   *     @see Drupal\trpcultivate_phenotypes\Service\ExperimentPhenoComboService::setExperimentPhenoComboStatus()
+   *
+   * @return array
+   *   A validated and sanitized experiment PhenoCombo details where status flag
+   *   values are guaranteed to be 0 or 1 in value and label is trimmed and is a
```
laceysanderson: You have an extra trailing space.

```suggestion
   *   values are guaranteed to be 0 or 1 in value and label is trimmed and is a
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+      return [];
+    }
+
+    $sanitized_phenocombo_details = [];
+
+    // If label is provided, ensure it is not an empty string and is a unique
+    // entry within the experiment context.
+    $field_label = 'label';
+
+    if (isset($phenocombo_details[$field_label])) {
```
laceysanderson, line 793: What if you pass in a $phenocombo_details where the label key is not set? I don't think you will get an error in this case 😬 I would suggest checking it isset and is a string here and throwing an excepting if it's not.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+      if (empty($label) || (!empty($label) && !$this->labelIsUniqueInExperiment($label))) {
+        throw new \InvalidArgumentException(
+          sprintf(
+            'Invalid PhenoCombo label error. %s provided must be a unique entry within the experiment and not an empty string.',
+            ucfirst($field_label)
+          )
+        );
+      }
+
+      $sanitized_phenocombo_details[$field_label] = $label;
```
laceysanderson: This still feels very wrong 🙈 It would feel much better as the following. This also ensures any coverage of the assignment to the sanitized array is caused by a valid case only.

```suggestion
      if (!empty($label) && $this->labelIsUniqueInExperiment($label)) {
        $sanitized_phenocombo_details[$field_label] = $label;
      }
      else {
        throw new \InvalidArgumentException(
          sprintf(
            'Invalid PhenoCombo label error. %s provided must be a unique entry within the experiment and not an empty string.',
            ucfirst($field_label)
          )
        );
      }
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,928 @@
...
+    foreach (self::PHENOCOMBO_STATUS_FLAG_FIELD_MAP as $field) {
+      if (isset($phenocombo_details[$field])) {
+        $status_flag_val = $phenocombo_details[$field];
+
+        if ($status_flag_val != 0 && $status_flag_val != 1) {
+          array_push($unexpected_values, $field);
+          continue;
+        }
+
+        $sanitized_phenocombo_details[$field] = $status_flag_val;
```
laceysanderson: Only fill in valid values to sanitized value array. It would also be good to cast the value in case it was provided as a string.

```suggestion
        if ($status_flag_val === 0 && $status_flag_val === 1) {
          $sanitized_phenocombo_details[$field] = (int) $status_flag_val;
        }
        else {
          array_push($unexpected_values, $field);
          continue;
        }
```

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Great catch! I've included the expression into the condition https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L886

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: The if condition is not intended to check for array or object types. Each expression inside the if() represents a failed return value from methods used to resolve project_id - null from getBackendRecordId(), 0 from getProjectName() and '' from getProjectId(). The ordering is the same as the flow of if-elseif. I have simplified this check to use empty($project) to capture all 3 current failed values (null, 0, '') plus others.

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L901

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Excellent suggestion, class property implemented https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L52

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Inline comment expanded - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L693

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Great catch X 2! Revised the test to include both - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php#L1074

image

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Oh dear, my bad, I meant to add a continue statement after adding a failed item. https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L671

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: The label is optional, the check is only carried out if details contained an element keyed 'label', that way this method can also be used to sanitize just status flags phenocombo detail.

```
// Can sanitize with label and status flag
$phenocombo_details = [
  'label' => 'A',
  'is_required' => 1,
  ...
]

// Or purely status flags
$phenocombo_details = [
  'was_collected' => '1',
  'is_required' => 1,
  ...
]

// Or just label
$phenocombo_details = [
   'label' => 'A',
]
```

I've expanded the test to check if arguments contained a label key but is empty. https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/671833ecb4e360dcaa6c0f039a3c9fe4eb5baed1/trpcultivate_phenotypes/tests/src/Kernel/Services/ServiceExperimentPhenoComboTest.php#L904

and expanded doc block by adding an example 

https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/0cce63f999e3f8a4fe9623fe3d0fe0048c171a9e/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L740

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: I got too excessive with ensuring that $label passed to labelIsUniqueInExperiment($label) is not empty - good catch and updated to https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/7bdc7cdbdba9a9b79bc198c0c8606f60b49a12f9/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L792

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: This has the line continue after registering an invalid value so it should not leak into the sanitized array. Type casted the status flags - good catch https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/3e8303401bca55d2a6e6e725b53df00406ade2c5/trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php#L815

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
```diff
@@ -0,0 +1,987 @@
...
+
+    $combo_id = $this->resolvePhenoCombo($combo);
+
+    // Ensure that this exp. PhenoCombo has no Phenotypes associated to it in
+    // Chado phenotype table.
+    $pheno_combo = $this->getExperimentPhenoCombo($combo_id);
+
+    $query = $this->chado_connection->select('1:phenotype', 'tbl');
+    if ($this->chado_connection->schema()->fieldExists('phenotype', 'project_id')) {
+      $query->condition('tbl.project_id', $this->experiment_context, '=');
```
laceysanderson: You don't have your test environment set up to adequately test this line.

When I looked into this closer I found it's because we actually haven't altered the chado phenotype table at all yet 😬  See PR #245 which alters chado... once that is merged you can use `Drupal\trpcultivate_phenotypes\Service\SetupModuleService::alterChadoTables();` to alter your test chado.

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: I added a comment to describe fully this block

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
reynoldtan: Addressed in commit - https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/223/commits/0d479e73ac508eb5d1e9d4c0290ea289e1d71ed9 with https://github.com/TripalCultivate/TripalCultivate-Phenotypes/pull/246 as dependent PR

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
laceysanderson: Looks good 👍

laceysanderson (commented):
✅ All my concerns previously brought up have been resolved 
✅ The test coverage is at a good place now with only SQL error untested

@carolyncaron and @vidu-wil were all your concerns addressed?

vidu-wil (dismissed):

trpcultivate_phenotypes/src/Service/ExperimentPhenoComboService.php
carolyncaron: The shorter variable name definitely helps - thanks! 👍

carolyncaron (dismissed):
All of my comments have been addressed! ✅

vidu-wil (approved):

PR #225, opened 2026-04-10
G4.224 Move setup methods from .module to a dedicated service class

vidu-wil:
**Issue #224**

## Motivation

The .module file for this module currently includes setup methods, which makes it harder to manage and keeping this logic in the module makes maintenance and testing harder. We need a separate Service class to replace this.

## What does this PR do?

Move all setup methods into a separate service class. Update the .module file and the .install file to use this service instead of containing the setup logic directly.

### Automated Testing

This PR does not add any new test classes or methods. However the PhenotypeTermInstallTest class has been updated to test the newly implemented installOntologyTerms method of the new class instead of the previous methods on .module file.

### Manual Testing

1.Start a fresh docker image/container on this branch.
2. Check if all the terms that's installed by this service has been installed properly. (Example queries shown below)
Screenshot 2026-04-13 at 9 33 35 AM

### DIFF
--- a/trpcultivate_phenotypes/src/Service/SetupModuleService.php
+++ b/trpcultivate_phenotypes/src/Service/SetupModuleService.php
@@ -1,0 +1,29 @@
+<?php
+
+namespace Drupal\trpcultivate_phenotypes\Service;
+
+/**
+ * Service class for setup module functions.
+ */
+class SetupModuleService {
+
+  /**
+   * Install the ontology terms.
+   *
+   * Expected to be run by a Tripal Job.
+   */
+  public static function installOntologyTerms() {
+    // We need the schema to ensure we are inserting into the right one.
+    $connection = \Drupal::service('tripal_chado.database');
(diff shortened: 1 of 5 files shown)


ruobinLiu:
Yes, all the terms that's installed by the service has been installed properly as expected during my test: 
Screenshot 2026-04-13 at 11 08 18 AM

ruobinLiu (dismissed):

reynoldtan:
Performed the manual test and verified the terms were installed correctly. I have one minor comment below.

Screenshot 2026-04-14 at 4 03 28 PM


1. The class doc block comment is still referring to .module when this is now in the .install
https://github.com/TripalCultivate/TripalCultivate-Phenotypes/blob/11c908f3154c9d7091bc31b9908da4551a3251d1/trpcultivate_phenotypes/tests/src/Kernel/PhenotypeTermInstallTest.php#L11

reynoldtan (approved):
Code looks good 👍

